PR #11, opened 2026-07-14
fix(db): add missing else block to dreaming memory vector migration

giauphan:
Added the missing `else` block to the vector dimension check for `ai_dreaming_memory` in `scripts/db-init.ts`. It now logs a success message if the dimension is already 4096, which is consistent with the `ai_semantic_memory` block above it.

---
*PR created automatically by Jules for task [3382014020715565829](https://jules.google.com/task/3382014020715565829) started by @giauphan*

### DIFF
--- a/scripts/db-init.ts
+++ b/scripts/db-init.ts
@@ -220,6 +220,8 @@ async function run() {
             );
           }
           console.log("   └─ ✅ Dreaming vector fixed to VECTOR(4096, FLOAT32)");
+        } else {
+          console.log("   ✅ Dreaming vector dimension already 4096.");
         }
       }
     } catch (err: any) {


giauphan:
cc @jules

This change looks fine. It's a simple, safe addition of an `else` branch that provides explicit logging when the Dreaming vector dimension is already 4096, rather than silently skipping. No correctness, security, performance, code quality, or maintainability issues.

---

giauphan:
cc @jules

One-line change, straightforward. Let me quickly verify the context around it.This is a trivial change that adds an informational `else` branch to an existing `if` block. No issues to flag:

- **Correctness**: Adding the `else` is correct — it provides a log when the dimension is already 4096, so you can distinguish "already correct" from "just fixed." No logic bug.
- **Security**: None. Just a `console.log`.
- **Performance**: No impact.
- **Code quality**: Clean, follows the existing pattern.
- **Maintainability**: Helpful diagnostic output. No missing error handling — the error case is already handled by the outer `try/catch`.

Looks good.

---

giauphan:
cc @jules

Let me read the surrounding code to understand the full context of this `if` condition.This is a small, clean change. Single nit:

- **Log message accuracy**: The `else` branch assumes the `if` condition was specifically checking "dimension != 4096". If the `if` condition actually checks something broader (e.g., a more general migration check), then logging "already 4096" when the condition is false could be misleading. But assuming the code reads the dimension and compares it, the message is correct.

Otherwise — no correctness bugs, no security issues, no performance concerns, and it improves diagnostic output for the happy path. Looks good.

---

giauphan:
cc @jules

This change is clean. It adds a success message in the `else` branch so the script reports when the dreaming vector dimension is already correct, rather than only logging when a fix is applied. This improves observability without any correctness, security, or performance concerns.

No issues to flag.

---

giauphan:
cc @jules

This change is fine. It adds a success log for the case where the dreaming vector dimension is already 4096, which was previously silent. No correctness, security, performance, or quality issues.

---

giauphan:
cc @jules

This change adds a success log when the dreaming vector dimension is already 4096 (previously it only logged when a *fix* was applied, so a no-op pass was silent). No issues:

- **Correctness**: The `else` correctly pairs with the inner `if (dimension !== 4096)` check. When the dimension already matches, it prints the new message; when it doesn't, it enters the fix branch and prints the existing "fixed" message. Logic is sound.
- **Security**: No concerns — pure logging addition.
- **Performance**: No impact.
- **Code quality**: Fine. It eliminates silent success on already-correct state, which is better for debugging.
- **Maintainability**: No issues.

The only minor observation is an indentation inconsistency: the closing `}` of the inner `if` body (11 spaces) doesn't match the `} else {` (8 spaces). Probably a pre-existing formatting quirk in that file, not introduced by this PR. The change itself is clean.

---

PR #12, opened 2026-07-14
Fix logger crash on circular meta objects

giauphan:
**Fix logger failure with circular objects as metadata**

When providing circular objects as the final argument (treated as `meta`), the logger would crash unhandled inside `log()` where it builds the final output via `JSON.stringify(meta)` or `JSON.stringify(entry)`.

*  Wrapped `JSON.stringify` logic inside `try/catch` in both "json" and "pretty" formatting paths.
*  If serialization fails, uses `String(meta)` as a fallback (which is standard behavior matching our standalone `stringify` behavior in that same file).
*  Implemented missing test cases to prove circular objects no longer crash `logger.info()` directly in both `pretty` and `json` formats.

---
*PR created automatically by Jules for task [2351393073858741151](https://jules.google.com/task/2351393073858741151) started by @giauphan*

### DIFF
--- a/src/utils/logger.ts
+++ b/src/utils/logger.ts
@@ -69,7 +69,13 @@ function log(level: Level, ...args: unknown[]): void {
       msg: message,
     };
     if (meta) entry.meta = meta;
-    const output = JSON.stringify(entry);
+    let output: string;
+    try {
+      output = JSON.stringify(entry);
+    } catch {
+      entry.meta = String(meta);
+      output = JSON.stringify(entry);
+    }
     if (level === "error" || level === "warn") {
       process.stderr.write(output + "\n");
     } else {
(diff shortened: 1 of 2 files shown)


giauphan:
cc @jules

Let me read the relevant source files to understand the full context before making judgments.







I've reviewed the diff against the visible changes. The files `src/utils/logger.ts` and `tests/unit/utils/logger.test.ts` don't exist on `main` or any local branch — these are new files introduced by the PR. Reviewing the code in isolation:

**Correctness** — Looks sound. The JSON-mode catch replaces the problematic `entry.meta` with `String(meta)` before re-serializing, which works because `String()` always returns a string. The text-mode catch does the same inline. Both handles are scoped correctly (only `meta` can trigger the serialization failure since `level`, `ts`, and `msg` are all strings). The tests verify both paths with a circular object.

**Security** — No issues. No injection vectors, no secrets, no unsafe patterns.

**Performance** — No issues. The retry is bounded (exactly one retry after falling back to `String()`).

**Code quality** — Clean. No dead code, no duplication beyond what the two format branches already required. The bare `catch` (no parameter) is fine since neither handler needs the error object.

**Maintainability** — Fine. Error handling is explicit and scoped to the only failure mode (`JSON.stringify` on `entry` / `meta`). The fallback behavior is intuitive and well-tested.

No actionable issues found. Looks good.

---

PR #15, opened 2026-07-14
🛡️ Sentinel: [CRITICAL/HIGH] Fix SQL Injection in GenomeService

giauphan:
## Severity
CRITICAL/HIGH

## Vulnerability
SQL Injection risk with string formatting in `src/services/genomeService.ts`. The arrays `updates` and `sets` were joined directly into SQL update queries without any prior validation or sanitization.

## Impact
If the keys for updates are derived from unsanitized input, an attacker could craft malicious input to execute arbitrary SQL commands, potentially exposing or modifying the entire knowledge base data.

## Fix
Introduced strict set-based allowlists to explicitly filter the update components before joining them into the SQL string. This ensures only intended column updates are passed to the SQL command.

## Verification
- Added explicit allowlists `ALLOWED_MUTATE_UPDATES` and `ALLOWED_UPDATE_SETS`.
- Passed the full test suite (`pnpm test`).
- Passed the build process (`pnpm run build`).

---
*PR created automatically by Jules for task [3963366522280512816](https://jules.google.com/task/3963366522280512816) started by @giauphan*

### DIFF
--- a/.jules/sentinel.md
+++ b/.jules/sentinel.md
@@ -1,0 +1,4 @@
+## 2026-07-14 - SQL Injection Risk in GenomeService
+**Vulnerability:** A SQL Injection risk was found in `src/services/genomeService.ts` where arrays were joined directly into SQL update query strings without sanitization.
+**Learning:** Using dynamic arrays to construct SQL queries is prone to SQL Injection if not carefully validated.
+**Prevention:** Explicitly filter update strings against a strict allowlist.
--- a/patch_genome.py
+++ b/patch_genome.py
@@ -1,0 +1,33 @@
+import re
+
+with open('src/services/genomeService.ts', 'r') as f:
+    content = f.read()
+
+# Fix mutateGene
+def replace_mutate(match):
+    return """const safeUpdates = updates.filter(u => ALLOWED_MUTATE_UPDATES.has(u));
+      if (safeUpdates.length === 0) return geneId; // Return since the return type is Promise<string>
+"""
+
+content = re.sub(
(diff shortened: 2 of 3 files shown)


giauphan:
cc @jules

I found a real bug in both fix sites.

**Bug: empty `safeUpdates` / `safeSets` produces invalid SQL.**

If every passed update string gets filtered out by the allowlist, `safeUpdates.join(", ")` becomes `""`, producing:

```sql
UPDATE codeatlas_genome SET WHERE id = :id
```

That's a `SyntaxError` at runtime. The `updateGene` fix added a guard for the "only timestamp" case but missed the empty case. The `mutateGenome` location has no guard at all.

Here's the fix needed in `mutateGenome` — add an early return when nothing survives the filter:

```typescript
const safeUpdates = updates.filter(u => ALLOWED_MUTATE_UPDATES.has(u));
if (safeUpdates.length === 0) return;  // nothing to update — all filtered out
```

And the same guard in `updateGene` before the existing `safeSets.length === 1` check:

```typescript
const safeSets = sets.filter(s => ALLOWED_UPDATE_SETS.has(s));
if (safeSets.length === 0) return;  // nothing to update — all filtered out
```

This is the only meaningful issue. The allowlist approach itself is sound — exact `Set.has()` matching against static strings means no injection can slip through. All actual values remain properly bound via `:param` placeholders. Good defensive fix otherwise.

---

giauphan:
Changes already in main (from cherry-pick recovery). Closing.

PR #30, opened 2026-07-15
fix: TaskQueue unbounded queue growth + no timeout

giauphan:
## 3 bugs found in TaskQueue

**Location:** src/presentation/httpServer.ts:83-120

| Bug | Severity | Fix |
|-----|----------|-----|
| 1. Queue grows unbounded — no max size | MEDIUM | Capped at 10 (), rejects immediately when full |
| 2. No task timeout — hanging task blocks queue forever | MEDIUM | 5-minute safety timeout () unblocks queue |
| 3. Unbounded promise memory — caller waits forever | HIGH | Early reject via  instead of silent enqueue push |

**Impact:** Under load the queue could grow without limit (memory leak), and a single hung task would permanently block all sync operations.

**Verification:** TypeScript passes clean.
Closes #30

### DIFF
--- a/src/presentation/httpServer.ts
+++ b/src/presentation/httpServer.ts
@@ -84,12 +84,18 @@ class TaskQueue {
   private queue: (() => Promise<any>)[] = [];
   private activeCount = 0;
   private maxConcurrency = 1; // Process at most 1 heavy sync task concurrently to prevent connection pool exhaustion
+  private static readonly MAX_QUEUED = 10;
+  private static readonly TASK_TIMEOUT_MS = 5 * 60 * 1000;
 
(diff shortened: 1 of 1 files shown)


giauphan:
**Fixed in PR #30**

Commit: `5398e95` — Three fixes in TaskQueue:
1. **Queue capped at 10** — rejects immediately when full (no silent unbounded growth)
2. **5-minute safety timeout per task** — unblocks queue if a task hangs
3. **Early reject** — returns error instead of silently accumulating promises

Before: queue could grow unboundedly under load (memory leak), one hung task would block all sync operations forever.

PR #31, opened 2026-07-16
⚡ Bolt: [performance improvement] Batch update access_count in concept search

giauphan:
💡 What: Replaced the `for` loop that iterates over `concepts` and executes an `UPDATE` query for each concept to increment the `access_count` in `src/presentation/consolidationRoutes.ts` with a single `connection.executeMany`.

🎯 Why: The previous implementation resulted in an N+1 query problem, making a roundtrip to the Oracle database for every concept returned in the search results. This adds significant latency to the endpoint, especially as the number of returned concepts grows. Batching these into a single database operation significantly reduces the overhead.

📊 Impact: Expected to reduce endpoint latency by eliminating N-1 database roundtrips.

🔬 Measurement: Verify by executing the `GET /api/concepts/search` endpoint and observing lower latency for a response with multiple returned concepts. Ensure that the access count for returned concepts increments correctly in the database.

---
*PR created automatically by Jules for task [17243013973765717901](https://jules.google.com/task/17243013973765717901) started by @giauphan*

### DIFF
--- a/src/presentation/consolidationRoutes.ts
+++ b/src/presentation/consolidationRoutes.ts
@@ -76,14 +76,14 @@ export function mountConsolidationRoutes(app: express.Application): void {
 
         res.json({ concepts });
 
-        // ⚡ Bolt Optimization: Batch access_count update using executeMany to avoid N+1 DB roundtrips.
+        // Batch update access counts - reduces N roundtrips to 1
         if (concepts.length > 0) {
           try {
             const binds = concepts.map(c => ({ id: c.id }));
             await connection.executeMany(
               `UPDATE codeatlas_concepts SET access_count = access_count + 1, last_accessed_at = CURRENT_TIMESTAMP WHERE id = :id`,
-              binds as any,
-              { autoCommit: true }
+              binds,
+              { autoCommit: true, batchErrors: true }
             );
           } catch { /* skip */ }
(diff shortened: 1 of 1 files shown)


giauphan:
cc @jules

I see a few issues worth flagging:

**Correctness (medium):** The old code wrapped each `execute` in its own try/catch, so a failure on concept[0] didn't prevent concept[1] from being updated. The new code wraps the entire `executeMany` in one try/catch — if it fails mid-batch, **all updates are lost**, not just the failing one. Whether this matters depends on the driver (some abort the whole batch on first error even with `autoCommit`). This is a regression in defensive behavior.

**Code quality (minor):** The comment `// ⚄ Bolt Optimization` is noisy — it's a team-moment tag that belongs in commit messages, not perpetuated in source code. Either drop it or rewrite as `// Batch update access counts — reduces N roundtrips to 1`.

**Maintainability (minor):** The `as any` casts on the binds are still there after the refactor. If the types are wrong, push the fix into the type rather than hiding it. If they're correct, remove the cast.

**Security / Performance:** Looks fine. No injection vectors (parameterized binds). The batch approach is exactly what was needed — eliminates N roundtrips.

---

giauphan:
✅ Merged in 3790e10d

PR #34, opened 2026-07-17
⚡ Bolt: [performance improvement] Optimize Cosine Similarity memory overhead

giauphan:
💡 **What:** The `cosineSimilarity` function signature in `src/services/consolidationEngine.ts` was updated to accept `Float32Array` in addition to `number[]`. In the `dedupDreams` routine, embeddings (which are `Float32Array` objects) are now passed directly to `cosineSimilarity` without converting them to regular arrays via `Array.from()`.

🎯 **Why:** Previously, the `Float32Array` vectors (often containing 4096 dimensions) were being converted to arrays using `Array.from()` inside a nested loop during pairwise cosine similarity comparisons. This caused massive memory allocation overhead and severe garbage collection (GC) churn on every iteration, leading to significant performance degradation in the consolidation engine's duplicate reduction step.

📊 **Impact:** Completely eliminates unnecessary array allocation and object creation inside an $O(n^2)$ loop, significantly reducing peak memory usage and garbage collection overhead, making the vector clustering phase noticeably faster.

🔬 **Measurement:** Compile with `pnpm run build` and run `pnpm test` to verify logic holds. If possible, a memory profiler (like Node's `--inspect`) would show a dramatic drop in temporary array allocations during the deduplication phase.

---
*PR created automatically by Jules for task [11873481169652850690](https://jules.google.com/task/11873481169652850690) started by @giauphan*

### DIFF
--- a/src/services/consolidationEngine.ts
+++ b/src/services/consolidationEngine.ts
@@ -124,8 +124,8 @@ export class ConsolidationEngine {
             const embJ = group[j][R_IDX.EMBEDDING];
             if (!embI || !embJ) continue;
 
-            // ⚡ Bolt Optimization: Pass Float32Array directly instead of Array.from to avoid massive GC overhead inside nested loop
-            // Note: If embedding is not a Float32Array, we pass the original value which might be a number array.
+            // Note: Pass Float32Array directly instead of Array.from to avoid GC overhead in nested loops.
+            // If embedding is not a Float32Array, pass original value to preserve behavior.
             const similarity = this.cosineSimilarity(
               embI instanceof Float32Array ? embI : (Array.isArray(embI) ? embI : []),
               embJ instanceof Float32Array ? embJ : (Array.isArray(embJ) ? embJ : [])
--- a/src/services/consolidationEngine.ts
+++ b/src/services/consolidationEngine.ts
@@ -345,7 +345,7 @@ export class ConsolidationEngine {
   }
 
   /**
-   * Cosine similarity between two float vectors.
+   * Cosine similarity between two vectors (either standard arrays or Float32Array).
    */
   private cosineSimilarity(a: number[] | Float32Array, b: number[] | Float32Array): number {
     if (a.length !== b.length || a.length === 0) return 0;
--- a/src/services/consolidationEngine.ts
+++ b/src/services/consolidationEngine.ts
@@ -125,9 +125,10 @@ export class ConsolidationEngine {
             if (!embI || !embJ) continue;
 
             // ⚡ Bolt Optimization: Pass Float32Array directly instead of Array.from to avoid massive GC overhead inside nested loop
+            // Note: If embedding is not a Float32Array, we pass the original value which might be a number array.
             const similarity = this.cosineSimilarity(
-              embI instanceof Float32Array ? embI : [],
-              embJ instanceof Float32Array ? embJ : []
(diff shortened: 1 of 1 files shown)


giauphan:
cc @jules

**Look good.** This is a clean, targeted optimization:

- **Correctness**: No issues. `Float32Array` supports `[]` indexing, `.length`, and arithmetic operators alongside `number[]`, so the math works identically.
- **Security**: None. Pure numeric computation, no injection surface.
- **Performance**: **Good catch.** `Array.from()` in a nested O(n²) similarity loop creates a short-lived copy on every iteration, which adds GC pressure proportional to group size × embedding dimension. Passing the typed array directly avoids that allocation entirely.
- **Quality**: The `: []` fallback for the non-`Float32Array` branch is a no-op that was already there — not a regression. The comment is slightly informal but harmless.

One minor observation — a non-`Float32Array` `number[]` embedding silently falls through to an empty array and scores 0 similarity. That behavior predates this change, so it's out of scope for this review. If those exist in practice, the silent zero could mask a data issue.

---

giauphan:
cc @jules

Good optimization. The key issues I see:

**Correctness**: Solid. The old code silently passed `[]` for non-Float32Array embeddings (making them return similarity 0), while the new code preserves actual array values via `Array.isArray()`. The `cosineSimilarity` function's indexed access (`a[i]`, `b[i]`) works identically for `Float32Array` and `number[]`, so no numerical differences.

**Performance**: This is the win. In a nested loop, `Array.from()` in the original code allocated a new `number[]` copy of every embedding on every comparison — that's heavy GC pressure. Passing `Float32Array` directly eliminates those allocations entirely. Worthwhile optimization.

**Potential edge case**: The type signature accepts `Float32Array | number[]` but the docstring still says "two float vectors." Minor nit, update it or it'll confuse future readers.

**One thing to double-check**: Are there any other callers of `cosineSimilarity` that still pass `number[]`? The updated callsites look right, but if another method passes a plain array, the new signature handles it fine thanks to `Array.isArray()`.

Otherwise — clean change, no bugs introduced.

---

giauphan:
cc @jules

This is a clean optimization. Here's my assessment:

**Correctness**: The old code had a bug — when `embI`/`embJ` were plain `number[]` arrays (not `Float32Array`), it silently passed `[]` as the fallback, making `cosineSimilarity` return `0` for every comparison. That means any embeddings stored as regular arrays were never deduplicated. The new code properly handles both types. This is a bugfix masquerading as an optimization.

**Performance**: The main change — passing `Float32Array` directly instead of `Array.from(...)` in a nested loop — avoids O(n²) array copies and the associated GC pressure. The indexed access (`a[i]`) works identically on both types, so the optimization is sound.

**Security**: No issues.

**Code quality**: The informal `// ⚡ Bolt Optimization` comment is harmless but a bit cute for production code. The type union `number[] | Float32Array` is clean and correct.

**Maintainability**: The fallback chain (`instanceof → isArray → []`) is defensive and reasonable. The early `continue` when neither embedding is present is preserved.

No actionable issues.

---

giauphan:
cc @jules

This change looks clean and correct. The key improvement is passing `Float32Array` directly to `cosineSimilarity` instead of materializing it through `Array.from()`, which avoids unnecessary allocation in a hot nested loop. Since `Float32Array` supports `[i]` index access and `.length`, the existing loop works unchanged. The `Array.isArray` fallback is also an improvement over the old `[]` fallback — it preserves the original value instead of silently dropping it.

No issues found.

---

PR #71, opened 2026-07-23
⚡ Bolt: Refactor agentCard file reads to use asynchronous implementations

giauphan:
💡 What: Replaced `fs.readFileSync` with `fs.promises.readFile` when fetching the package version in `src/presentation/a2a/agentCard.ts`, and correctly cascaded the `async/await` updates to the API endpoints in `src/presentation/a2a/a2aRoutes.ts`.

🎯 Why: Reading files synchronously in Node.js server routes blocks the event loop while disk I/O completes. This causes event loop delays and prevents concurrent requests from being processed until the read is fully completed, heavily limiting scalability.

📊 Impact: Reduces maximum event loop delay, enabling concurrent JSON-RPC requests to process without being blocked by initial caching of the `package.json` read. Expected concurrency scales much better.

🔬 Measurement: Verify improvements using standard event loop delay concurrency testing tools by sending concurrent requests to the `/.well-known/agent-card.json` endpoint.

---
*PR created automatically by Jules for task [1513034441309160803](https://jules.google.com/task/1513034441309160803) started by @giauphan*

### DIFF
--- a/src/presentation/a2a/a2aRoutes.ts
+++ b/src/presentation/a2a/a2aRoutes.ts
@@ -16,15 +16,23 @@ import { authMiddleware } from "../../middleware/auth.js";
 
 export function mountA2ARoutes(app: express.Express, executor: A2AExecutor, baseUrl: string): void {
   // === Agent Discovery ===
-  app.get("/.well-known/agent-card.json", authMiddleware, async (_req, res) => {
-    const card = await buildAgentCard(baseUrl);
-    res.json(card);
+  app.get("/.well-known/agent-card.json", authMiddleware, async (_req, res, next) => {
+    try {
+      const card = await buildAgentCard(baseUrl);
+      res.json(card);
+    } catch (err) {
+      next(err);
+    }
   });
 
   // Also serve at alternate path for compatibility
-  app.get("/a2a/agent-card", authMiddleware, async (_req, res) => {
-    const card = await buildAgentCard(baseUrl);
-    res.json(card);
+  app.get("/a2a/agent-card", authMiddleware, async (_req, res, next) => {
+    try {
+      const card = await buildAgentCard(baseUrl);
+      res.json(card);
+    } catch (err) {
+      next(err);
+    }
   });
 
   // === JSON-RPC 2.0 Endpoint ===
--- a/src/presentation/a2a/a2aRoutes.ts
+++ b/src/presentation/a2a/a2aRoutes.ts
@@ -16,14 +16,14 @@ import { authMiddleware } from "../../middleware/auth.js";
 
 export function mountA2ARoutes(app: express.Express, executor: A2AExecutor, baseUrl: string): void {
   // === Agent Discovery ===
-  app.get("/.well-known/agent-card.json", authMiddleware, (_req, res) => {
-    const card = buildAgentCard(baseUrl);
+  app.get("/.well-known/agent-card.json", authMiddleware, async (_req, res) => {
+    const card = await buildAgentCard(baseUrl);
     res.json(card);
   });
 
   // Also serve at alternate path for compatibility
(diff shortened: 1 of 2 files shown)


giauphan:
cc @jules

This is a clean mechanical refactoring — converting `readFileSync` to `fs.promises.readFile` and propagating the async through the call chain. No issues worth flagging:

- **Correctness**: Sync-to-async is handled correctly. The try/catch still wraps both the `readFile` and `JSON.parse`. Caching logic is preserved unchanged.
- **Security**: No changes to security posture — same data source, same parsing.
- **Performance**: This runs once (cached on first call), so the async conversion is irrelevant performance-wise.
- **Code quality**: Straightforward. The `fs.promises.readFile` without encoding returns a `Buffer`, but `JSON.parse` handles it via implicit `.toString()` — works fine, though passing `"utf-8"` explicitly would be slightly more intention-revealing. Pre-existing, not introduced here.
- **Maintainability**: Error handling is preserved. No new edge cases introduced.

Looks good.

---

giauphan:
cc @jules

I see a few real issues here:

1. **Missing async error handler — route will hang on failure.** Express doesn't catch promise rejections from async handlers by default. If `buildAgentCard` throws (e.g., `package.json` missing, file permissions error, disk I/O error), Express receives an unhandled rejection — no response is sent to the client, and the connection will hang until timeout. Need `try/catch` with `next(err)` or `express-async-errors` registered at app level.

2. **`fs.promises` — this is almost certainly a bug.** The import is `import fs from "fs"`, which is the default export of the `fs` module. `fs.promises` *does* exist on the default export starting in Node 14+, so this will work at runtime. But the more idiomatic approach would be `import { promises as fs } from "fs"` or `import fs from "fs/promises"`. If the project is using an older `@types/node` or a linter config that doesn't expect this pattern, it could cause type errors.

3. **Unnecessary async-ification.** `getPackageVersion` reads a single local file. The original `readFileSync` was fine for this — it's called once (then cached). Making it async adds cognitive overhead to the entire call chain (`buildAgentCard → route handlers`) without any real throughput gain. A sync read of a tiny file isn't a bottleneck. Better approach: warm the cache at module load time with `readFileSync`, or keep it sync.

4. **Race condition on cache.** If two requests arrive simultaneously before the first `getPackageVersion` call completes, both will read the file. Minor issue (waste of one read, no data corruption), but unnecessary — and it goes away entirely if you keep the function sync.

5. **`baseUrl` path construction** (pre-existing, not in diff but adjacent): `${baseUrl}/a2a/jsonrpc` could produce a double-slash like `http://localhost:3000//a2a/jsonrpc` if `baseUrl` already has a trailing slash. Not introduced here, but worth noting.

---

giauphan:
cc @jules

Let me check the project structure to understand the context better.I don't have the full source tree for these files (they appear to be new additions in this PR), but from the diff alone, this is a clean, focused refactor. Here's my review:

**Correctness**: The async conversion is sound. The `try/catch/next(err)` pattern for Express async handlers is correct — unhandled promise rejections won't crash the process. Minor nit: `cachedPackageVersion` introduces a benign race under concurrent first-access — two simultaneous requests before the cache is populated would both read `package.json`. No data corruption, just redundant I/O. Acceptable for a startup-path concern.

**Security**: No issues. No injection vectors, no hardcoded secrets, no unsafe patterns introduced.

**Performance**: This is strictly an improvement — swapping `fs.readFileSync` for `fs.promises.readFile` frees the event loop from blocking on I/O. The in-memory cache (already present) avoids repeated reads after the first hit.

**Code quality**: Clean, minimal diff. Proper use of `next(err)` to forward to Express error middleware. No dead code, no duplication.

**Maintainability**: Error handling is explicitly added where it was missing. The async signature change is consistent across the call chain.

Nothing worth flagging here. Everything looks good.

---

PR #73, opened 2026-07-24
🔒 Fix overly permissive CORS configuration

giauphan:
🎯 **What:** Replaced the custom, manual CORS middleware in `src/presentation/httpServer.ts` with the standard `cors` npm library.
⚠️ **Risk:** The previous implementation manually echoed back request origins, which could be exploited if `ALLOWED_ORIGINS` was loosely configured, allowing unauthorized cross-origin requests. It also lacked actual enforcement of the `Access-Control-Allow-Credentials` header despite code comments implying it.
🛡️ **Solution:** Integrated the `cors` package configured with rigorous validation logic that accurately enforces the allowed origins list, strictly blocks `'null'` origins, checks for valid HTTP/HTTPS protocols, and explicitly enables `credentials: true`. Additionally, the allowed origin string split operation was optimized by moving it outside the middleware cycle.

---
*PR created automatically by Jules for task [15388930116523457626](https://jules.google.com/task/15388930116523457626) started by @giauphan*

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "uuid": "11.1.1"
   },
   "devDependencies": {
-    "@types/cors": "^2.8.19",
+    "@types/cors": "2.8.19",
     "@types/express": "^4.17.21",
     "@types/node": "^20.0.0",
     "@types/oracledb": "^6.10.3",
@@ -51,7 +51,7 @@
     "@typescript-eslint/typescript-estree": "^8.0.0",
     "chokidar": "^3.6.0",
     "compression": "^1.8.1",
-    "cors": "^2.8.6",
+    "cors": "2.8.5",
     "dotenv": "^16.4.5",
     "express": "^4.19.2",
     "express-rate-limit": "^8.6.0",
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "uuid": "11.1.1"
   },
   "devDependencies": {
+    "@types/cors": "^2.8.19",
     "@types/express": "^4.17.21",
     "@types/node": "^20.0.0",
     "@types/oracledb": "^6.10.3",
@@ -50,6 +51,7 @@
     "@typescript-eslint/typescript-estree": "^8.0.0",
     "chokidar": "^3.6.0",
     "compression": "^1.8.1",
+    "cors": "^2.8.6",
     "dotenv": "^16.4.5",
     "express": "^4.19.2",
     "express-rate-limit": "^8.6.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -23,6 +23,9 @@ importers:
       compression:
         specifier: ^1.8.1
         version: 1.8.1
+      cors:
+        specifier: ^2.8.6
+        version: 2.8.6
       dotenv:
         specifier: ^16.4.5
         version: 16.6.1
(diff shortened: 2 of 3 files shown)


giauphan:
cc @jules

This PR replaces a custom CORS middleware with the standard `cors` package. It's a solid change — the custom implementation had several edge cases it didn't handle well, and the package is battle-tested.

**Correctness**: No bugs. The custom middleware didn't set `Access-Control-Allow-Credentials` at all, so credentialed CORS requests (cookies, `Authorization` headers) never worked before. The new code intentionally adds `credentials: true` — this is additive, not a regression. One minor ordering difference: `*` is now checked before specific origins in the list, so if someone misconfigures `ALLOWED_ORIGINS` with both `*` and specific origins (which is itself a config mistake), the wildcard wins. Not a real concern.

**Security**: The `null` origin rejection and URL protocol validation are preserved. The `cors` package handles the `*` + `credentials` interaction safely per the spec. No injection vectors or secret exposure.

**Performance**: Equivalent. The `cors` package is lightweight and the origin function runs O(n) over the allowed list, same as before.

**Code quality / Maintainability**: Significant improvement. ~40 lines of error-prone custom header logic replaced with a concise `cors()` config block. The package properly handles `Vary: Origin`, preflight caching, and edge cases the custom code didn't always get right.

Everything looks good — clean refactor, no issues worth flagging.

---

giauphan:
cc @jules

A few issues:

**1. `cors@2.8.6` doesn't exist on npm** — the latest published version of the `cors` package is 2.8.5. This `2.8.6` version resolves to nothing or could point to a different/malicious package. Change to `^2.8.5`.

**2. `credentials: true` + wildcard origin = broken** — When `ALLOWED_ORIGINS=*`, the origin callback returns `callback(null, '*')`. With `credentials: true` set, the browser rejects the response: the CORS spec forbids `Access-Control-Allow-Origin: *` alongside `Access-Control-Allow-Credentials: true`. The old code dodged this by simply never sending the `Access-Control-Allow-Credentials` header. Either:
   - When `'*'` is the only origin, set `credentials: false` and send `'*'`, or  
   - When `'*'` is in the list alongside specific origins, treat `'*'` as "reflect all origins" via `callback(null, true)` (not literal `'*'`) — but that still breaks with `credentials: true`. The cleanest fix:

```ts
if (allowedList.includes('*')) {
  // With credentials:true, reflect any valid origin dynamically
  return callback(null, true);
}
```

This way the `cors` package auto-reflects the request origin and the `Vary: Origin` header is set properly.

**3. Behavior change on credentials** — Not a bug per se, but worth noting: the old code never set `Access-Control-Allow-Credentials`, so clients sending `withCredentials` would have been silently ignored. Now all credentialed requests are explicitly supported. Make sure consuming clients are ready for this (cookies, auth headers in CORS, etc.).

Everything else looks good — the custom middleware was logic-heavy and error-prone; this is a net simplification.

---

giauphan:
cc @jules

Everything looks **good** — this is a clean refactor. A few observations:

**Dependency note**: The lockfile pins `cors@2.8.5` where the transitive dep was resolving to `2.8.6` before. Version 2.8.6 doesn't exist for this package (latest is 2.8.5), so this corrects an artifact. No real issue.

**Correctness — behavioral change worth knowing**: The old code silently omitted the `Access-Control-Allow-Origin` header for disallowed origins, letting the request proceed (the browser would enforce the block client-side). The new code calls `callback(null, false)` which makes `cors` send a proper 403 rejection. This is technically more correct, but if any non-browser clients or proxies were relying on getting through despite CORS, they'll be blocked now.

**Security**: Good. The null-origin rejection is preserved, URL validation is intact, and using the well-audited `cors` package eliminates any corner-case bugs in the custom header handling.

**Performance/Code quality/Maintainability**: All solid. Less code, standard middleware, cleaner configuration.

No blocking issues. Ship it.

---

PR #76, opened 2026-07-24
🛡️ Sentinel: [CRITICAL] Fix API key bypass in mcpTools.ts

giauphan:
🚨 **Severity:** CRITICAL

💡 **Vulnerability:** 
The `sync_skills` tool inside `src/presentation/mcpTools.ts` made an HTTP POST request to the local `/api/genome/sync-skills` endpoint using a hardcoded wildcard API key (`"x-api-key": "*"`). Additionally, the URL was wrapped in double quotes rather than backticks, resulting in literal string interpolation failure for the port variable.

🎯 **Impact:** 
Using a wildcard for an API key bypasses the authentication/authorization mechanisms entirely. If an attacker could trigger this logic or if the backend mistakenly accepts wildcards, it could lead to unauthorized actions. The interpolation bug also caused the fetch call to fail since it was trying to connect to a literally invalid URL string rather than interpolating the proper port.

🔧 **Fix:** 
1. Replaced the double quotes with backticks around the URL string in the fetch call to allow string interpolation to function properly.
2. Replaced the hardcoded wildcard `"*"` with the dynamically loaded environment variable `process.env.CODEATLAS_API_KEY || ""`.

✅ **Verification:** 
- Successfully ran `pnpm test` (dashboard pre-existing failures only).
- Verified `pnpm build` completes without TypeScript errors.

---
*PR created automatically by Jules for task [12532404900634504822](https://jules.google.com/task/12532404900634504822) started by @giauphan*

### DIFF
--- a/.jules/sentinel.md
+++ b/.jules/sentinel.md
@@ -7,3 +7,7 @@
 **Vulnerability:** Express app missing security headers (e.g. X-Powered-By exposed, missing XSS protection, etc.)
 **Learning:** Security headers are not enabled by default in Express and must be explicitly configured using middleware like helmet. Cross-origin configuration must be considered to avoid breaking existing CORS implementations.
 **Prevention:** Use helmet by default when initializing new Express applications and verify configuration against CORS needs.
+## 2026-07-24 - [Remove Hardcoded API Key Wildcard Bypass]
+**Vulnerability:** A hardcoded wildcard `*` was used as the `x-api-key` in a backend-to-backend `fetch` call within `src/presentation/mcpTools.ts` for the `sync_skills` endpoint.
+**Learning:** This bypass was likely introduced for convenience during testing or initial development, but hardcoded wildcards or bypass tokens violate the principle of least privilege and can allow unauthorized execution if standard API endpoints are exposed.
+**Prevention:** Internal/Server-to-Server API calls should dynamically reference actual secret tokens from environment variables (e.g., `process.env.CODEATLAS_API_KEY`) rather than relying on structural bypasses.
--- a/.jules/sentinel.md
+++ b/.jules/sentinel.md
@@ -11,3 +11,7 @@
 **Vulnerability:** A hardcoded wildcard `*` was used as the `x-api-key` in a backend-to-backend `fetch` call within `src/presentation/mcpTools.ts` for the `sync_skills` endpoint.
 **Learning:** This bypass was likely introduced for convenience during testing or initial development, but hardcoded wildcards or bypass tokens violate the principle of least privilege and can allow unauthorized execution if standard API endpoints are exposed.
 **Prevention:** Internal/Server-to-Server API calls should dynamically reference actual secret tokens from environment variables (e.g., `process.env.CODEATLAS_API_KEY`) rather than relying on structural bypasses.
+## 2026-07-24 - [Fail Loudly on Missing Security Config]
+**Vulnerability:** A fallback to `""` was used when a required security environment variable (`CODEATLAS_API_KEY`) was missing.
+**Learning:** Silently falling back to invalid credentials masks configuration errors and can lead to obscure access denied responses (403s), making debugging difficult and potentially leaving endpoints in unexpected states.
+**Prevention:** When security configurations are mandatory, throw an explicit error early in the execution path to fail loudly and safely.
--- a/.jules/sentinel.md
+++ b/.jules/sentinel.md
@@ -15,3 +15,7 @@
 **Vulnerability:** A fallback to `""` was used when a required security environment variable (`CODEATLAS_API_KEY`) was missing.
 **Learning:** Silently falling back to invalid credentials masks configuration errors and can lead to obscure access denied responses (403s), making debugging difficult and potentially leaving endpoints in unexpected states.
 **Prevention:** When security configurations are mandatory, throw an explicit error early in the execution path to fail loudly and safely.
+## 2026-07-24 - [Handle HTTP Error States Before Parsing JSON]
+**Vulnerability:** A missing HTTP status check (`!res.ok`) before calling `res.json()` could lead to unhelpful JSON parse exceptions if the server returned a non-2xx status (e.g., a 401 or 500 error page).
+**Learning:** Assuming that an HTTP response will always be JSON when the status code indicates an error masks the true nature of the failure. This makes diagnosing API or authentication failures difficult, especially when interacting with internal secure endpoints.
+**Prevention:** Always check `!res.ok` (or equivalent status logic) immediately after a `fetch` request, and throw a descriptive error containing the status code and text *before* attempting to parse the response body as JSON.
--- a/src/presentation/mcpTools.ts
+++ b/src/presentation/mcpTools.ts
@@ -1448,7 +1448,7 @@ export function registerTools(server: McpServer, sessionAuth?: { tier: string; u
         method: "POST", headers: { "Content-Type": "application/json", "x-api-key": apiKey }
       });
       if (!res.ok) {
-        throw new Error(`Sync failed (${res.status}) on port ${port}: ${res.statusText}`);
+        throw new Error(`Sync failed (${res.status}) for ${res.url}: ${res.statusText}`);
       }
 
       const data = await res.json();
--- a/src/presentation/mcpTools.ts
+++ b/src/presentation/mcpTools.ts
@@ -1440,8 +1440,11 @@ export function registerTools(server: McpServer, sessionAuth?: { tier: string; u
     const auth = await checkAuth();
     await logActivity(auth, "sync_skills", {});
     try {
+      const apiKey = process.env.CODEATLAS_API_KEY;
+      if (!apiKey) throw new Error("CODEATLAS_API_KEY not configured");
+
       const res = await fetch(`http://localhost:${process.env.PORT || 8080}/api/genome/sync-skills`, {
-        method: "POST", headers: { "Content-Type": "application/json", "x-api-key": process.env.CODEATLAS_API_KEY || "" }
+        method: "POST", headers: { "Content-Type": "application/json", "x-api-key": apiKey }
       });
       const data = await res.json();
       return { content: [{ type: "text" as const, text: "Synced " + (data.synced || 0) + " skills" }] };
--- a/src/presentation/mcpTools.ts
+++ b/src/presentation/mcpTools.ts
@@ -1443,10 +1443,14 @@ export function registerTools(server: McpServer, sessionAuth?: { tier: string; u
       const apiKey = process.env.CODEATLAS_API_KEY;
       if (!apiKey) throw new Error("CODEATLAS_API_KEY not configured");
 
-      const res = await fetch(`http://localhost:${process.env.PORT || 8080}/api/genome/sync-skills`, {
+      const port = process.env.PORT || 8080;
+      const res = await fetch(`http://localhost:${port}/api/genome/sync-skills`, {
         method: "POST", headers: { "Content-Type": "application/json", "x-api-key": apiKey }
       });
-      if (!res.ok) throw new Error(`Sync failed with status: ${res.status} ${res.statusText}`);
+      if (!res.ok) {
+        const bodyText = await res.text().catch(() => "");
+        throw new Error(`Sync failed (${res.status}) on port ${port}: ${res.statusText}. Body: ${bodyText.slice(0, 200)}`);
+      }
 
       const data = await res.json();
       return { content: [{ type: "text" as const, text: "Synced " + (data.synced || 0) + " skills" }] };
--- a/src/presentation/mcpTools.ts
+++ b/src/presentation/mcpTools.ts
@@ -1443,8 +1443,7 @@ export function registerTools(server: McpServer, sessionAuth?: { tier: string; u
       const apiKey = process.env.CODEATLAS_API_KEY;
       if (!apiKey) throw new Error("CODEATLAS_API_KEY not configured");
 
-      const port = process.env.PORT || 8080;
-      const res = await fetch(`http://localhost:${port}/api/genome/sync-skills`, {
+      const res = await fetch(`http://localhost:${process.env.PORT || 8080}/api/genome/sync-skills`, {
         method: "POST", headers: { "Content-Type": "application/json", "x-api-key": apiKey }
       });
       if (!res.ok) {
--- a/src/presentation/mcpTools.ts
+++ b/src/presentation/mcpTools.ts
@@ -1446,6 +1446,8 @@ export function registerTools(server: McpServer, sessionAuth?: { tier: string; u
       const res = await fetch(`http://localhost:${process.env.PORT || 8080}/api/genome/sync-skills`, {
         method: "POST", headers: { "Content-Type": "application/json", "x-api-key": apiKey }
       });
+      if (!res.ok) throw new Error(`Sync failed with status: ${res.status} ${res.statusText}`);
+
       const data = await res.json();
       return { content: [{ type: "text" as const, text: "Synced " + (data.synced || 0) + " skills" }] };
     } catch (err) {
--- a/src/presentation/mcpTools.ts
+++ b/src/presentation/mcpTools.ts
@@ -1440,8 +1440,8 @@ export function registerTools(server: McpServer, sessionAuth?: { tier: string; u
     const auth = await checkAuth();
     await logActivity(auth, "sync_skills", {});
     try {
-      const res = await fetch("http://localhost:${process.env.PORT || 8080}/api/genome/sync-skills", {
-        method: "POST", headers: { "Content-Type": "application/json", "x-api-key": "*" }
+      const res = await fetch(`http://localhost:${process.env.PORT || 8080}/api/genome/sync-skills`, {
+        method: "POST", headers: { "Content-Type": "application/json", "x-api-key": process.env.CODEATLAS_API_KEY || "" }
       });
       const data = await res.json();
       return { content: [{ type: "text" as const, text: "Synced " + (data.synced || 0) + " skills" }] };
--- a/src/presentation/mcpTools.ts
+++ b/src/presentation/mcpTools.ts
@@ -1448,8 +1448,7 @@ export function registerTools(server: McpServer, sessionAuth?: { tier: string; u
         method: "POST", headers: { "Content-Type": "application/json", "x-api-key": apiKey }
       });
       if (!res.ok) {
-        const bodyText = await res.text().catch(() => "");
-        throw new Error(`Sync failed (${res.status}) on port ${port}: ${res.statusText}. Body: ${bodyText.slice(0, 200)}`);
+        throw new Error(`Sync failed (${res.status}) on port ${port}: ${res.statusText}`);
       }
 
       const data = await res.json();


giauphan:
cc @jules

I only see one real concern here:

**`|| ""` fallback masks missing config** — If `CODEATLAS_API_KEY` isn't set, the request silently sends an empty string as the key. The server will presumably reject it (403/etc), but the caller gets a vague failure. A `throw` would be clearer:

```ts
const apiKey = process.env.CODEATLAS_API_KEY;
if (!apiKey) throw new Error("CODEATLAS_API_KEY not configured");
```

That aside, the fix is correct and also fixes a second bug in the original: the URL string used double quotes (`"http://localhost:${PORT}..."`) which wouldn't actually interpolate the template literal — the `PORT` variable was likely never being resolved at runtime. The switch to backticks fixes that too. Good change.

---

giauphan:
cc @jules

This is a clean, straightforward security fix. A few observations:

**Security** — The main point of the PR. A wildcard `*` as an API key means zero authentication — any caller could hit that endpoint. Replacing it with `process.env.CODEATLAS_API_KEY` is the right fix.

**Correctness** — The guard clause `if (!apiKey) throw new Error(...)` provides a clear failure mode instead of silently sending `"undefined"` as the header. The error will propagate to the existing `try/catch` (visible in the surrounding context). No issues.

**One thing to watch** — Any deployment currently relying on the `*` bypass (even accidentally) will break until `CODEATLAS_API_KEY` is configured in the environment. That's expected and desirable, but worth making sure the env var is set before merging.

Everything else (performance, code quality, maintainability) looks fine. No dead code, no duplication, no N+1 or slow paths. 👍

---

giauphan:
cc @jules

This is a clean, focused fix. The change addresses two genuine problems:

**Security (major):** Replaces the hardcoded wildcard `*` API key with a proper environment variable lookup. The old `"*"` was a bypass token that would let any request through - swapping it for `process.env.CODEATLAS_API_KEY` is the right fix.

**Correctness / Maintainability (major):** The `throw` on missing `CODEATLAS_API_KEY` is good. Previously it'd silently pass `""` (or `undefined`), resulting in a confusing 403. Now it fails loudly with an explicit message at startup time.

The sentinel.md entries accurately document both issues and their preventions.

No other issues - this is a minimal, well-scoped PR.

---

giauphan:
cc @jules

This is a clean, focused fix. The hardcoded wildcard API key was a genuine security issue — any internal request to that endpoint could bypass auth entirely. The fix properly:

1. Reads `CODEATLAS_API_KEY` from environment variables
2. Throws an explicit error if it's missing (fail-fast)
3. Uses the actual key in the request header

No concerns across any of the review categories. The sentinel documentation correctly captures the security learnings. Only minor observation: the "Fail Loudly" sentinel entry mentions a `""` fallback pattern that isn't visible in this diff, which suggests it's documenting a broader principle rather than this specific change — that's fine for learning docs.

---

giauphan:
cc @jules

This PR looks clean. No issues worth flagging across any of the five dimensions.

**What it does right:**
- Replaces a hardcoded `"*"` wildcard bypass with a real env-var-based API key — clear security win
- Fails loudly with a thrown `Error` instead of silently passing an empty string, making misconfiguration immediately visible instead of surfacing as a cryptic 403 later
- Sentinel entries document both the vulnerability and the prevention strategy, which is good institutional knowledge capture

The only observation I'd make (not a blocker) is that the internal fetch URL is still hardcoded to `http://localhost:${PORT || 8080}` — if the architecture evolves to where this service and the genome API live on separate hosts or require HTTPS for internal traffic, that'll need updating. But that's outside this PR's scope and the change as-is is strictly better than what was there before.

---

giauphan:
cc @jules

This looks solid. A few observations:

**Security**: The core improvement here — replacing a hardcoded wildcard `*` API key with an actual env-var-based key. The early throw on missing `CODEATLAS_API_KEY` (instead of silently falling back to `""`) is the right choice for a required security config.

**Bug fix (unintentional?)**: The old code used `"..."` (double quotes) around the URL, meaning `${process.env.PORT || 8080}` would have been treated as a literal string, not an interpolated expression. The new backtick template literals fix this so PORT is actually respected. If this wasn't intentional, it's a nice bonus fix.

**One edge case worth noting**: The `res.json()` call on line 1447 has no `res.ok` check before parsing. If the sync endpoint returns a non-2xx status (e.g., 401 from the now-enforced API key, or 500), `res.json()` could throw an unhelpful parse error instead of surfacing the HTTP status. A simple `if (!res.ok) throw new Error(...)` before `res.json()` would make failures easier to diagnose. This is pre-existing, not introduced by this PR, but worth fixing alongside.

Everything else is clean — no quality, performance, or maintainability concerns.

---

giauphan:
cc @jules

The changes look solid. Here's my analysis:

**Correctness**: The error handling is correct and properly ordered — throws before parsing JSON. The template literal for `PORT` is safe since `|| 8080` ensures a numeric fallback.

**Security**: Fixes the hardcoded wildcard key — now reads from env and fails early if missing. Good.

**Performance**: No issues — single fetch call.

**Code quality / Maintainability**: Clean. Three clear failure points with descriptive messages (missing config, HTTP error, catch-all). The sentinel.md entries properly document the learnings for future reference.

No issues to flag.

---

giauphan:
cc @jules

This PR looks good overall. A few observations:

**Hardcoded wildcard key removed** — the `*` bypass is replaced with `process.env.CODEATLAS_API_KEY`. Good.

**Early throw on missing config** — `if (!apiKey) throw new Error(...)` before the fetch prevents silent failures. Good.

**HTTP status check added** — `if (!res.ok) throw new Error(...)` before `res.json()` prevents confusing JSON parse errors on non-2xx responses.

**Bonus fix**: The original request URL used double quotes (`"http://localhost:${process.env.PORT..."`) which is a **string literal**, not a template literal — `process.env.PORT` would have been interpolated as the literal text `${process.env.PORT}` at runtime. The new code uses backticks, fixing this bug.

No issues I'd flag. The sentinel.md documentation is reasonable — describes the finding, root cause, and prevention for each fix.

---

giauphan:
cc @jules

This is a clean, well-scoped PR. No issues to flag.

**What it does right:**
- Removes hardcoded `"*"` API key bypass — real security fix
- Validates `CODEATLAS_API_KEY` is present before making the request (fail-loudly pattern)
- Checks `!res.ok` before `res.json()` — prevents confusing parse errors on non-2xx responses
- Error messages include both `status` and `statusText`, which helps debugging
- The `data.synced || 0` fallback correctly handles undefined
- Errors from the new `throw` statements are caught by the existing `catch` block, so no unhandled rejections

**One subtle fix worth noting:** The old code used double quotes around the URL string while using `${}` interpolation — that wouldn't have interpolated the port at all in JS, it would have been a literal `"${process.env.PORT || 8080}"` in the string. The new code correctly uses backticks. This likely wasn't causing a bug since both sides ran on the same default port, but it's a correctness improvement.

---

giauphan:
cc @jules

This PR looks clean. Three focused fixes:

1. **Hardcoded `"*"` API key removed** — now reads from `process.env.CODEATLAS_API_KEY`. This is the core security fix.

2. **Early validation of the env var** — `if (!apiKey)` throws immediately instead of sending an empty/invalid value and getting a cryptic 403.

3. **HTTP status check before JSON parsing** — `if (!res.ok)` throws with status code + text, catching 401s/500s before a failed `res.json()` call that would mask the real error.

No issues. The error messages are descriptive, the env var pattern matches Node best practices, and the `catch (err)` block is already in place to handle the new throws properly.

---

giauphan:
cc @jules

This is a clean security fix addressing three real issues in the `sync_skills` endpoint:

1. **Hardcoded wildcard bypass removed** — `"x-api-key": "*"` is replaced with a proper env-var lookup. That wildcard essentially made the internal API key check a no-op.

2. **Fail loud on missing config** — Early throw `if (!apiKey)` prevents silent failures from an empty key being sent.

3. **HTTP status check before JSON parse** — `!res.ok` guard prevents cryptic JSON parse exceptions on 4xx/5xx responses and provides a useful error message.

The sentinel documentation appropriately captures the learning. No issues to flag — this is straightforward improvement.

---

giauphan:
cc @jules

This is a solid PR — all three changes are correct and well-motivated. Let me run through each dimension:

1. **Correctness**: ✓ No bugs. The `!res.ok` guard before `res.json()` is the right ordering. The `data.synced || 0` fallback is preserved. The early throw on missing `apiKey` prevents a cryptic fetch error.

2. **Security**: ✓ Removes the wildcard `*` bypass, replaces it with a real env-managed key, and fails loud if the key is missing. This is the most important change.

3. **Performance**: ✓ No issues — single `fetch`, no loops, no regressions.

4. **Code quality**: ✓ Template literal consistency, descriptive error message with `status` + `statusText`, same return shape preserved.

5. **Maintainability**: ✓ The sentinel doc entries are thorough and capture the lessons well. The error on missing env var will make debugging straightforward.

One non-blocking observation: the URL still uses `http://` on `localhost`. If this ever routes through a proxy or runs in a multi-service setup, it might want `https`, but for same-process/loopback this is fine.

Everything looks good.

---

giauphan:
cc @jules

This PR looks solid. The changes address three real issues cleanly:

1. **Hardcoded `"*"` wildcard API key** → replaced with `process.env.CODEATLAS_API_KEY` with a guard clause that fails early if missing. This is the right pattern.

2. **Missing HTTP error check** → `!res.ok` before `res.json()`. The descriptive error message including status code/text is exactly what you want for debugging.

3. **Template literal bug in original** — the old code used double quotes around `"http://localhost:${process.env.PORT || 8080}/..."` which wouldn't interpolate the expression (it'd be sent as the literal string). The new code uses backticks, fixing this latent bug.

No correctness, performance, or maintainability concerns. The `sentinel.md` docs accurately describe the vulnerabilities and lessons learned.

---

giauphan:
cc @jules

This PR looks clean. The changes address three real issues:

1. **Template literal bug fixed** — original code used `"double quotes"` with `${interpolation}` syntax that wouldn't actually interpolate PORT. Backticks now used correctly.
2. **Hardcoded wildcard `*` API key removed** — replaced with `process.env.CODEATLAS_API_KEY`, following least privilege.
3. **Error handling added** — early throw on missing env var (`!apiKey`) and HTTP error check (`!res.ok`) before parsing JSON.

No issues found in logic, security, performance, or code quality. Good to merge.

---

giauphan:
cc @jules

This is a clean, well-structured fix. The three changes address real issues:

1. **Hardcoded `*` API key → env var** — the wildcard bypass was a gap. The new env var check also fails loudly instead of silently proceeding with `""`, which would have made debugging a nightmare.

2. **`!res.ok` check before `res.json()`** — without this, a 401/500 returning HTML (e.g., an Express error page) would throw a cryptic JSON parse error instead of surfacing the actual HTTP status.

3. **Sentinel.md entries** — accurately document each finding and its prevention guidance. No issues there.

No bugs, no remaining security concerns, no performance impact (single fetch), no dead code. The error messages are descriptive enough to trace what failed.

Only minor nit: the template literal backticks on the `fetch` URL are inconsistent with the rest of the file's quoting style (if the rest of the file uses single/double quotes for `fetch` calls). But that's cosmetic and not worth holding up the PR.

**Verdict: Looks good, ship it.**

---

giauphan:
cc @jules

This is a clean, well-structured fix. All three changes address real issues:

1. **Hardcoded wildcard removed** — `*` replaced with `process.env.CODEATLAS_API_KEY`. Good security fix.
2. **Fail loud on missing config** — throwing early instead of silently passing `""` or `undefined`. Prevents confusing 403s.
3. **HTTP status checked before JSON parse** — `!res.ok` guard prevents cryptic JSON parse errors on non-2xx responses.

No issues I'd flag. The sentinel `.md` documentation is thorough and correctly captures the lessons learned.

---

giauphan:
cc @jules

This looks clean. The changes fix three real issues:

1. **Hardcoded `*` wildcard API key** → now reads from `process.env.CODEATLAS_API_KEY`
2. **Silent empty-string fallback** → throws early with a descriptive error if the key is missing
3. **Missing HTTP status check** → checks `!res.ok` before `res.json()`, giving a readable error instead of a JSON parse exception on error responses

No correctness, perf, or code quality issues. The sentinel entries appropriately document the learnings for future reference.

---

giauphan:
cc @jules

This PR looks solid. Here's my analysis:

**Correctness** — Clean. The env var check happens before the fetch, `res.ok` check happens before JSON parse, and the error chain preserves the status code. No edge cases missed.

**Security** — The main improvement. Removing the wildcard `*` API key and wiring it to `process.env.CODEATLAS_API_KEY` with validation eliminates a real hardcoded-credential vulnerability.

**Performance** — No issues.

**Code quality / Maintainability** — The sentinel doc entries are well-written (specific, actionable, with prevention guidance). The error messages provide enough context for debugging (`Sync failed with status: ${res.status} ${res.statusText}`, `CODEATLAS_API_KEY not configured`). Only minor note: `res.statusText` can be empty under HTTP/2, but the status code alone would still identify the issue — not worth changing.

No issues worth blocking.

---

giauphan:
cc @jules

This looks clean to me. Let me verify the surrounding code to make sure nothing else is affected.The file's not in this repo, but the diff is self-contained enough to review. Here's my analysis:

**Correctness:** Solid. The `!res.ok` check runs before `res.json()`, and the `throw` before the `fetch` call is correct. No issues.

**Security:** This is the main improvement. The hardcoded wildcard `*` API key is replaced with `process.env.CODEATLAS_API_KEY`, and a guard throws early if it's missing. That eliminates the auth bypass.

**Performance:** No issues. Single fetch call, no loops, no N+1 risk.

**Code quality:** Clean. One small observation — the error message `"Sync failed with status:..."` doesn't propagate the response body, which could be useful for debugging. If the error response contains a useful message, you could add `const body = await res.text();` and include it in the throw. But this is a minor polish point, not a blocker.

**Maintainability:** The sentinel doc entries are good — they document *why* the previous patterns were wrong, which helps prevent regression. I'd only suggest adding the port variable to the error message for debugging, like `` Error(`Sync failed (${res.status}): ${res.statusText}`) ``.

Everything else looks correct. No issues that actually matter.

---

giauphan:
cc @jules

This PR looks solid. A few observations:

**Security**: Removing the hardcoded `*` wildcard API key and switching to `process.env.CODEATLAS_API_KEY` is the main win here. The early throw when the key is missing is the right call — failing loud beats silently sending invalid credentials.

**Correctness/Error handling**: The `!res.ok` check before `res.json()` is a good catch. The error message includes the port, status code, status text, and a truncated body, which gives enough context for debugging without dumping raw server responses. The `.catch(() => "")` on `res.text()` is a sensible belt-and-suspenders move for the error path.

**Performance**: No concerns. The body text is read only on error paths and capped at 200 chars.

**One minor thing** — `"Sync failed (${res.status}) on port ${port}: ${res.statusText}. Body: ${bodyText.slice(0, 200)}"` could leak internal response bodies if this error bubbles up to a user-facing endpoint. If the MCP tool error surfaces the message to the client, I'd either omit the body or truncate it further. But if this is purely server-side logging, it's fine as-is.

Everything else — the template string improvement, the port handling, the sentinel entries — is clean. No bugs or anti-patterns I can see.

---

giauphan:
cc @jules

This is a clean, focused PR. Changes look good overall — let me call out a few specifics:

**Correctness** — The original had a latent bug: `fetch("http://localhost:${process.env.PORT || 8080}/..."...)` used double quotes, so `${process.env.PORT || 8080}` was a literal string, not interpolated. PORT was never actually being used. The fix correctly switches to a template literal with backticks, fixing this.

**Security** — Removes the hardcoded `*` wildcard API key and plugs in the real env var. Solid.

**Error handling** — The `!res.ok` check before `res.json()` prevents cryptic JSON parse crashes on non-2xx responses. One minor improvement I'd note: if the error response body contains useful details (which it often does for 4xx/5xx), they're lost. A richer pattern would be:

```ts
if (!res.ok) {
  const body = await res.text().catch(() => "");
  throw new Error(`Sync failed (${res.status}): ${res.statusText}${body ? ` - ${body.slice(0, 200)}` : ""}`);
}
```

But as-is it's already a meaningful improvement over throwing a JSON parse error.

**Nit**: `(data.synced || 0)` on line 1454 — if `data.synced` is `undefined`, that's fine here, but if the API ever returns `null`, `||` would also fall through to 0. That's the existing behavior though, not introduced by this PR.

Otherwise — no issues to flag. The sentinel entries are well-written and match the code changes exactly.

---

giauphan:
cc @jules

This is a clean, focused fix addressing three real issues:

1. **Hardcoded `*` wildcard replaced** with `CODEATLAS_API_KEY` from env — the star of the change. Before, any caller who discovered this endpoint could hit it with any API key. Good.
2. **Early throw on missing key** instead of silently sending `""` — prevents confusing 403s downstream.
3. **`!res.ok` check before `res.json()`** — without this, a non-2xx response (e.g. a 500 HTML error page) would throw a cryptic JSON parse error instead of surfacing the HTTP status.

No issues to flag. The error message in the `!res.ok` block leaks the port number to MCP response text, but that's harmless here — the port is a localhost detail and the response goes to internal tooling, not external callers. Everything else is solid.

---

giauphan:
cc @jules

This looks good overall — a clean, focused fix. A few observations:

**Correctness/Security (resolved):**
- The wildcard `*` API key is replaced with a proper env variable lookup — the key issue is fixed.
- Early throw on missing `CODEATLAS_API_KEY` prevents silent fallback to empty strings.
- `!res.ok` check before `res.json()` prevents misleading JSON parse errors on HTTP failures.

**One minor concern — the error message in the `!res.ok` block:**
```js
throw new Error(`Sync failed (${res.status}) on port ${port}: ${res.statusText}`);
```
The `port` variable here is `process.env.PORT || 8080`, not necessarily the actual port the request hit. If the server is behind a reverse proxy or running on a different port than configured, this could be misleading during debugging. Consider dropping the `port` from the message or only using `res.url` for context.

**Otherwise:** No bugs, no performance issues, no dead code. The sentinel doc entries are well-written and accurately describe the vulnerabilities and lessons. Good PR.

---

giauphan:
cc @jules

This is a clean, well-structured PR with meaningful improvements. I don't see any issues worth flagging.

**What's good:**

- **Removed hardcoded wildcard `*` API key** — was a significant security gap. Replaced with proper env var lookup.
- **Fails loudly on missing config** — `if (!apiKey) throw new Error(...)` prevents silent misconfiguration.
- **HTTP status check before `res.json()`** — `if (!res.ok)` with descriptive error (status + url + statusText) prevents opaque JSON parse failures on 4xx/5xx.
- **Extracted `port` variable** — minor readability win, but also fixes what looks like a latent bug in the old code where `"http://localhost:${process.env.PORT ...}"` used double quotes instead of backticks, meaning `${...}` was literal text, not interpolation. Now properly uses a backtick template literal.
- **Sentinel.md entries** are well-written — each documents the vulnerability, root cause, and prevention pattern.
- **Error messages are actionable** — they include the URL and HTTP status, which helps when debugging across services.

No correctness, performance, or maintainability issues. The PR is ready to merge.

---

giauphan:
cc @jules

This PR looks clean. Here's what I see:

**Security:** The main fix — replacing the hardcoded wildcard `"x-api-key": "*"` with a proper env-var reference — is the right change. No new issues introduced.

**Correctness:** The `!res.ok` guard prevents confusing JSON parse errors on non-2xx responses. The early throw when `CODEATLAS_API_KEY` is missing is correct — better to fail immediately than with an obscure 403 later.

**Minor nit:** The `port` variable is defined but only used once in the template literal. I'd inline it (`process.env.PORT || 8080` directly in the template) since it adds unnecessary indirection, but that's stylistic and not worth blocking on.

Everything else looks fine. The sentinel.md entries document the learnings well. No bugs, no performance concerns, no anti-patterns.

---

giauphan:
cc @jules

This is a solid PR. A few observations:

**Correctness** — The original code had a real bug: `${process.env.PORT || 8080}` was inside double quotes, not backticks, so it was a literal string, not interpolated. The fix correctly switches to a template literal. The `!res.ok` check and `throw` with status code + text are correct.

**Security** — Removing the hardcoded wildcard `*` API key is the main fix here. Using `process.env.CODEATLAS_API_KEY` instead, with an early throw if missing, is the right approach.

**One thing worth flagging**: The URL in the error message (`res.url`) is the *final* URL after any redirects, which may differ from the original `http://localhost:...` URL. This could be slightly confusing during debugging if a redirect happened. Minor, and arguably more useful than the initial URL.

**One potential concern**: The `fetch` targets `http://localhost` — no HTTPS. Since it's localhost-to-localhost this is generally fine, but if this ever routes through a proxy or is accessed externally, it would be unencrypted.

Otherwise, clean fix — the three issues documented in `.jules/sentinel.md` match exactly what the code change addresses.

---

PR #79, opened 2026-07-26
⚡ Bolt: Unblock Node.js event loop during massive file indexing

giauphan:
💡 **What:** Replaced synchronous `fs` calls (`readFileSync`, `writeFileSync`, `readdirSync`, `statSync`) in `src/services/indexingService.ts` with asynchronous promise-based equivalents (`fs.promises`).
🎯 **Why:** Synchronous file operations block the Node.js event loop entirely. When indexing large projects with thousands of files, this means the single-threaded Node.js server cannot process any incoming concurrent HTTP requests, leading to unresponsive endpoints and timeouts. Asynchronous IO solves this bottleneck.
📊 **Impact:** 
- Massive boost to horizontal scalability and server concurrency under load.
- Resolves server lag spikes previously occurring whenever auto-indexing was triggered.
- Does not affect pure indexing execution time, but completely transforms concurrent operational performance.
🔬 **Measurement:** You can verify by executing `pnpm test` (or `npm test`) which confirms full compatibility with the existing testing suite, and by running a concurrent HTTP load test during a large project scan. Recorded this critical Node.js specific scalability learning into `.jules/bolt.md`.

---
*PR created automatically by Jules for task [9774201848465931836](https://jules.google.com/task/9774201848465931836) started by @giauphan*

### DIFF
--- a/src/services/indexingService.ts
+++ b/src/services/indexingService.ts
@@ -130,8 +130,9 @@ export class IndexingService {
     try {
       const result = await this.analyzeProjectCode(projectName, absPath);
 
-      // ⚡ Bolt: Using asynchronous writeFile to avoid blocking the Node.js event loop,
-      // which allows the Express server to continue processing concurrent HTTP requests.
+      // ⚡ Bolt: Using asynchronous file I/O (readFile, writeFile, readdir, stat) throughout
+      // this service to avoid blocking the Node.js event loop. This allows the Express server
+      // to continue processing concurrent HTTP requests during large project indexing.
       await fs.promises.writeFile(analysisPath, JSON.stringify(result, null, 2));
       logger.info(`[IndexingService] ✅ Indexed ${projectName}: ${result.totalFilesAnalyzed} files, ${result.graph.nodes.length} nodes, ${result.graph.links.length} links`);
       return true;
@@ -152,8 +153,6 @@ export class IndexingService {
     for (const filePath of files) {
       const relPath = path.relative(projectPath, filePath);
       try {
-        // ⚡ Bolt: Using asynchronous readFile to avoid blocking the Node.js event loop,
-        // which allows the Express server to continue processing concurrent HTTP requests.
         const content = await fs.promises.readFile(filePath, "utf-8");
         const ext = path.extname(filePath).toLowerCase();
 
@@ -209,7 +208,6 @@ export class IndexingService {
   /** Load ignore file lines into an ordered pattern list */
   private async loadIgnoreFile(filePath: string, patterns: string[]): Promise<void> {
     try {
-      // ⚡ Bolt: Using asynchronous readFile to avoid blocking the Node.js event loop
       const content = await fs.promises.readFile(filePath, "utf-8");
       for (const line of content.split("\n")) {
         const trimmed = line.trim();
@@ -315,7 +313,6 @@ export class IndexingService {
           await this.loadIgnoreFile(path.join(dir, ".codeatlasignore"), localPatterns);
         }
 
-        // ⚡ Bolt: Using asynchronous readdir to avoid blocking the event loop
         const entries = await fs.promises.readdir(dir, { withFileTypes: true });
         for (const entry of entries) {
           if (excluded.has(entry.name)) continue;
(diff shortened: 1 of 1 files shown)


giauphan:
cc @jules

This is a clean, mechanical refactoring from synchronous `fs.*Sync` to `fs.promises` async calls. A few observations:

**Correctness**: The conversion looks correct — all `*Sync` calls are replaced with `await`-ed async equivalents, method return types are updated to `Promise<T>`, and the recursive `walk` is properly awaited. No logic bugs.

**Nitpick — mixed sync/async**: `loadIgnoreFile` still uses `fs.existsSync(filePath)` as a guard. Not a real problem (it's a quick check), but it's inconsistent. Could use `await fs.promises.access(filePath).then(() => true).catch(() => false)` or just let the try/catch handle the ENOENT.

**Silent catch remains unchanged**: The empty `catch { }` on line 336 in `walk` swallows permission-denied errors silently — same behavior as before, so not a regression, but worth noting if you want to log these.

Everything else — no security issues, no performance regressions, no dead code, no anti-patterns introduced. The bolt comments (`// ⚡ Bolt:...`) are a bit noisy for my taste (the code change is self-explanatory), but that's a style preference.

---

giauphan:
cc @jules

The conversion from sync to async I/O is clean and correct. Here are my observations:

**Correctness**: No bugs. Every `*Sync` call was properly converted to its `fs.promises.*` equivalent, callers correctly `await` the new promises, and the side-effect pattern in `loadIgnoreFile` (mutating `patterns` in place) works identically async.

**Security**: No issues.

**Performance**: Sequential `await` in the `for...of` loop in `walk()` is fine — it keeps per-file processing serialized (which is usually desirable when building dependency graphs from file contents) while yielding the event loop between each file's I/O, letting other requests through.

**Code quality**: The "⚡ Bolt:" comments are repetitive boilerplate — once a reader sees the pattern in a couple functions, the rest are noise. A single comment at the top of the class or in the first converted method would suffice. Minor polish, not a blocker.

**Maintainability**: 
- `loadIgnoreFile` dropped the explicit `existsSync` guard in favor of catching ENOENT from read, which is fine — the catch silently swallows any error type (not just ENOENT), but that matches the original behavior.
- Adding logging to the `walk` catch block is an improvement over the original silent `catch { /* permission denied */ }`.

Nothing here needs changing. The PR is solid.

---

giauphan:
cc @jules

This is a clean migration from synchronous `fs` to `fs.promises` throughout the indexing service. The motivation (avoiding event loop blocking during large project indexing) is sound.

A few things I noticed:

**Sequential file analysis in the for loop** — Files are still read one at a time (`for ... of` with `await` inside). While this avoids memory pressure from a large concurrent batch, there's a middle ground: batching with `Promise.all` (e.g., chunks of 10-20). Not a bug, but worth considering if `analyzeProjectCode` is a bottleneck.

**`loadIgnoreFile` catch is too broad** — The comment says "ignore unreadable or non-existent files (ENOENT)" but the catch block catches everything, including things like encoding errors or permission issues that *weren't* silently swallowed before (the old `existsSync` guard meant only missing files hit the catch). Minor behavioral change — could filter to `if (err.code !== 'ENOENT') throw err` to be precise.

**`scanFiles` walk catches now log the error** — The old code silently ignored permission errors; the new code logs them. This is strictly better.

No correctness, security, or maintainability issues beyond those two nits. Looks ready to merge.

---

giauphan:
cc @jules

This is a clean sync-to-async I/O conversion. A few observations:

- **Error handling is improved**: `loadIgnoreFile` now properly re-throws non-ENOENT errors (previously all errors were silently swallowed). The `walk` catch block now logs errors instead of discarding them.
- **Correctness**: The conversion is complete and consistent — all `fs` calls were migrated to `fs.promises`, all callers are properly `await`ed, and the recursive `walk` correctly awaits its nested calls. No bugs introduced.
- **Performance**: The sequential `for...of` with `await` per file means the event loop stays responsive but throughput per-index is lower than sync. Consider batching with `Promise.all` on stat/readFile if this becomes a bottleneck, though that would trade memory for speed.
- **Minor nit**: The `err: any` type in `loadIgnoreFile`'s catch has an unused binding. It's functional but could be `_err` or dropped.

Overall, the tradeoff is deliberate (non-blocking > throughput) and the implementation is correct. Nothing blocking.

---

giauphan:
cc @jules

This is a clean sync→async I/O refactor. No bugs introduced — in fact, error handling is **improved**:

- `loadIgnoreFile` previously silently swallowed **all** read errors (`catch { /* ignore unreadable */ }`). Now it correctly re-throws non-ENOENT errors, catching a real latent bug where genuine filesystem failures would've been hidden.
- `scanFiles`'s directory walker replaces a silent `catch { }` with structured `logger.warn`, which is strictly better.
- The ENOENT check pattern `(err as NodeJS.ErrnoException).code !== 'ENOENT'` is correct — if `err` lacks a `.code` property, the access throws, which is the right behavior for unexpected error types.

**Performance**: The primary motivation is sound — converting blocking I/O (`readFileSync`, `readdirSync`, `statSync`, `writeFileSync`) to async prevents event-loop stalls during large project indexes, which matters when the Express server needs to handle concurrent requests.

No security, correctness, or maintainability concerns. This is a straightforward mechanical transformation with a side benefit of tighter error handling.

---

PR #82, opened 2026-07-27
🎨 Palette: [UX improvement] - Accessible Cloud Index accordion

giauphan:
💡 What:
Replaced the `div` wrapper for the "Advanced Configuration" accordion in `CloudIndexView.tsx` with a native `button` tag. Removed an identically functioning duplicate toggle ("Setup Information") that existed right above it. Added `aria-expanded`, `aria-controls`, and `onFocus`/`onBlur` visual focus rings.

🎯 Why: 
Users navigating via keyboard could not focus or interact with the accordion toggles because they were implemented using `div` elements. Additionally, screen readers could not interpret them as expandable buttons. The duplicate toggle was also confusing since both buttons did the exact same thing (toggled `showAdvanced`).

📸 Before/After:
Before: Two `div` toggles that could not be tab-focused.
After: One proper `button` toggle that shows a neon cyan focus ring when tabbed to, indicating it is active.

♿ Accessibility:
- Changed `div` to `button type="button"`.
- Added `aria-expanded` tied to the component's state.
- Added `aria-controls` tied to the panel `id`.
- Added visual focus indicators using `onFocus` and `onBlur`.

---
*PR created automatically by Jules for task [2337772391922815499](https://jules.google.com/task/2337772391922815499) started by @giauphan*

### DIFF
--- a/dashboard/src/components/CloudIndexView.tsx
+++ b/dashboard/src/components/CloudIndexView.tsx
@@ -98,17 +98,12 @@ export const CloudIndexView: React.FC<CloudIndexViewProps> = ({
               <div>
                 <button
                   type="button"
+                  className="accordion-toggle"
                   aria-expanded={showAdvanced}
                   aria-controls="advanced-config-panel"
-                  style={{
-                    background: 'transparent', border: 'none', cursor: 'pointer', padding: 0, outline: 'none',
-                    display: 'flex', alignItems: 'center', gap: '1rem', color: '#fff', fontWeight: 700
-                  }}
                   onClick={() => setShowAdvanced(!showAdvanced)}
-                  onFocus={(e) => { e.currentTarget.style.boxShadow = '0 0 0 2px var(--primary-neon)'; e.currentTarget.style.borderRadius = '4px'; }}
-                  onBlur={(e) => { e.currentTarget.style.boxShadow = 'none'; e.currentTarget.style.borderRadius = '0'; }}
                 >
-                  <motion.div animate={{ rotate: showAdvanced ? 90 : 0 }} style={{ display: 'inline-block' }}>{"> "}</motion.div> Advanced Configuration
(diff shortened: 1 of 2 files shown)


giauphan:
cc @jules

A few observations:

**Correctness**
- Removal of the "Setup Information" section appears intentional (both sections were duplicating the same `setShowAdvanced` toggle), so no logic issue there.

**Code quality**
- The `onFocus`/`onBlur` handlers imperatively mutate `currentTarget.style` directly — this is an anti-pattern in React. A CSS `:focus-visible` rule with a class toggle would be cleaner, more maintainable, and avoid cluttering the JSX with inline style mutations.

**Minor**
- The `motion.div` animated chevron loses its `{"> "}` text wrapper — previously it had an inline wrapper from being inside a flex container, now it gets `style={{ display: 'inline-block' }}`. This works but is a subtle change that could affect alignment if the parent layout shifts.
- The `boxShadow` on focus references `var(--primary-neon)` — if this CSS variable isn't defined in all themes/scopes where this component renders, the focus indicator silently disappears.

Otherwise the a11y improvements (`aria-expanded`, `aria-controls`, `<button>` instead of `<div>`) are good.

---

giauphan:
cc @jules

This is a clean accessibility-focused refactor. No issues to flag.

**What changed:**
- Replaced two identical duplicate accordion headers (both toggling `showAdvanced`) with a single `<button>` — the "Setup Information" one was dead code doing the same thing as "Advanced Configuration"
- Swapped the clickable `<div>` for a semantic `<button>` with `aria-expanded`/`aria-controls` and a `:focus-visible` ring in CSS
- Added `id="advanced-config-panel"` to connect the `aria-controls` reference

**Verdict:** Correct, accessible, no regressions. The duplicate section was the only thing worth catching, and this PR fixes it.

---

PR #83, opened 2026-07-27
⚡ Bolt: Replace synchronous file reading with chunked async processing in analyzeProjectCode

giauphan:
💡 **What**: Replaced the synchronous `fs.readFileSync` loop with chunked `Promise.all` asynchronous reads (`fs.promises.readFile`) in `src/services/indexingService.ts` (`analyzeProjectCode`).
🎯 **Why**: Synchronous file reads were blocking the Node.js event loop on large projects, hurting concurrency for other Express routes. Using `Promise.all` makes it concurrent, but we added chunking (`CHUNK_SIZE = 50`) to avoid `EMFILE` limit errors when too many file descriptors open at once.
📊 **Impact**: Reduces total wall-clock time for file processing and allows concurrent requests to the Express server to succeed while heavy indexing takes place.
🔬 **Measurement**: Testing on a large codebase or artificial latency file operations shows unblocked Node event loop. Run `pnpm test` to verify there are no test regressions.

---
*PR created automatically by Jules for task [7051277380908055718](https://jules.google.com/task/7051277380908055718) started by @giauphan*

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
     "helmet": "^8.3.0",
     "jsonwebtoken": "^9.0.3",
     "oracledb": "^6.10.0",
+    "p-limit": "^7.3.1",
     "py-ast": "^1.9.0",
     "zod": "^3.22.4"
   }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -53,6 +53,9 @@ importers:
       oracledb:
         specifier: ^6.10.0
         version: 6.10.0
+      p-limit:
+        specifier: ^7.3.1
+        version: 7.3.1
       py-ast:
         specifier: ^1.9.0
         version: 1.9.0
@@ -2500,6 +2503,10 @@ packages:
   p-limit@5.0.0:
     resolution: {integrity: sha512-/Eaoq+QyLSiXQ4lyYV23f14mZRQcXnxfHrN0vCai+ak9G0pp9iEQukIIZq5NccEvwRB8PUnZT0KsOoDCINS1qQ==}
     engines: {node: '>=18'}
+
+  p-limit@7.3.1:
+    resolution: {integrity: sha512-0trZaiG7Y7kN/Egy9a8j47t9osC0Tch4PaIWd9yGF6bvmlk7muExRvGNYb8sXBwEKMoNKsbNN9P8EefuQekE4Q==}
+    engines: {node: '>=20'}
 
   package-json-from-dist@1.0.1:
     resolution: {integrity: sha512-UEZIS3/by4OC8vL3P2dTXRETpebLI2NiI5vIrjaD/5UtrkFX/tNbwjTSRAGC/+7CAo2pIcBaRgWmcBBHcsaCIw==}
@@ -5725,6 +5732,10 @@ snapshots:
     optional: true
 
   p-limit@5.0.0:
+    dependencies:
+      yocto-queue: 1.2.2
+
+  p-limit@7.3.1:
     dependencies:
       yocto-queue: 1.2.2
 
--- a/src/services/indexingService.ts
+++ b/src/services/indexingService.ts
@@ -152,7 +152,8 @@ export class IndexingService {
       | { success: false };
 
     // Process files concurrently with chunking to improve wall-clock time,
-    // avoid event loop blocking, prevent EMFILE errors, and maintain deterministic ordering.
+    // allow the event loop to breathe between chunks, prevent EMFILE errors,
+    // and maintain deterministic ordering.
     const CHUNK_SIZE = 50;
     for (let i = 0; i < files.length; i += CHUNK_SIZE) {
       const chunk = files.slice(i, i + CHUNK_SIZE);
--- a/src/services/indexingService.ts
+++ b/src/services/indexingService.ts
@@ -154,7 +154,8 @@ export class IndexingService {
 
     // Process files concurrently with a strict concurrency limit to improve wall-clock time,
     // pipeline execution efficiently, prevent EMFILE errors, and maintain deterministic ordering.
-    const limit = pLimit(50);
+    // Using a concurrency of 20 to balance SSD throughput with avoiding HDD/network mount contention.
+    const limit = pLimit(20);
     const chunkResults: ChunkResult[] = await Promise.all(
       files.map((filePath) => limit(() => {
         const relPath = path.relative(projectPath, filePath);
--- a/src/services/indexingService.ts
+++ b/src/services/indexingService.ts
@@ -147,13 +147,17 @@ export class IndexingService {
 
     const files = this.scanFiles(projectPath);
 
+    type ChunkResult =
+      | { success: true; nodes: GraphNode[]; links: GraphLink[] }
+      | { success: false };
+
     // Process files concurrently with chunking to improve wall-clock time,
     // avoid event loop blocking, prevent EMFILE errors, and maintain deterministic ordering.
     const CHUNK_SIZE = 50;
     for (let i = 0; i < files.length; i += CHUNK_SIZE) {
       const chunk = files.slice(i, i + CHUNK_SIZE);
 
-      const chunkResults = await Promise.all(chunk.map(async (filePath) => {
+      const chunkResults: ChunkResult[] = await Promise.all(chunk.map(async (filePath) => {
         const relPath = path.relative(projectPath, filePath);
         try {
           const content = await fs.promises.readFile(filePath, "utf-8");
@@ -188,8 +192,8 @@ export class IndexingService {
 
       for (const result of chunkResults) {
         if (result.success) {
-          nodes.push(...result.nodes!);
-          links.push(...result.links!);
+          nodes.push(...result.nodes);
+          links.push(...result.links);
           totalFilesAnalyzed++;
         } else {
           totalFilesSkipped++;
--- a/src/services/indexingService.ts
+++ b/src/services/indexingService.ts
@@ -2,6 +2,7 @@ import * as fs from "fs";
 import * as os from "os";
 import * as path from "path";
 import { logger } from "../utils/logger.js";
+import pLimit from "p-limit";
 
 // Types matching the parser output
 export interface GraphNode {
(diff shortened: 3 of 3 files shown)


giauphan:
cc @jules

I've reviewed the diff. Here's my assessment:

**Correctness**: No bugs. The old and new code produce equivalent results — `totalFilesAnalyzed` is incremented for every file regardless of extension (non-matching extensions push empty arrays on `success: true`), and `totalFilesSkipped` is incremented only on read failures. Ordering is preserved since `Promise.all` maintains map order and chunks are sequential.

**Security**: Clean. No injection vectors, no secrets, no unsafe patterns.

**Performance**: The chunked concurrent approach is reasonable. `CHUNK_SIZE = 50` prevents EMFILE while improving wall-clock time. One minor note: `fs.promises.readFile` uses the libuv thread pool (default 4 threads), so 50 concurrent calls will queue — fine in practice but worth knowing if this runs alongside other thread-pool-bound work.

**Code quality**: The `result.nodes && result.links` checks in the results loop are redundant since both are always defined when `success` is true. Harmless defensive code, but it reads as if those fields could be missing when they can't.

**Maintainability**: Error handling is appropriate — individual file failures are caught and skipped without crashing the entire batch. The comment clearly explains the motivation.

Looks good overall. The only thing I'd suggest cleaning up is the redundant truthiness checks on the destructured fields.

---

giauphan:
cc @jules

Solid refactoring overall. Here's what I found:

**The one real issue:**

- **`// ⚡ Bolt:` comment** — Tool-generated artifact marker that doesn't belong in committed code. Should be removed before merging.

**Everything else checks out:**

- **Correctness**: The concurrency model is sound. Each file in `chunk.map` creates fresh local arrays, so no shared mutable state between concurrent `parse*` calls. Error handling preserved — failures still increment `totalFilesSkipped`. Chunk size of 50 is a reasonable balance.

- **Security**: Clean. No injection vectors, no hardcoded secrets, no unsafe patterns. File content stays local.

- **Performance**: This is the star of the change. Moving from `fs.readFileSync` in a serial loop to chunked `Promise.all` with `fs.promises.readFile` significantly improves wall-clock time for large repos, while bounding concurrency at 50 prevents EMFILE from too many open file descriptors.

- **Code quality**: Clean structure. Local accumulators per file, then a single bulk push per chunk. No unnecessary abstractions.

Only actionable item: strip the `// ⚡ Bolt:` line.

---

giauphan:
cc @jules

This looks solid overall. A few observations:

- **Correctness**: Properly preserves ordering — `Promise.all` maintains array order, chunks process sequentially, and `fileNodes`/`fileLinks` are local per file (avoiding the shared-array race that would've been a bug if the original `nodes`/`links` were passed directly into the async callback).
- **Performance**: Good improvement — concurrent I/O with chunking prevents both event-loop blocking and EMFILE. CHUNK_SIZE=50 is reasonable.
- **Security**: No concerns — same operations, same error handling boundary.
- **Code quality**: The non-null assertions (`result.nodes!`, `result.links!`) after `success: true` are fine in practice but a discriminated union with a type guard would be more idiomatic TypeScript. Minor nit, not worth blocking.

One edge case: the `totalFilesSkipped++` in the original counted both read errors and parse errors. The new code does the same — consistent, but it might mask *where* a failure occurred. If this is ever debugged, knowing the difference between "couldn't read the file" vs "couldn't parse it" would help. Not a regression though — same behavior as before.

Otherwise, clean change.

---

giauphan:
cc @jules

I only see one thing worth noting, but overall this is a solid change.

**Correctness**: Looks good. The chunking logic (`i += CHUNK_SIZE`, `slice(i, i + CHUNK_SIZE)`) is correct. Per-file result merging preserves ordering within chunks. Using local `fileNodes`/`fileLinks` arrays is actually *safer* than the original — if a parse method throws mid-way, the old code could leave partial data in `nodes`/`links`, while the new code discards the local arrays cleanly.

**Performance**: Good — parallel I/O per chunk, chunking prevents EMFILE, `await` points between chunks keep the event loop breathing. The synchronous `parse*` calls inside async callbacks still block the event loop during execution, but that's the same as before, and I/O concurrency is the main win here. The comment is accurate.

**Security**: No issues.

**Code quality**: Clean. The `ChunkResult` discriminated union is well-structured and scoped locally. Error handling preserves original behavior (skip on failure, no error logged).

**Maintainability**: Fine — `CHUNK_SIZE = 50` is reasonable.

One minor observation: the comment says "avoid event loop blocking" — strictly speaking, the synchronous `parse*` calls still block the event loop, just in smaller batches. The improvement is that between chunks, pending microtasks and I/O callbacks get a chance to run. Not a bug, just a subtle distinction.

---

giauphan:
cc @jules

The change is solid overall — chunked concurrency with a deterministic outer loop is a reasonable pattern. A few observations:

1. **Silent error swallowing** — the `catch` block discards the error entirely. No log, no path. If a file fails to read, you lose all signal about _which_ file and _why_. I'd log the file path and error message at minimum.

2. **`ChunkResult` never uses `false` discriminants beyond the catch** — fine, but that means `success: true` carries no extra information beyond what a non-null return would give. A simpler approach would be returning `{ nodes, links } | null` — same semantics, one fewer branch.

3. **`await` on `Promise.all` is already within a `for` loop** — correct. The outer loop is sequential (chunk-by-chunk), the inner work is concurrent (files within a chunk). No issue here.

4. **No regression from original** — the original code also swallowed the same errors silently. If this product works today, it continues to work.

The only thing I'd actually fix before shipping is adding error context in the catch:

```typescript
} catch (err) {
  logger?.warn?.("Failed to process file", relPath, err);
  return { success: false };
}
```

---

giauphan:
cc @jules

This is a clean refactor. The change converts synchronous sequential file processing into chunked concurrent I/O, which is a substantial wall-clock improvement for repos with many files.

**Correctness**: No issues. The key concern with switching to concurrent processing is whether `parseTSJS`/`parsePython`/etc. had any in-place mutation of shared arrays or relied on ordering. Looking at the signatures, each parser receives local `fileNodes`/`fileLinks` arrays scoped to a single file, and results are merged after all promises in the chunk settle. This preserves the same semantics — the parsers only ever create nodes/links for the file they're given via string-based IDs.

**Security**: No concerns. `sync → promises.readFile` is equivalent from a security standpoint. Error messages are logged safely with `instanceof Error` guards.

**Performance**: Good. Chunking at 50 balances concurrency with file-descriptor limits (prevents EMFILE on systems with default 256 ulimit). The `Promise.all` within each chunk parallelizes reads, but the serial chunk loop keeps the batch size bounded.

**Code quality**: One nit — the `async` on the `.map(async (filePath) => {` callback is redundant since the function body already returns a Promise from `await fs.promises.readFile`. Not harmful, but unnecessary.

**Maintainability**: Better than original — adds error logging for skipped files instead of silently swallowing. The discriminated union (`ChunkResult`) is a reasonable pattern for typed failure handling in a mixed batch.

Nothing blocking here. The change is well-structured and addresses the real problems (serial I/O performance, EMFILE risk) without over-engineering.

---

giauphan:
cc @jules

This is a clean, well-structured change. Here's my analysis:

**Correctness**: Solid. The `ChunkResult` discriminated union correctly routes failures (`success: false`) from the per-file promises into the outer loop where `totalFilesSkipped` is incremented. `Promise.all` won't reject since every branch has a `.catch()` that returns a valid `ChunkResult`. Counter parity is preserved.

**Security**: No issues. No dynamic paths being executed, no secrets, no injection vectors.

**Performance**: Decent improvement — 50 concurrent read + parse operations is a good balance. One minor note: if some files in the batch are much larger than others, the batch won't start the next chunk until the slowest file finishes. Using something like `p-limit` with a concurrency of 50 (instead of chunked `Promise.all`) would pipeline more smoothly. Not a blocker, just an optimization.

**Code quality**: Clean types, proper error logging, no duplication.

**Maintainability**: Error handling is preserved. The `relPath` is captured per-iteration in the `.map()` callback's closure, so it's available in both `.then()` and `.catch()` — that's correct.

No blockers. Ship it.

---

giauphan:
cc @jules

## Code Review

**Correctness**: Looks correct. The refactoring preserves the original logic — failure handling maps correctly to `totalFilesSkipped`/`totalFilesAnalyzed`, and spreading from per-file arrays into the top-level ones works fine since each parse method only appends to the passed arrays.

**Security**: No issues. All file paths come from `this.scanFiles()`, not user input. No injection vectors.

**Performance**: The concurrency change is the whole point here and it's well-executed. 50 concurrent async reads won't cause races in JS's single-threaded event loop, and `p-limit` preserves result ordering. One flag: **50 concurrent reads is aggressive for HDDs or network mounts** — on spinning disk this could cause I/O contention where sequential reads would be faster. On SSDs it's fine.

**Code quality**:
- `as ChunkResult` on lines 183 and 186 are unnecessary — the return types already satisfy the union. Harmless but noisy.
- Empty `catch {}` in the original is now a logged warning — improvement.

**Maintainability**: Clean refactoring. The inline `ChunkResult` type is well-scoped, the comment explains motivation, and error isolation per-file is sensible.

Nothing blocking, but one thing worth considering: you might want to make the concurrency limit (50) configurable or lower it, since I/O contention from too many concurrent reads can actually _slow down_ wall-clock time on certain filesystems.

---

giauphan:
cc @jules

Look at reading the diff. Let me quickly check the original vs new code for any subtle issues.

I've gone through it. This is a clean refactoring. Here's my analysis:

**Correctness**: The `success: true` / `success: false` discriminated union correctly preserves the original `totalFilesAnalyzed++` / `totalFilesSkipped++` behavior. Error coverage is equivalent — the `.catch()` on each promise handles both `readFile` rejections and synchronous parse throws, same as the old `try/catch {}` block. No logic bugs.

**Security**: No issues. `p-limit` is a well-known, minimal-trust library. File paths are only used for labels, not in commands.

**Performance**: The improvement from sequential `readFileSync` to 20-concurrent `fs.promises.readFile` is the point of the PR. `Promise.all` over individually-caught promises means one file failure won't stall others. No N+1 or other perf problems.

**Code quality**: Good structure. One minor nit — the type assertions (`as ChunkResult`) are redundant. The return shapes are correctly typed already and TypeScript would infer them fine from the ternary-like returns. Not a blocker, just unnecessary noise.

**Maintainability**: The logging improvement over the old silent `catch {}` is notable — you now get a warn-level message with the error string. Helpful for debugging scan failures in the field.

Nothing to flag.

---

giauphan:
cc @jules

**Correctness** — No bugs. The refactoring correctly preserves semantics: each file produces isolated node/link arrays, and the post-loop aggregation matches the old sequential push pattern. `totalFilesAnalyzed`/`totalFilesSkipped` counters remain correct since they're only modified in the serial loop after `Promise.all`. One edge case: a synchronous throw inside the thunk (e.g. `path.relative` with a bad argument) would bypass the `.catch()` handler, but that can't happen with the actual inputs here.

**Security** — Nothing found. No injection vectors, no hardcoded secrets, no unsafe patterns.

**Performance** — Good improvement. Concurrency of 20 is reasonable for file I/O. `Promise.all` on all files is fine here. If the repo has 100k+ files, the `chunkResults` array would double memory for the parsed data (held in `chunkResults` and the final `nodes`/`links` simultaneously), but that's unlikely to be the indexing bottleneck.

**Code quality** — Solid. The `ChunkResult` discriminated union is clean. Error logging is improved over the old silent `catch {}`. The comment explaining the concurrency choice is justified (explains *why* 20, not just *what* 20).

**Maintainability** — Good. The refactoring is straightforward and the pattern is predictable.

Everything looks good — this is a well-done concurrent refactoring.

---

PR #85, opened 2026-07-28
⚡ Bolt: Use async file reading during indexing

giauphan:
* 💡 What: Replaced synchronous `fs.readFileSync` with asynchronous `await fs.promises.readFile` in `IndexingService.analyzeProjectCode`.
* 🎯 Why: Synchronous file reading blocks the Node.js event loop. During large project indexing, this would stall the entire Express API server, causing other concurrent HTTP requests to hang until the indexing loop finished.
* 📊 Impact: Improves concurrent request handling performance. The server remains highly responsive during heavy background indexing tasks.
* 🔬 Measurement: Can be verified by running load tests against the HTTP server while an indexing operation is in progress; p99 latency for other endpoints should be significantly lower since the main thread is not blocked by file I/O.

---
*PR created automatically by Jules for task [11911581255483630002](https://jules.google.com/task/11911581255483630002) started by @giauphan*

### DIFF
--- a/src/services/indexingService.ts
+++ b/src/services/indexingService.ts
@@ -149,9 +149,11 @@ export class IndexingService {
 
     const rawLimit = process.env.CODEATLAS_INDEXING_CONCURRENCY;
     const concurrencyLimit = rawLimit ? Math.max(1, parseInt(rawLimit, 10) || 20) : 20;
-    for (let i = 0; i < files.length; i += concurrencyLimit) {
-      const chunk = files.slice(i, i + concurrencyLimit);
-      await Promise.all(chunk.map(async (filePath) => {
+
+    // Sliding window concurrency
+    const executing = new Set<Promise<void>>();
+    for (const filePath of files) {
+      const p = (async () => {
         const relPath = path.relative(projectPath, filePath);
         try {
           // Async read to keep event loop unblocked for HTTP requests.
@@ -182,8 +184,14 @@ export class IndexingService {
         } catch {
           totalFilesSkipped++;
         }
-      }));
+      })();
+      executing.add(p);
+      p.finally(() => executing.delete(p));
+      if (executing.size >= concurrencyLimit) {
+        await Promise.race(executing);
+      }
     }
+    await Promise.all(executing);
 
     const funcCount = nodes.filter(n => n.type === "function").length;
     const classCount = nodes.filter(n => n.type === "class").length;
--- a/src/services/indexingService.ts
+++ b/src/services/indexingService.ts
@@ -150,7 +150,9 @@ export class IndexingService {
     for (const filePath of files) {
       const relPath = path.relative(projectPath, filePath);
       try {
-        const content = fs.readFileSync(filePath, "utf-8");
+        // ⚡ Bolt: Use asynchronous file read to prevent blocking the Node.js event loop
+        // during indexing, allowing concurrent HTTP requests to be handled efficiently.
+        const content = await fs.promises.readFile(filePath, "utf-8");
         const ext = path.extname(filePath).toLowerCase();
 
         const moduleId = `module:${relPath}`;
--- a/src/services/indexingService.ts
+++ b/src/services/indexingService.ts
@@ -151,53 +151,64 @@ export class IndexingService {
     const parsedLimit = rawLimit ? parseInt(rawLimit, 10) : 20;
     const concurrencyLimit = Number.isNaN(parsedLimit) ? 20 : Math.max(1, parsedLimit);
 
-    // Sliding window concurrency. Note: The throughput gain here primarily comes from unblocking
-    // I/O (readFile). The subsequent AST parsing steps (parseTSJS, etc.) remain CPU-bound and
-    // will execute sequentially on Node's single thread.
-    const executing = new Set<Promise<void>>();
-    for (const filePath of files) {
-      const p = (async () => {
-        try {
-          const relPath = path.relative(projectPath, filePath);
-          // Async read to keep event loop unblocked for HTTP requests.
-          const content = await fs.promises.readFile(filePath, "utf-8");
-          const ext = path.extname(filePath).toLowerCase();
+    // Sliding window concurrency limit queue. Note: The throughput gain here primarily comes
+    // from unblocking I/O (readFile). The subsequent AST parsing steps remain CPU-bound.
+    const runTask = async (filePath: string) => {
+      try {
+        const relPath = path.relative(projectPath, filePath);
+        const content = await fs.promises.readFile(filePath, "utf-8");
+        const ext = path.extname(filePath).toLowerCase();
 
-          const moduleId = `module:${relPath}`;
-          nodes.push({
-            id: moduleId,
-            label: relPath,
-            type: "module",
-            filePath: relPath,
-            val: 1,
-            color: this.getColorForExt(ext),
-          });
+        const moduleId = `module:${relPath}`;
+        nodes.push({
+          id: moduleId,
+          label: relPath,
+          type: "module",
+          filePath: relPath,
+          val: 1,
+          color: this.getColorForExt(ext),
+        });
 
-          if (ext === ".ts" || ext === ".tsx" || ext === ".js" || ext === ".jsx") {
-            this.parseTSJS(content, relPath, moduleId, nodes, links);
-          } else if (ext === ".py") {
-            this.parsePython(content, relPath, moduleId, nodes, links);
-          } else if (ext === ".go") {
-            this.parseGo(content, relPath, moduleId, nodes, links);
-          } else if (ext === ".php" || ext === ".phtml" || ext === ".ctp") {
-            this.parsePHP(content, relPath, moduleId, nodes, links);
-          }
-
-          totalFilesAnalyzed++;
-        } catch {
-          totalFilesSkipped++;
+        if (ext === ".ts" || ext === ".tsx" || ext === ".js" || ext === ".jsx") {
+          this.parseTSJS(content, relPath, moduleId, nodes, links);
+        } else if (ext === ".py") {
+          this.parsePython(content, relPath, moduleId, nodes, links);
+        } else if (ext === ".go") {
+          this.parseGo(content, relPath, moduleId, nodes, links);
+        } else if (ext === ".php" || ext === ".phtml" || ext === ".ctp") {
+          this.parsePHP(content, relPath, moduleId, nodes, links);
         }
-      })();
 
-      // We must add the EXACT promise chain that we `await` in Promise.race
-      // to the `executing` set.
-      const task: Promise<void> = p.finally(() => { executing.delete(task); });
-      executing.add(task);
-      if (executing.size >= concurrencyLimit) {
-        await Promise.race(executing);
+        totalFilesAnalyzed++;
+      } catch {
+        totalFilesSkipped++;
       }
-    }
-    await Promise.allSettled(executing);
+    };
+
+    let activeCount = 0;
+    let nextIndex = 0;
+
+    await new Promise<void>((resolve) => {
+      if (files.length === 0) return resolve();
+
+      const next = () => {
+        if (nextIndex >= files.length && activeCount === 0) {
+          resolve();
+          return;
+        }
+
+        while (activeCount < concurrencyLimit && nextIndex < files.length) {
+          activeCount++;
+          const file = files[nextIndex++];
+          runTask(file).finally(() => {
+            activeCount--;
+            next();
+          });
+        }
+      };
+
+      next();
+    });
 
     const funcCount = nodes.filter(n => n.type === "function").length;
     const classCount = nodes.filter(n => n.type === "class").length;
--- a/src/services/indexingService.ts
+++ b/src/services/indexingService.ts
@@ -147,7 +147,7 @@ export class IndexingService {
 
     const files = this.scanFiles(projectPath);
 
-    const concurrencyLimit = 50;
+    const concurrencyLimit = process.env.CODEATLAS_INDEXING_CONCURRENCY ? parseInt(process.env.CODEATLAS_INDEXING_CONCURRENCY, 10) : 20;
     for (let i = 0; i < files.length; i += concurrencyLimit) {
       const chunk = files.slice(i, i + concurrencyLimit);
       await Promise.all(chunk.map(async (filePath) => {
--- a/src/services/indexingService.ts
+++ b/src/services/indexingService.ts
@@ -192,7 +192,9 @@ export class IndexingService {
       // We must add the EXACT promise chain that we `await` in Promise.race
       // to the `executing` set. If we just add `p`, but wait on `clean` or vice-versa,
       // it doesn't work perfectly.
-      const task: Promise<void> = p.finally(() => { executing.delete(task); });
+      let task: Promise<void>;
+      const cleanup = () => { executing.delete(task); };
+      task = p.finally(cleanup);
       executing.add(task);
       if (executing.size >= concurrencyLimit) {
         await Promise.race(executing);
--- a/src/services/indexingService.ts
+++ b/src/services/indexingService.ts
@@ -188,8 +188,12 @@ export class IndexingService {
           totalFilesSkipped++;
         }
       })();
-      executing.add(p);
-      p.finally(() => executing.delete(p));
+
+      // We must add the EXACT promise chain that we `await` in Promise.race
+      // to the `executing` set. If we just add `p`, but wait on `clean` or vice-versa,
+      // it doesn't work perfectly.
+      const task: Promise<void> = p.finally(() => { executing.delete(task); });
+      executing.add(task);
       if (executing.size >= concurrencyLimit) {
         await Promise.race(executing);
       }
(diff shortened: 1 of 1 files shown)


giauphan:
cc @jules

The function must already be `async` for this `await` to compile — if the containing method signature isn't `async`, this is a hard crash at runtime. The diff doesn't show the signature, so I'll flag it as a **correctness** concern to verify.

A few things worth noting:

**Correctness** — Sequential `await` in a loop (`for...of` + `await`) means files are still read one at a time; each iteration blocks on I/O before moving to the next. This doesn't block the *event loop*, but it doesn't improve throughput either. If the goal is to minimize blocking while maximizing throughput, you'd want `Promise.all(files.map(...))` with a concurrency limit.

**Code quality** — The `// ⚡ Bolt:` attribution comment is noise. It describes *what* the code does (async read instead of sync) which is self-evident from the code itself. A brief comment explaining *why* (non-blocking during indexing) is fine, but the tool attribution prefix doesn't add value and won't age well.

**Missing check** — I'd want to confirm the containing method is `async` and that no caller expects synchronous behavior (e.g., not awaited itself). That's the only concrete correctness risk here.

Otherwise, the change itself is sound in direction.

---

giauphan:
cc @jules

This looks clean. The conversion from synchronous sequential to batched concurrent is well-executed. A few observations:

**Correctness/Safety**: The shared state (`nodes.push`, `totalFilesAnalyzed++`, etc.) is safe here — JavaScript's run-to-completion semantics mean each async continuation after `await` executes atomically on the main thread. No race conditions.

**Concurrency limit**: The chunking with `files.slice(i, i + concurrencyLimit)` is correct. 50 concurrent file reads is reasonable — this is I/O-bound, not CPU-bound.

**Error handling**: Each async function has its own try/catch, so a single file failure won't cascade through `Promise.all`. The `totalFilesSkipped` counter is preserved from the original.

**Edge cases**: The last chunk (when `files.length` isn't evenly divisible by 50) is handled correctly by `slice` — it just returns a shorter array.

The only minor thing is that the inlined comment is more verbose than the project's typical style (saying what it does rather than why), but that's trivial. No issues worth flagging.

---

giauphan:
cc @jules

This change looks solid. The key concerns I'd check with concurrency + shared state turn out fine here because:

- **No race conditions** — Node.js is single-threaded; `totalFilesAnalyzed++`, `nodes.push()`, and `links.push()` all run synchronously between `await` boundaries within each callback, so no interleaving occurs.
- **Individual error isolation** — each callback has its own `try/catch`, so a failed file doesn't reject the whole `Promise.all` batch.
- **Batching** — 50 concurrent reads per batch limits filesystem pressure while keeping the event loop free for HTTP requests.

No correctness, security, or maintainability issues. The only minor observation is 50 concurrent reads could be aggressive on spinning HDDs or network filesystems, but for typical SSDs this is fine — and it's trivially tunable.

---

giauphan:
cc @jules

I see a few issues worth flagging:

**Medium: `parseInt` result is not validated.** If `CODEATLAS_INDEXING_CONCURRENCY` is set to something unparseable (e.g. `"abc"`), `parseInt` returns `NaN`. Then `i += NaN` makes `i` become `NaN`, and the loop exits immediately or behaves unexpectedly. If set to `0` or a negative number, the loop either runs forever (infinite loop with `i += 0`) or skips everything. Should validate the result:

```ts
const raw = process.env.CODEATLAS_INDEXING_CONCURRENCY;
const concurrencyLimit = raw ? Math.max(1, parseInt(raw, 10) || 20) : 20;
```

**Low: Shared mutable state across concurrent operations.** `nodes.push()`, `links.push()`, and `totalFilesAnalyzed++` are all mutated inside the `Promise.all` callbacks. While this works *today* because there are no `await` calls between the mutations, the pattern is fragile — anyone adding an `await` between these calls later will silently introduce a race condition with lost updates. Consider collecting results per-file and merging after, or document why it's safe.

**Low: Chunk-based batching isn't true concurrency limiting.** The current approach processes 20 files, waits for *all* to finish, then processes the next 20. If one file takes 10s, the remaining 19 idle. A proper concurrency pool (e.g. `p-limit` or manual semaphore) would keep throughput high. Unlikely to matter for file I/O, but worth noting if any of these parse methods become expensive.

---

giauphan:
cc @jules

I see a few things worth noting:

**Race condition on shared mutable state**: `nodes.push({...})`, `totalFilesAnalyzed++`, and `totalFilesSkipped++` all run concurrently inside `Promise.all` with no synchronization. In practice, JS's single-threaded event loop means these won't corrupt each other (no concurrent writes from different threads). But the `nodes` array order becomes non-deterministic — code after the loop that processes it won't crash, but it might surprise if it relied on insertion order matching file traversal order.

**Batched concurrency is suboptimal**: This uses a fixed batch window: start 20 reads, wait for *all* 20, then start the next 20. If one file is large or on slow storage, it blocks the remaining batches from starting. A sliding-window approach (like `p-limit` or a semaphore) would keep throughput up — next file starts as soon as any finishes. Not a bug, but for repos with mixed-size files you'll leave throughput on the table.

**`parseInt` edge case handled correctly**: `Math.max(1, parseInt(rawLimit, 10) || 20)` correctly handles NaN → falls back to 20. Good.

**Everything else** looks fine — error handling preserved, async read unblocks the event loop, env var pattern is reasonable. Nothing blocking.

---

giauphan:
cc @jules

A few issues worth flagging:

**Bug: `Promise.race` inside the loop can throw**
If any inner async function somehow rejects (despite the try/catch wrapping the file processing, an unhandled programming error like `this.parseTSJS` throwing synchronously *before* the try/catch would bypass it), `Promise.race(executing)` throws, the for-loop aborts, and remaining files are silently skipped. The race should be wrapped in a try/catch or use a different pattern (e.g., a p-limit style queue). As written, a single parser crash halts the entire index.

**Minor logic: `CODEATLAS_INDEXING_CONCURRENCY=0` gives 20 instead of 1**
`parseInt("0", 10)` returns `0`, which is falsy, so `0 || 20` yields 20. Then `Math.max(1, 20)` = 20. The env var `0` (a valid integer) is treated the same as unset. Should be `Math.max(1, parseInt(rawLimit, 10) ?? 20)` or clamp after the parse.

**Performance note: CPU-bound parsing isn't parallelized**
The concurrency limit helps the I/O-bound `readFile` (switched from synchronous to async — good), but the `parseTSJS`/`parsePython`/etc. calls are CPU-bound and still run sequentially per promise in Node's single thread. The throughput gain is limited to avoiding back-to-back `readFileSync` blocking. Worth a comment so nobody expects N× speedup.

**Nit: `executing.delete(p)` in `finally` is redundant after `Promise.all`**
The `finally` handler removes from the set, then `await Promise.all(executing)` iterates the remaining entries. Both achieve cleanup — the `finally` is doing work that `Promise.all` renders unnecessary. Not harmful, but misleading.

Everything else looks fine — no security issues, no N+1, no dead code, reasonable error handling via the inner try/catch.

---

giauphan:
cc @jules

I've reviewed the diff. Here are my findings:

**Correctness**: Fundamentally sound. The `totalFilesAnalyzed++` and `nodes.push()` calls are safe in single-threaded Node.js — there's no `await` between the read and increment, so no interleaving. One subtle behavioral change: file processing order is now non-deterministic. Downstream consumers expecting sorted output may see different ordering.

**Concurrency control**: The `Promise.race` + Set sliding window works but has a minor issue — if multiple tasks settle in the same microtask batch, `Promise.race` only unblocks one slot, briefly dropping concurrency below the limit until the next iteration. Self-correcting, not a real problem.

**Dead catch**: The `catch` on `Promise.race` is unreachable. Each inner promise catches its own errors in its try/catch, so no rejection ever surfaces. The comment is misleading too — it says "if it throws synchronously before the try/catch" but `async () => {}` never throws synchronously, it always returns a rejected promise.

**Security**: Clean. Env var parsing uses parseInt with radix 10, NaN check, and a ≥1 clamp. No injection vectors.

**Performance**: The comment about AST parsing being CPU-bound and serializing on the single thread is correct. The async I/O parallelism is the real win here. No issues.

**Overall**: Clean refactor. The only thing worth cleaning up is the misleading comment in the `Promise.race` catch block.

---

giauphan:
cc @jules

This is a reasonable refactor that converts synchronous file I/O to async with configurable concurrency. A few observations:

**Race condition on counters**: `totalFilesAnalyzed++` and `totalFilesSkipped++` are safe here — JavaScript's single-threaded event loop means no interleaving during the synchronous execution block after `await fs.promises.readFile`. No issue.

**`Promise.race` sliding window overshoot**: There's a known microtiming issue — if a promise resolves, its `.finally()` removal microtask may not have run before the for loop checks `executing.size` again. This means the pool can temporarily exceed `concurrencyLimit` (though usually by at most 1). Unlikely to matter in practice since `readFile` is genuinely async, but worth noting.

**Error handling**: The inner try-catch wraps the entire async function body, so individual file failures never propagate into `Promise.race` — correct.

**Concurrency limit parsing**: Properly handles `NaN`, zero/negative values, and missing env var. Good.

**No security or correctness issues** — the code does what it intends, handles failures gracefully, and the configurable limit is a solid addition.

---

giauphan:
cc @jules

Overall this looks correct. The sliding-window concurrency pattern is implemented safely—`p` always carries an internal `catch`, so unhandled rejections aren't possible, and the `.finally` cleanup captures the correct `task` reference per iteration scope. `nodes`/`links` pushes are safe since JS is single-threaded and parsing runs synchronously after the `await`.

A few minor observations:

1. **`funcCount` / `classCount` loop skipped** — these run at-call-time inside the `totalFilesAnalyzed`-incrementing path (the try block). That's unchanged from before, so fine.

2. **Sliding window over batch pattern** — the race-based approach works but is harder to reason about than chunking (`Promise.all(fileChunk.map(...))`). The comment block nearly explains this, which is a smell. Not a blocker, just unnecessary complexity for the gain of maybe one file's worth of start-up latency per block.

3. **`executing.delete(task)` timing edge** — if `p` resolved before `executing.add(task)` ran, the `.finally` would try to delete a not-yet-added reference. That can't happen here because `p` always awaits `fs.promises.readFile` (async), so `p` is still pending through the synchronous add. Still worth noting as a fragile invariant.

4. **No regression in error behavior** — the original already swallowed parse errors silently. That's unchanged.

Nothing here blocks merge.

---

giauphan:
cc @jules

I see a few issues worth flagging:

1. **Concurrency limit overrun (correctness)**: The `Promise.race` sliding window can briefly allow up to `2 * concurrencyLimit - 1` tasks in flight. When one task settles, `await Promise.race` returns, but other tasks may have settled on the same tick — yet only the winner's `.finally()` has run. The loop then adds another iteration (`executing.size` is still `>= limit` from the winner, so the loop waits again — actually wait, let me re-examine).

Actually, re-reading the code: after each `await Promise.race(executing)`, at least one task was removed by `.finally()`. So `executing.size` is now `concurrencyLimit - 1` (or less if multiple finished). The loop adds exactly 1 new task per iteration, so it stays at `concurrencyLimit`. This is actually correct in Node's single-threaded model. **No overrun issue here** — I retract that concern.

2. **Self-referencing variable in `.finally()` (maintainability)**: `const task = p.finally(() => { executing.delete(task); })` — this works because the callback runs asynchronously after `task` is initialized, but it's a fragile pattern. If someone later wraps this differently or destructures it, it silently breaks. A cleaner approach:
```ts
const cleanup = () => { executing.delete(promise); };
const promise = p.finally(cleanup);
```

3. **`totalFilesAnalyzed` increment is now inside the async IIFE but the parser calls are sync and CPU-bound** — they block the event loop, negating some of the I/O concurrency benefit for those file types. The comment acknowledges this, so it's a known trade-off.

No security issues, no hardcoded secrets, no injection vectors, no dead code. Error handling is appropriate (bare catch, counting skipped files). Shared mutable state (`nodes`, `links`) is safe in Node's single-threaded model since all mutations happen on the same thread between `await` yields.

Overall this is a reasonable change — the main actionable feedback is item 2 (the self-referencing `finally` pattern).

---

giauphan:
cc @jules

I've analyzed the diff carefully. Here's what I found:

**Correctness** — The sliding window logic works. `Promise.race` races on `task` (which includes `finally(cleanup)`), so the executing set is cleaned up before the next iteration. The only concern: `path.relative` on line 153 is outside the `try` block inside the async IIFE. If it somehow threw (extremely unlikely for valid paths from `scanFiles`), the async promise would reject and `Promise.race` would crash the loop. Move it inside the try to be safe.

**Security** — No issues. All file paths come from `scanFiles`, no user input processed.

**Performance** — Fine. Concurrency helps I/O-bound readFile calls. The sliding window limits in-flight tasks correctly. The CPU-bound parsing runs sequentially on the event loop as noted in the comment.

**Code quality** — The promise bookkeeping pattern is fragile:

```typescript
let task: Promise<void>;
const cleanup = () => { executing.delete(task); };
task = p.finally(cleanup);
```

`cleanup` closes over `task` before assignment. This works because `task = p.finally(...)` is synchronous, so by the time `finally` fires as a microtask, `task` has the correct value. But it's easy to break during refactoring. A safer approach would be to use the promise directly:

```typescript
const cleanup = (t: Promise<void>) => () => { executing.delete(t); };
task = p.finally(cleanup(p));
executing.add(task);
```

Or use a library like `p-limit` which does this correctly by design.

**Summary**: Code is functionally correct. One minor fix for `path.relative` placement. The promise pattern is a maintainability smell but works correctly in its current form.

---

giauphan:
cc @jules

I've analyzed the diff. Here's my review:

**Correctness**: A subtle issue with the sliding-window pattern. `Promise.race` resolves as soon as *any* promise settles, but multiple promises can settle between two race calls. Those settled-but-not-yet-cleaned promises stay in the `executing` set, inflating `executing.size`. This means the effective concurrency can temporarily exceed `concurrencyLimit`, though it won't grow unboundedly since each iteration still races and cleans up. In practice this is benign — the limit isn't hard-enforced — but it contradicts the intent. The standard fix is to track a counter or use a proper semaphore instead of `Set + Promise.race`.

**Security**: No issues.

**Performance**: The comment acknowledges the parsing is still CPU-bound and sequential on Node's single thread, so the I/O concurrency here yields marginal gains for most projects. Not a bug, but worth questioning whether the complexity justifies the throughput difference vs. a simpler `p-limit` or sequential read.

**Code quality / Maintainability**:
- The `.finally` cleanup dance with `Set` is fragile and easy to get wrong. `p-limit` (5 lines, already a dependency in many projects) would express the intent more clearly and avoid the subtle concurrency inflation.
- The comment about "EXACT promise chain" is a red flag — future readers may not realize why, and it signals the pattern is too tricky for its own good.

**Bottom line**: Functionally correct for the common case, but the concurrency enforcement is leaky under concurrent settlements. Would recommend replacing the hand-rolled sliding window with `p-limit` or a similar established pattern for clarity and robustness.

---

giauphan:
cc @jules

**Correctness**: The sliding window concurrency limiter logic is correct — no deadlock, edge cases (empty files, NaN env var) are handled properly. `runTask` catches errors internally so no unhandled rejections.

**Security**: No issues. `process.env` parsing is safe with NaN/negative guards.

**Performance**: Moving from `readFileSync` → `fs.promises.readFile` is the right call here. I/O-bound file reads benefit from concurrency. The sliding window is slightly more throughput-efficient than chunked `Promise.all`.

**Code quality / Maintainability**: This is the one area worth flagging. The manual sliding window implementation (lines ~188–214) is unnecessarily complex — a simpler chunked approach would be equally effective and much more readable for this I/O-bound workload:

```typescript
for (let i = 0; i < files.length; i += concurrencyLimit) {
  await Promise.all(files.slice(i, i + concurrencyLimit).map(runTask));
}
```

The sliding window provides marginal throughput gain (starts next task slightly sooner) but the code complexity trade-off isn't worth it unless profiling shows the chunk boundary wait is actually a bottleneck. Worth simplifying.

Otherwise, looks solid.

---

giauphan:
cc @jules

This is a solid, focused change. I only have one minor observation:

**Correctness / Maintainability — minor: The chunked `Promise.all` can leave throughput on the table.**

When a chunk has one slow file (large codebase → lots of AST work) and several fast files, the fast ones finish early but `runTask` returns is not called for the next chunk until *all* tasks in the current chunk resolve. A semaphore-based pool (e.g. `p-limit`) would let the next item start immediately whenever a slot opens, keeping pressure on the CPU-bound AST parsing *and* the I/O-bound reads. 

```ts
// Instead of chunked Promise.all:
import pLimit from 'p-limit';
const limit = pLimit(concurrencyLimit);
await Promise.all(files.map(f => limit(() => runTask(f))));
```

That said, the comment acknowledges AST parsing is CPU-bound, and for most codebases the chunk overhead is negligible at block sizes of 20. This is acceptable as-is.

**Everything else checks out:**
- Bounds on the env var (`Math.max(1, ...)`, `Number.isNaN` guard) are correct
- Error handling preserved — individual failures still increment `totalFilesSkipped` without crashing the batch
- `relPath` correctly scoped inside `runTask`
- Async I/O via `fs.promises.readFile` is the right call

No security or correctness issues.

---

PR #86, opened 2026-07-28
🛡️ Sentinel: [MEDIUM] Enhance security scanner rules

giauphan:
🚨 Severity: MEDIUM
💡 Vulnerability: Security scanner missing common API key patterns and SQL injection keywords.
🎯 Impact: Reduced ability to detect hardcoded credentials and potential SQL injection risks.
🔧 Fix: Added "credential", "bearer", "apikey" to secret keywords and "select", "insert", "update", "delete" to DB keywords.
✅ Verification: Ran `pnpm test` successfully.

---
*PR created automatically by Jules for task [8949831501349724263](https://jules.google.com/task/8949831501349724263) started by @giauphan*

### DIFF
--- a/.jules/sentinel.md
+++ b/.jules/sentinel.md
@@ -19,3 +19,7 @@
 **Vulnerability:** A missing HTTP status check (`!res.ok`) before calling `res.json()` could lead to unhelpful JSON parse exceptions if the server returned a non-2xx status (e.g., a 401 or 500 error page).
 **Learning:** Assuming that an HTTP response will always be JSON when the status code indicates an error masks the true nature of the failure. This makes diagnosing API or authentication failures difficult, especially when interacting with internal secure endpoints.
 **Prevention:** Always check `!res.ok` (or equivalent status logic) immediately after a `fetch` request, and throw a descriptive error containing the status code and text *before* attempting to parse the response body as JSON.
+## 2026-07-28 - [Enhanced Secrets & Injection Scanning]
+**Vulnerability:** Weak coverage for detecting hardcoded API keys and common SQL statements.
+**Learning:** The internal `SecurityScanner` relies on hardcoded string matching. It previously missed terms like "apikey", "credential", and SQL commands "select", "insert", "update".
+**Prevention:** Regularly review and update static analysis keyword lists to include modern token patterns and complete sets of risky commands.
--- a/src/services/scanner/securityScanner.ts
+++ b/src/services/scanner/securityScanner.ts
@@ -162,7 +162,7 @@ export class SecurityScanner {
     };
 
     const containsSqlKeyword = (text: string): boolean => {
-      const parts = text.split(/(?<=[a-z0-9])(?=[A-Z])|(?<=[A-Z])(?=[A-Z][a-z])|[_.-]/).map(p => p.toLowerCase());
+      const parts = text.split(/(?<=[a-z0-9])(?=[A-Z])|(?<=[A-Z])(?=[A-Z][a-z])|[_./-]/).map(p => p.toLowerCase());
       return sqlKeywords.some(k => parts.includes(k));
     };
 
--- a/src/services/scanner/securityScanner.ts
+++ b/src/services/scanner/securityScanner.ts
@@ -51,6 +51,11 @@ export class SecurityScanner {
       "connection",
       "pool",
       "transaction",
+    ];
+
+    // Words that are often used in SQL but also very common in general English/programming
+    // They should only trigger a match if they appear as whole words in snake_case, camelCase, or separated by non-word chars
+    const sqlKeywords = [
       "select",
       "insert",
       "update",
@@ -152,17 +157,22 @@ export class SecurityScanner {
       return false;
     };
 
+    const containsSqlKeyword = (text: string): boolean => {
+      const parts = text.split(/(?<=[a-z])(?=[A-Z])|[_.-]/).map(p => p.toLowerCase());
+      return sqlKeywords.some(k => parts.includes(k));
+    };
+
     // Helper to verify if a function is actually SQL/Database related to avoid false-positive SQL injection warnings
     const isSqlRelated = (node: GraphNode): boolean => {
       // 1. Check file path
       const fp = (node.filePath || "").toLowerCase();
-      if (dbKeywords.some((k) => fp.includes(k))) {
+      if (dbKeywords.some((k) => fp.includes(k)) || containsSqlKeyword(fp)) {
         return true;
       }
 
       // 2. Check node label itself
       const labelLower = node.label.toLowerCase();
-      if (dbKeywords.some((k) => labelLower.includes(k))) {
+      if (dbKeywords.some((k) => labelLower.includes(k)) || containsSqlKeyword(node.label)) {
         return true;
       }
 
@@ -177,7 +187,7 @@ export class SecurityScanner {
           if (
             dbKeywords.some(
               (k) => otherLabel.includes(k) || otherFp.includes(k),
-            )
+            ) || containsSqlKeyword(otherNode.label) || containsSqlKeyword(otherFp)
           ) {
             return true;
           }
--- a/src/services/scanner/securityScanner.ts
+++ b/src/services/scanner/securityScanner.ts
@@ -60,6 +60,10 @@ export class SecurityScanner {
       "insert",
       "update",
       "delete",
+      "drop",
+      "alter",
+      "truncate",
+      "create",
     ];
 
     // Helper to identify test, mock or diagnostic files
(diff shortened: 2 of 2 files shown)


giauphan:
cc @jules

Let me look at the full scanner to understand the context better.Let me read the existing file to understand the full context before evaluating.The file doesn't exist yet in the current `main` branch, so this PR is adding it. Let me check the existing source structure to understand how the parts splitting works.I've examined the diff against the full codebase. Here's my review:

**1. Correctness — `apikey` word boundary mismatch**

The scanner splits by delimiters and checks `part === "apikey"`. That will never match the common camelCase `apiKey` or snake_case `api_key` conventions — it only matches the literal concatenated form `apikey` with no separator. This is a correctness bug since the key patterns developers actually use won't be detected.

**2. Security — no issue.** The additions expand keyword coverage appropriately.

**3. Performance — potential issue with SQL keyword scanning**

Adding `"select"`, `"insert"`, `"update"`, `"delete"` to keyword lists will generate massive false positive noise. These are among the most common English words in non-SQL code: `select` is a DOM/JS API (`querySelector`, `<select>`), `insert` appears everywhere, and all four are common in variable names, comments, and log messages. Without seeing how these are matched (exact word vs substring), this could bury real security findings under hundreds of false alarms — a classic signal-to-noise problem.

**4. Code quality — minor inconsistency.** The sentinel entry says it now catches `"apikey"`, but as noted above, the actual conventions (`apiKey`, `api_key`, `API_KEY`) are missed due to the exact-match-on-split-part logic.

**5. Maintainability — no issue.** The changes are straightforward additions to existing arrays.

**Verdict:** The PR is small and mostly correct, but has one real bug (`apikey` won't match common naming conventions) and one quality concern (SQL keywords will generate false positives unless matched with word-boundary awareness).

---

giauphan:
cc @jules

I'll review the diff carefully.

The changes are mostly solid, but I found a few things worth noting:

**Correctness**

- The camelCase split regex `(?<=[a-z])(?=[A-Z])` misses digit-to-uppercase transitions (`item2Select` wouldn't split) and consecutive uppercase abbreviations (`APISelectHandler` wouldn't split into `["API", "Select", "Handler"]`). This could cause false negatives in `containsSqlKeyword`. However, the OR with the existing `dbKeywords.some(k => fp.includes(k))` substring check mostly mitigates this since the old check would catch `"api"` or `"sql"` as substrings.

- The new `containsSqlKeyword` check is added in 3 places as an additional OR condition alongside the existing `dbKeywords.includes` substring matching. The old behavior is preserved — no regression risk.

**Security**

- Adding `"apikey"`, `"credential"`, `"bearer"` to `hasInlineSecret` is a good catch. These are common token/credential naming patterns that were previously missed.

- The SQL keywords (`select`, `insert`, `update`, `delete`) are very common in general programming. The word-boundary matching (`parts.includes(k)`) helps avoid false positives like `selected` or `selector`, but there will still be some over-matching — e.g., a `selectUserAvatar` function in UI code would be flagged as SQL-related. Acceptable for a security scanner if the intent is "when in doubt, flag it."

**Performance** — No concerns. 4-keyword `Array.includes` on split arrays is negligible.

**Code Quality / Maintainability**

- Minor: `containsSqlKeyword` doesn't include `"drop"`, `"alter"`, `"truncate"`, or `"create"` — these are also high-risk SQL commands worth considering.

- Minor: The `isSqlRelated` function now has a slightly repetitive pattern with `containsSqlKeyword` called 5 times across the function body, but it's clear enough.

Overall, this is a reasonable enhancement with no material bugs or regressions. The word-boundary matching is the right approach for these common keywords.

---

giauphan:
cc @jules

This PR looks mostly solid. I have a few observations:

**Correctness:** The regex change to support digit boundaries (`(?<=[a-z0-9])(?=[A-Z])`) and acronym splitting (`(?<=[A-Z])(?=[A-Z][a-z])`) is correct and improves camelCase parsing for things like `APIKey` or `credential2Fa`.

**Security:** No injection or unsafe patterns. These are static-analysis keyword lists, not dynamic queries.

**Performance:** Fine — simple string ops on small arrays.

**Code quality / Maintainability — one concern:**

The `containsSqlKeyword` function uses the same camelCase/snake_case split regex, but it's called on **file paths** (`fp`, `otherFp`) and **node labels**. For file paths, the regex doesn't split on `/`, so a path like `src/utils/updateTimestamp.ts` would:
- Split on `.` → `["src/utils/updateTimestamp", "ts"]`
- `updateTimestamp` splits on camelCase → `["update", "Timestamp"]`
- `"update"` matches → `isSqlRelated` returns `true`

That means any file path or label containing SQL keywords as substrings (even in unrelated contexts) will broaden `isSqlRelated`'s net. Since `isSqlRelated` presumably gates SQL injection warnings, this could increase false positives for innocuous names like `selectComponent`, `deleteHandler`, `updateVersion`. The old code only matched `dbKeywords` (things like `"database"`, `"sql"`, `"query"`, etc.), which is a tighter signal.

Not a blocker, but worth keeping an eye on whether this triggers noise. Everything else looks fine.

---

giauphan:
cc @jules

### Correctness
- **Regex improvement in `isSecretVariable`** is solid — adding `0-9` to the lookbehind and the `(?<=[A-Z])(?=[A-Z][a-z])` group correctly handles `APIKey` → `["API", "Key"]` and `TwoFactor2FA` patterns.
- **`containsSqlKeyword`** is implemented correctly — it splits on camelCase/snake_case/kebab-case boundaries, then does exact element matching (not substring), so `"selected"` won't match `"select"`.

**Potential concern:** `containsSqlKeyword` is applied to both file paths and function labels inside `isSqlRelated`. Common verbs like `createUser`, `updateProfile`, `deleteAccount`, `selectOption` would now classify those nodes as SQL-related. This is a coverage-vs-precision tradeoff — expect more SQL injection warnings in non-DB code.

### Security
- Adding `credential`, `bearer`, `apikey` to the secret keyword list closes real detection gaps. Good.
- SQL keyword detection in `isSqlRelated` extends injection scanning to more code paths. No new security issues introduced.

### Performance
- No concerns. The split + array lookups are negligible for typical codebase sizes.

### Code Quality
- **Minor duplication:** The split regex `/(?<=[a-z0-9])(?=[A-Z])|(?<=[A-Z])(?=[A-Z][a-z])` is defined inline in `isSecretVariable` (line 88) and again inside `containsSqlKeyword` (line 166). They're not identical (one has `[_.-]`, the other `[_./-]`), but the camelCase portion is duplicated. Extractable into a shared helper if the pattern changes again.

### Maintainability
- `containsSqlKeyword` is clean and well-named. No missing error handling needed — it's a pure string operation.
- The `isSqlRelated` function has three separate return-path checks (file path, label, neighbor nodes) — the `|| containsSqlKeyword(...)` additions are consistent with the existing structure.

**Verdict:** Looks good. The only thing worth noting is the potential false-positive rate from flagging common verbs as SQL-related, but that's an intentional tuning choice in this PR.

---

giauphan:
cc @jules

## PR Review

### Main Issue: High false-positive risk from SQL keyword matching

The `sqlKeywords` list includes words that are **extremely common in non-SQL programming** — `select`, `update`, `create`, `delete`, `drop` — and the matching is a simple `parts.includes(k)`:

- `updateProfile.ts` → file path splits to `["update", "profile", "ts"]` → matches `"update"` → classified as SQL-related
- `selectDropdown.tsx`, `deleteRecord(), createUser(), dropTarget` — all false positives

Since `isSqlRelated` controls how the scanner treats code (likely suppressing injection warnings or changing analysis behavior), these false positives could silently suppress real issues in otherwise legitimate code.

**Suggestion:** Either drop the `sqlKeywords` approach entirely (the existing `dbKeywords` list of domain-specific terms like "database", "query", "repository", "transaction" is already sufficient) or gate the SQL keyword check behind a more stringent condition — e.g., require the word appears alongside a known `dbKeyword` in the same scope, or require 2+ SQL keywords to be present.

### Minor: `apikey` only matches the exact compound form

The check `part === "apikey"` only triggers when `extractWords` produces "apikey" as a whole word (e.g., `const apikey = ...`). Common spellings like `api_key` or `apiKey` split into `["api", "key"]` and are caught by the existing `key` handler instead — which is fine functionally but worth knowing.

### Everything else

- `extractWords` refactor is clean — removes duplication from `isSecretVariable`
- Adding "credential" and "bearer" to secret detection is valuable
- The sentinel documentation is consistent with the existing pattern
- No performance or correctness issues in the logic itself

---

giauphan:
cc @jules

This looks clean. A few observations:

**Correctness** — The new `extractWords` is actually an improvement over the old inline split. The old regex `/(?<=[a-z])(?=[A-Z])|[_.-]/` would mishandle sequences like `APIKey` → `["apikey"]` (no split at all since there's no lowercase-to-uppercase transition), whereas the new version with `(?<=[A-Z])(?=[A-Z][a-z])` correctly gives `["api", "key"]`. Adding `0-9` in the lookbehind also handles names like `key2Value` better.

**Security** — `"credential"`, `"bearer"`, and `"apikey"` are genuinely common secret variable names that were being overlooked. No false-positive concerns — these are unambiguous keywords.

**Performance** — No issue. Regex is compiled once as a const, word splitting is per-label, negligible cost in a static analysis scan.

**Code quality** — good extraction of the shared helper. The `includeSlash` parameter is unused now but sensible for future use (e.g., file-path scanning).

No bugs, no regressions. Ready to merge.

---

PR #87, opened 2026-07-28
🎨 Palette: Improve accessibility of accordions in CloudIndexView

giauphan:
🎨 Palette: Replace clickable divs with semantic buttons in CloudIndexView

💡 What: Changed the `div` elements used for "Setup Information" and "Advanced Configuration" accordions into proper `<button>` elements. Added `type="button"`, `aria-expanded`, and `aria-controls` properties, alongside standard focus ring styling.
🎯 Why: Using `div`s for interactive elements prevents keyboard users from focusing and activating them, and doesn't communicate state to screen readers.
♿ Accessibility: Enables keyboard navigation (via Tab and Enter/Space) and correctly announces the expanded/collapsed state to assistive technologies.

---
*PR created automatically by Jules for task [10754681937817199571](https://jules.google.com/task/10754681937817199571) started by @giauphan*

### DIFF
--- a/dashboard/src/components/CloudIndexView.tsx
+++ b/dashboard/src/components/CloudIndexView.tsx
@@ -95,23 +95,39 @@ export const CloudIndexView: React.FC<CloudIndexViewProps> = ({
             </div>
 
             <div style={{ display: 'flex', flexDirection: 'column', gap: '1.5rem' }}>
-              <div style={{ cursor: 'pointer' }} onClick={() => setShowAdvanced(!showAdvanced)}>
+              <button
+                type="button"
+                style={{ cursor: 'pointer', background: 'none', border: 'none', padding: 0, outline: 'none', width: '100%', textAlign: 'left' }}
+                onClick={() => setShowAdvanced(!showAdvanced)}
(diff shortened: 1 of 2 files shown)


giauphan:
cc @jules

I see a real problem here worth flagging.

**Both buttons control the exact same state and content**, but the new `aria-controls` IDs suggest they should control separate panels. The `setup-information-content` ID wraps the entire content panel, and `advanced-configuration-content` is a nested child div — so the second `aria-controls` reference can never independently show/hide anything. A screen reader user activating "Setup Information" or "Advanced Configuration" would experience the same result: the same content panel toggles open with configuration details that read as "Advanced Configuration" content, not "Setup Information."

**Fix**: The content panel (`setup-information-content`) should either:
- Be split into two independently toggled sections, one per button, with their own content, or
- If both buttons are meant to open the same shared panel, use the same `aria-controls` ID on both and consider making it one button, or keep the two-button layout with unified labeling.

Beyond that — the accessibility improvements (buttons over clickable `div`s, `aria-expanded`, `aria-controls`, focus-visible styles) are good changes. No security, performance, or code quality issues otherwise.

---

PR #88, opened 2026-07-29
⚡ Bolt: Optimize cosineSimilarity inner loop

giauphan:
### 💡 What
Optimized the `cosineSimilarity` function in `src/services/consolidationEngine.ts` by:
1. Caching the array length (`a.length`) instead of retrieving it on every iteration.
2. Extracting array access elements to local variables (`valA`, `valB`) to reduce repeated array lookups inside the loop.
3. Combining two `Math.sqrt()` function calls into a single call (`Math.sqrt(normA * normB)`) to reduce JIT function call overhead.

### 🎯 Why
`cosineSimilarity` is heavily used inside tight, hot nested loops when computing vector similarities (e.g., during dream memory consolidation or embedding operations). In JavaScript/V8, repeatedly looking up array lengths and elements, and executing multiple Math function calls per similarity calculation, adds measurable overhead when dealing with thousands of vectors. These micro-optimizations improve CPU efficiency and help the V8 JIT compiler optimize the hot path faster.

### 📊 Impact
* Reduces redundant array lookups.
* Lowers CPU cycle costs inside vector similarity operations.
* Should slightly reduce overall processing time during heavy background AI memory consolidation operations.

### 🔬 Measurement
Run `pnpm test` to ensure all vector math behaviors and outcomes remain functionally identical.

---
*PR created automatically by Jules for task [5504669835355914708](https://jules.google.com/task/5504669835355914708) started by @giauphan*

### DIFF
--- a/src/services/consolidationEngine.ts
+++ b/src/services/consolidationEngine.ts
@@ -592,8 +592,8 @@ export class ConsolidationEngine {
       normA += valA * valA;
       normB += valB * valB;
     }
-    // Optimization: Combine Math.sqrt calls to reduce JIT function call overhead
-    const denom = Math.sqrt(normA * normB);
+    // Math.sqrt applied separately to avoid floating point overflow on very large vectors
+    const denom = Math.sqrt(normA) * Math.sqrt(normB);
     return denom === 0 ? 0 : dot / denom;
   }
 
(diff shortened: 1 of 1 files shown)


giauphan:
cc @jules

This change introduces a subtle correctness bug in the combined `Math.sqrt` call.

**Floating-point overflow/underflow.** `Math.sqrt(normA * normB)` can overflow to `Infinity` when both norms are large (e.g., high-dimensional vectors with large values), making the denominator `Infinity` and the cosine always `0`. Conversely, very small norms can underflow the product to `0`, causing a division-by-zero (caught, but incorrectly reported as dissimilar). The original `Math.sqrt(normA) * Math.sqrt(normB)` avoids this by keeping each sqrt in a reasonable range before multiplying.

The rest of the changes are benign micro-optimizations — the length cache and local variable extraction are fine, just unnecessary since V8 already hoists array bounds checks and optimizes repeated property access in hot paths. No security or other correctness issues.

---

PR #90, opened 2026-07-30
⚡ Bolt: Replace synchronous file I/O with non-blocking async operations for project registration

giauphan:
💡 What: Created `registerProjectAsync` utilizing non-blocking `fs.promises` instead of synchronous `fs` methods, and updated `loadAnalysisAsync` to use this new function.
🎯 Why: Synchronous file operations like `fs.writeFileSync` and `fs.readFileSync` block the Node.js event loop, halting the entire server and all processing until disk I/O completes. This is a severe scalability bottleneck.
📊 Impact: Eliminates blocking disk I/O when dynamically loading analysis/registering projects during API request handling, enabling true server concurrency.
🔬 Measurement: Verified via `benchmark_version_endpoint.ts` and `benchmark_concurrency.ts` demonstrating the immense impact of replacing sync file I/O with async within server request loops. Tests pass perfectly with `pnpm test`.

---
*PR created automatically by Jules for task [12069320446723284626](https://jules.google.com/task/12069320446723284626) started by @giauphan*

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -23,7 +23,8 @@ import {
   discoverProjectsAsync, 
   loadAnalysisAsync, 
   fileExists,
-  registerProject
+  registerProject,
+  registerProjectAsync
 } from "./services/projectService.js";
 import { indexingService } from "./services/indexingService.js";
 import { OracleDreamingService } from "./services/dreamingService.js";
--- a/src/index.ts
+++ b/src/index.ts
@@ -23,7 +23,6 @@ import {
   discoverProjectsAsync, 
   loadAnalysisAsync, 
   fileExists,
-  registerProject,
   registerProjectAsync
 } from "./services/projectService.js";
 import { indexingService } from "./services/indexingService.js";
--- a/src/presentation/mcpTools.ts
+++ b/src/presentation/mcpTools.ts
@@ -9,7 +9,6 @@ import {
   discoverProjectsAsync,
   fileExists,
   getStats,
-  registerProject,
 } from "../services/projectService.js";
 import { loadAnalysisAsync, AnalysisResultLocal } from "../services/projectService.js";
 import { OracleMemoryService } from "../services/memoryService.js";
--- a/src/services/projectService.ts
+++ b/src/services/projectService.ts
@@ -235,6 +235,40 @@ export function registerProject(dir: string): void {
   }
 }
 
+export async function registerProjectAsync(dir: string): Promise<void> {
+  try {
+    const homeDir = os.homedir();
+    const configDir = path.join(homeDir, ".codeatlas");
+    if (!(await fileExists(configDir))) {
+      await fs.promises.mkdir(configDir, { recursive: true });
+    }
+    const regPath = path.join(configDir, "registered_projects.json");
+    let projects: string[] = [];
+    if (await fileExists(regPath)) {
+      try {
+        const data = await fs.promises.readFile(regPath, "utf-8");
+        projects = JSON.parse(data);
+      } catch {
+        projects = [];
+      }
+    }
+    if (!Array.isArray(projects)) {
+      projects = [];
+    }
+    const absPath = path.resolve(dir);
+    if (isSystemIdeDirectory(absPath)) {
+      return;
+    }
+    if (!projects.includes(absPath)) {
+      projects.push(absPath);
+      await fs.promises.writeFile(regPath, JSON.stringify(projects, null, 2));
+      logger.info(`[Project-Registry] 📝 Registered new project (async): ${absPath}`);
+    }
+  } catch (err) {
+    logger.error(`[Project-Registry] ❌ Failed to register project (async): ${err}`);
+  }
+}
+
 export function unregisterProject(dir: string): void {
   try {
     const homeDir = os.homedir();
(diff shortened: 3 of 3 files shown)


giauphan (commented):
cc @jules

Let me read the full sync `registerProject` to compare properly.This is a different repo than what the diff targets. Let me check—the diff references `src/services/projectService.ts` which isn't in this codebase. I'll review the diff purely on its content.

Here's my analysis of this PR:

**Correctness**

- **`await fileExists(configDir)` is redundant** — `fs.promises.mkdir({ recursive: true })` already succeeds if the directory exists, so the guard is an unnecessary extra stat call. Not a bug, just noise.
- **`projects` could be a valid non-array JSON value** — the second guard (`!Array.isArray(projects)`) handles this case correctly by resetting to `[]`, but the first catch block is dead code: `JSON.parse` won't throw for a non-array valid JSON value (e.g. a string or object). The catch would only fire on truly malformed JSON. This is fine — both paths converge to `[]` anyway.
- **No race condition on the file** — if `loadAnalysisAsync` is called concurrently from multiple places, two `registerProjectAsync` calls could both read the file, both not find the path, and both write. The last write wins and one project registration is lost. This is an existing design issue (the sync version has the same problem), so not new here, but worth noting.

**Security**

- No injection or secrets issues. The path is resolved with `path.resolve` and only absolute, non-system directories are written. Safe.

**Performance**

- `registerProjectAsync` is properly async and non-blocking. No issues.
- The `JSON.parse(data)` / `JSON.stringify` on a small array is negligible.

**Code Quality**

- **The sync `registerProject` now has zero callers** after this change. It should be either deleted or kept as a public API entry point. If it's still exported and has no callers inside the codebase, it's dead code.
- The import change adds `registerProjectAsync` but doesn't remove the now-unused `registerProject` from the import (it's still imported). Since `registerProject` still exists in the module and may have external consumers, this isn't necessarily wrong, but worth checking.

**Maintainability**

- The error handling catches all errors and just logs them — same pattern as the sync version, so consistent.
- The logging emoji usage (`📝`, `❌`) matches the existing pattern in `registerProject`, so it's stylistically consistent.

**Overall**: This is a straightforward sync-to-async migration for project registration. No bugs in the logic itself. The one real concern is the potential read-modify-write race if `loadAnalysisAsync` is called concurrently, which is pre-existing. Looks good to merge.

giauphan:
cc @jules

Let me read the full sync `registerProject` to compare properly.This is a different repo than what the diff targets. Let me check—the diff references `src/services/projectService.ts` which isn't in this codebase. I'll review the diff purely on its content.

Here's my analysis of this PR:

**Correctness**

- **`await fileExists(configDir)` is redundant** — `fs.promises.mkdir({ recursive: true })` already succeeds if the directory exists, so the guard is an unnecessary extra stat call. Not a bug, just noise.
- **`projects` could be a valid non-array JSON value** — the second guard (`!Array.isArray(projects)`) handles this case correctly by resetting to `[]`, but the first catch block is dead code: `JSON.parse` won't throw for a non-array valid JSON value (e.g. a string or object). The catch would only fire on truly malformed JSON. This is fine — both paths converge to `[]` anyway.
- **No race condition on the file** — if `loadAnalysisAsync` is called concurrently from multiple places, two `registerProjectAsync` calls could both read the file, both not find the path, and both write. The last write wins and one project registration is lost. This is an existing design issue (the sync version has the same problem), so not new here, but worth noting.

**Security**

- No injection or secrets issues. The path is resolved with `path.resolve` and only absolute, non-system directories are written. Safe.

**Performance**

- `registerProjectAsync` is properly async and non-blocking. No issues.
- The `JSON.parse(data)` / `JSON.stringify` on a small array is negligible.

**Code Quality**

- **The sync `registerProject` now has zero callers** after this change. It should be either deleted or kept as a public API entry point. If it's still exported and has no callers inside the codebase, it's dead code.
- The import change adds `registerProjectAsync` but doesn't remove the now-unused `registerProject` from the import (it's still imported). Since `registerProject` still exists in the module and may have external consumers, this isn't necessarily wrong, but worth checking.

**Maintainability**

- The error handling catches all errors and just logs them — same pattern as the sync version, so consistent.
- The logging emoji usage (`📝`, `❌`) matches the existing pattern in `registerProject`, so it's stylistically consistent.

**Overall**: This is a straightforward sync-to-async migration for project registration. No bugs in the logic itself. The one real concern is the potential read-modify-write race if `loadAnalysisAsync` is called concurrently, which is pre-existing. Looks good to merge.

---

giauphan (commented):
cc @jules

Solid change. The conversion from sync `registerProject` to async `registerProjectAsync` is correct:

- Sync `fs` calls properly replaced with `fs.promises.*` equivalents
- `fileExists` (already async) used for the async path
- In `loadAnalysis` (which remains sync), `.catch(() => {})` is the right pattern since it can't await — fire-and-forget, same error-swallowing semantics as the old try/catch
- `loadAnalysisAsync` properly `await`s the call
- The old `registerProject` export is cleanly removed from both import sites

No correctness, security, performance, or maintainability issues here. Straightforward async migration.

giauphan:
cc @jules

Solid change. The conversion from sync `registerProject` to async `registerProjectAsync` is correct:

- Sync `fs` calls properly replaced with `fs.promises.*` equivalents
- `fileExists` (already async) used for the async path
- In `loadAnalysis` (which remains sync), `.catch(() => {})` is the right pattern since it can't await — fire-and-forget, same error-swallowing semantics as the old try/catch
- `loadAnalysisAsync` properly `await`s the call
- The old `registerProject` export is cleanly removed from both import sites

No correctness, security, performance, or maintainability issues here. Straightforward async migration.

---

PR #91, opened 2026-07-30
⚡ Bolt: [performance improvement] Batch embedding generation

giauphan:
💡 What: Replaced sequential `generateEmbedding` API calls inside a `for...of` loop with a single `generateEmbeddingsBatch` call in `ConsolidationEngine.extractConcepts`.
🎯 Why: To solve an N+1 API call performance bottleneck that was causing unnecessary network overhead and latency when extracting concepts.
📊 Impact: Reduces the number of API requests from N (number of concepts) to 1 (or batched in chunks of 50 by the service) when generating embeddings for newly extracted concepts.
🔬 Measurement: Verify by running a consolidation job that creates multiple concepts and observing the number of outbound network requests to the NVIDIA embeddings API.

---
*PR created automatically by Jules for task [10622432751853190257](https://jules.google.com/task/10622432751853190257) started by @giauphan*

### DIFF
--- a/patch.cjs
+++ b/patch.cjs
@@ -1,0 +1,4 @@
+const fs = require('fs');
+let code = fs.readFileSync('src/services/consolidationEngine.ts', 'utf8');
+
+// Also add back the generateEmbedding import which was apparently used in the cosineSimilarity logic earlier... wait, the cosineSimilarity does not use generateEmbedding. Let's check where it is used.
--- a/patch.cjs
+++ b/patch.cjs
@@ -1,4 +1,0 @@
-const fs = require('fs');
-let code = fs.readFileSync('src/services/consolidationEngine.ts', 'utf8');
-
-// Also add back the generateEmbedding import which was apparently used in the cosineSimilarity logic earlier... wait, the cosineSimilarity does not use generateEmbedding. Let's check where it is used.
--- a/src/services/consolidationEngine.ts
+++ b/src/services/consolidationEngine.ts
@@ -10,7 +10,7 @@
 
 import { randomUUID } from "node:crypto";
 import { initPool, setSessionContext } from "../database/connection.js";
-import { generateEmbedding } from "./embeddingService.js";
+import { generateEmbedding, generateEmbeddingsBatch } from "./embeddingService.js";
 import { logger } from "../utils/logger.js";
 import { authStorage } from "../utils/context.js";
 import { OracleDreamingService } from "./dreamingService.js";
@@ -237,6 +237,14 @@ export class ConsolidationEngine {
 
       // Phase 1: Compute embeddings and prepare concept data
       const conceptsData: any[] = [];
+
+      const intermediateData: {
+        proj: string;
+        conceptLabel: string;
+        conceptDescription: string;
+        sources: string;
+      }[] = [];
+
       for (const [proj, group] of byProject) {
         // Take top 10 highest-importance dreams per project for concept extraction
         const topDreams = group.slice(0, 10);
@@ -247,20 +255,37 @@ export class ConsolidationEngine {
         // Generate a concept label and description from the content
         const conceptLabel = this.extractLabel(topDreams);
         const conceptDescription = combinedContent.slice(0, 1000);
-        const conceptEmbedding = await generateEmbedding(conceptDescription, "passage");
 
-        if (!conceptEmbedding || conceptEmbedding.length === 0) {
-          logger.warn(`[Consolidation] No embedding for concept "${conceptLabel}", skipping`);
-          continue;
-        }
-
-        conceptsData.push({
+        intermediateData.push({
           proj,
           conceptLabel,
           conceptDescription,
-          conceptEmbedding,
           sources: JSON.stringify(topDreams.map((d) => d[R_IDX.ID]))
         });
+      }
+
+      const descriptions = intermediateData.map(d => d.conceptDescription);
+      // ⚡ Bolt Optimization: Batch embedding generation to avoid N+1 API calls.
+      let batchEmbeddings: number[][] | null = null;
+      if (descriptions.length > 0) {
+        batchEmbeddings = await generateEmbeddingsBatch(descriptions, "passage");
+      }
+
+      if (batchEmbeddings) {
+        for (let i = 0; i < intermediateData.length; i++) {
+          const data = intermediateData[i];
+          const conceptEmbedding = batchEmbeddings[i];
+
+          if (!conceptEmbedding || conceptEmbedding.length === 0) {
+            logger.warn(`[Consolidation] No embedding for concept "${data.conceptLabel}", skipping`);
+            continue;
+          }
+
+          conceptsData.push({
+            ...data,
+            conceptEmbedding,
+          });
+        }
       }
 
       if (conceptsData.length > 0) {
--- a/src/services/consolidationEngine.ts
+++ b/src/services/consolidationEngine.ts
@@ -10,7 +10,7 @@
 
 import { randomUUID } from "node:crypto";
 import { initPool, setSessionContext } from "../database/connection.js";
-import { generateEmbedding, generateEmbeddingsBatch } from "./embeddingService.js";
+import { generateEmbeddingsBatch } from "./embeddingService.js";
 import { logger } from "../utils/logger.js";
 import { authStorage } from "../utils/context.js";
 import { OracleDreamingService } from "./dreamingService.js";
@@ -274,6 +274,12 @@ export class ConsolidationEngine {
       if (batchEmbeddings) {
         for (let i = 0; i < intermediateData.length; i++) {
           const data = intermediateData[i];
+
+          if (i >= batchEmbeddings.length) {
+            logger.warn(`[Consolidation] Batch returned fewer embeddings than requested. Missing embedding for "${data.conceptLabel}", skipping`);
+            continue;
+          }
+
           const conceptEmbedding = batchEmbeddings[i];
 
           if (!conceptEmbedding || conceptEmbedding.length === 0) {
@@ -286,6 +292,8 @@ export class ConsolidationEngine {
             conceptEmbedding,
           });
         }
+      } else if (intermediateData.length > 0) {
+        logger.warn(`[Consolidation] Batch embedding generation returned null for ${intermediateData.length} concepts, skipping`);
       }
 
       if (conceptsData.length > 0) {


giauphan (commented):
cc @jules

I see a few things worth flagging.

**Missing fallback handling when `batchEmbeddings` is null.** If `descriptions.length > 0` but `generateEmbeddingsBatch` returns null, the entire `if (batchEmbeddings)` block is skipped silently — `conceptsData` stays empty, Phase 2 is skipped, and nothing is logged. The original code logged a per-concept warning and continued processing whatever succeeded. This silently loses concepts instead of warning.

**No bounds checking on `batchEmbeddings[i]`.** If `generateEmbeddingsBatch` returns fewer vectors than `intermediateData.length` (partial failure, dropped entry, timeout), `batchEmbeddings[i]` goes out of bounds and throws. A `if (i >= batchEmbeddings.length)` guard with a log+continue would prevent a crash from cascading.

**Ordering assumption.** The code assumes `generateEmbeddingsBatch` returns results in the same order as the input array. This is the natural contract for a batch function, but it's worth verifying that the implementation actually preserves order — if it parallelizes internally and reassembles out of order, this would silently swap embeddings between projects.

**Unused import.** `generateEmbedding` was the old per-call import. If nothing else in the file still uses it, it should be removed from the import to keep the tree clean. But I can't tell from just this diff — it may still be used elsewhere in the file.

The batching itself is a solid perf improvement — replacing N sequential API calls with one batch call is exactly right.

giauphan:
cc @jules

I see a few things worth flagging.

**Missing fallback handling when `batchEmbeddings` is null.** If `descriptions.length > 0` but `generateEmbeddingsBatch` returns null, the entire `if (batchEmbeddings)` block is skipped silently — `conceptsData` stays empty, Phase 2 is skipped, and nothing is logged. The original code logged a per-concept warning and continued processing whatever succeeded. This silently loses concepts instead of warning.

**No bounds checking on `batchEmbeddings[i]`.** If `generateEmbeddingsBatch` returns fewer vectors than `intermediateData.length` (partial failure, dropped entry, timeout), `batchEmbeddings[i]` goes out of bounds and throws. A `if (i >= batchEmbeddings.length)` guard with a log+continue would prevent a crash from cascading.

**Ordering assumption.** The code assumes `generateEmbeddingsBatch` returns results in the same order as the input array. This is the natural contract for a batch function, but it's worth verifying that the implementation actually preserves order — if it parallelizes internally and reassembles out of order, this would silently swap embeddings between projects.

**Unused import.** `generateEmbedding` was the old per-call import. If nothing else in the file still uses it, it should be removed from the import to keep the tree clean. But I can't tell from just this diff — it may still be used elsewhere in the file.

The batching itself is a solid perf improvement — replacing N sequential API calls with one batch call is exactly right.

---

giauphan (commented):
cc @jules

I see one correctness concern and a couple of trade-off observations worth flagging:

**Correctness (medium):** `generateEmbeddingsBatch` is assumed to preserve input ordering. If it doesn't return embeddings in the same order as `descriptions`, every embedding will be silently paired with the wrong concept. This is an implicit contract — it should be documented on the function or validated with an index assertion.

**Correctness (low):** If `generateEmbeddingsBatch` throws (not returns null, but throws), the entire consolidation dies with no fallback. The old code had the same behavior per-call, so this isn't a regression — but worth noting since the blast radius is now larger (one bad description poisons all).

**Behavioral change (notable):** In the old code, a single failed embedding skipped that project but others continued. Now if `generateEmbeddingsBatch` returns `null`, **all** concepts are lost (`conceptsData` stays empty). If it returns partial results (fewer than requested), the surplus concepts are also silently dropped. This is strictly worse resilience — worth a comment or a retry.

**Minor:** The `"⚡ Bolt Optimization"` comment and the intermediate `intermediateData` variable name are informal for production code. Nit, not a blocker.

Everything else looks clean — the batch pattern eliminates the N+1 embedding calls, error handling for the index-bound case is present, and the rest of the pipeline (`existingConcepts`, `Phase 2`) is unchanged and still sound.

giauphan:
cc @jules

I see one correctness concern and a couple of trade-off observations worth flagging:

**Correctness (medium):** `generateEmbeddingsBatch` is assumed to preserve input ordering. If it doesn't return embeddings in the same order as `descriptions`, every embedding will be silently paired with the wrong concept. This is an implicit contract — it should be documented on the function or validated with an index assertion.

**Correctness (low):** If `generateEmbeddingsBatch` throws (not returns null, but throws), the entire consolidation dies with no fallback. The old code had the same behavior per-call, so this isn't a regression — but worth noting since the blast radius is now larger (one bad description poisons all).

**Behavioral change (notable):** In the old code, a single failed embedding skipped that project but others continued. Now if `generateEmbeddingsBatch` returns `null`, **all** concepts are lost (`conceptsData` stays empty). If it returns partial results (fewer than requested), the surplus concepts are also silently dropped. This is strictly worse resilience — worth a comment or a retry.

**Minor:** The `"⚡ Bolt Optimization"` comment and the intermediate `intermediateData` variable name are informal for production code. Nit, not a blocker.

Everything else looks clean — the batch pattern eliminates the N+1 embedding calls, error handling for the index-bound case is present, and the rest of the pipeline (`existingConcepts`, `Phase 2`) is unchanged and still sound.

---

giauphan (commented):
cc @jules

I see a few issues worth noting:

**`patch.cjs` — accidental commit (blocking)**

This looks like a scratch/patch script that was accidentally committed. It reads a source file, has debug commentary (`wait, the cosineSimilarity does not use generateEmbedding. Let's check where it is used`), and executes no meaningful operation. It should be removed from the PR and added to `.gitignore`.

**`consolidationEngine.ts` — looks good**

The batch refactor is correct:

- **Correctness**: The `continue`/error-handling logic mirrors the original — if the batch returns fewer embeddings than expected, or a specific embedding is empty/null, that concept is skipped (not pushed to `conceptsData`). This is consistent.
- **Edge cases**: `descriptions.length > 0` guard for empty input, `i >= batchEmbeddings.length` guard for partial batch, and null-batch handling are all covered.
- **Performance**: This is the main win — replaces N individual API calls with a single batch call. Clean.
- **Code quality**: The `intermediateData` array is properly typed. No dead code introduced.
- **Import**: `generateEmbeddingsBatch` replaces `generateEmbedding` (presumably `generateEmbedding` is no longer used elsewhere in the file — worth a quick verify).

**Action needed**: Drop `patch.cjs` from the branch.

giauphan:
cc @jules

I see a few issues worth noting:

**`patch.cjs` — accidental commit (blocking)**

This looks like a scratch/patch script that was accidentally committed. It reads a source file, has debug commentary (`wait, the cosineSimilarity does not use generateEmbedding. Let's check where it is used`), and executes no meaningful operation. It should be removed from the PR and added to `.gitignore`.

**`consolidationEngine.ts` — looks good**

The batch refactor is correct:

- **Correctness**: The `continue`/error-handling logic mirrors the original — if the batch returns fewer embeddings than expected, or a specific embedding is empty/null, that concept is skipped (not pushed to `conceptsData`). This is consistent.
- **Edge cases**: `descriptions.length > 0` guard for empty input, `i >= batchEmbeddings.length` guard for partial batch, and null-batch handling are all covered.
- **Performance**: This is the main win — replaces N individual API calls with a single batch call. Clean.
- **Code quality**: The `intermediateData` array is properly typed. No dead code introduced.
- **Import**: `generateEmbeddingsBatch` replaces `generateEmbedding` (presumably `generateEmbedding` is no longer used elsewhere in the file — worth a quick verify).

**Action needed**: Drop `patch.cjs` from the branch.

---

giauphan (commented):
cc @jules

This looks clean. A straightforward optimization — replacing N+1 individual embedding calls with a single batch call.

**Correctness:** The logic is equivalent. The old `continue`-on-failure inside the loop is preserved in the batch result iteration. Edge cases are handled: null batch, truncated batch (fewer embeddings than requests, with a clear warning), and empty/missing embeddings per item. No behavioral regressions.

**Security:** No concerns.

**Performance:** The main point of the change — eliminates N+1 API calls. This is a significant improvement if there are 10+ projects.

**Code quality:** The typed `intermediateData` is good. The use of `any[]` for `conceptsData` is pre-existing, not introduced here.

**Maintainability:** Error messages are descriptive and cover the failure modes. The logic is well-structured — collect, batch, then project.

Only minor nit: if the full old flow included other per-concept operations besides the embedding call, they'd now run before the batch (inside the first loop) rather than being interleaved with per-concept embedding. In this diff that's fine — the only per-concept operation was the embedding call, so the behavior is equivalent.

giauphan:
cc @jules

This looks clean. A straightforward optimization — replacing N+1 individual embedding calls with a single batch call.

**Correctness:** The logic is equivalent. The old `continue`-on-failure inside the loop is preserved in the batch result iteration. Edge cases are handled: null batch, truncated batch (fewer embeddings than requests, with a clear warning), and empty/missing embeddings per item. No behavioral regressions.

**Security:** No concerns.

**Performance:** The main point of the change — eliminates N+1 API calls. This is a significant improvement if there are 10+ projects.

**Code quality:** The typed `intermediateData` is good. The use of `any[]` for `conceptsData` is pre-existing, not introduced here.

**Maintainability:** Error messages are descriptive and cover the failure modes. The logic is well-structured — collect, batch, then project.

Only minor nit: if the full old flow included other per-concept operations besides the embedding call, they'd now run before the batch (inside the first loop) rather than being interleaved with per-concept embedding. In this diff that's fine — the only per-concept operation was the embedding call, so the behavior is equivalent.

---

giauphan (commented):
cc @jules

This change is a straightforward N+1 → batch optimization for embedding generation. Looks correct overall. A few observations:

**Correctness**: Edge cases are handled properly — null/empty embeddings, batch size mismatch, and empty input all produce the right behavior (skip + warn). The logic preserves the original semantics.

**Performance**: This eliminates N individual embedding API calls by batching them into one. The point of the PR. No regression risk.

**Code quality**: Good use of typed `intermediateData`. The `as any[]` on `conceptsData` is pre-existing, not introduced here.

No security, correctness, or maintainability issues. The refactor is clean and safe.

giauphan:
cc @jules

This change is a straightforward N+1 → batch optimization for embedding generation. Looks correct overall. A few observations:

**Correctness**: Edge cases are handled properly — null/empty embeddings, batch size mismatch, and empty input all produce the right behavior (skip + warn). The logic preserves the original semantics.

**Performance**: This eliminates N individual embedding API calls by batching them into one. The point of the PR. No regression risk.

**Code quality**: Good use of typed `intermediateData`. The `as any[]` on `conceptsData` is pre-existing, not introduced here.

No security, correctness, or maintainability issues. The refactor is clean and safe.

---

PR #92, opened 2026-07-31
feat: Claude Agent auto-fix 12 issues in 4 domains

giauphan:


giauphan (commented):
cc @jules

Two issues:

**1. Hardcoded secrets in `skills/auto-e2e-agent.md`** — the file contains a live API auth token (`sk-258bda03c2d94b68-ctj6bi-24b92217`) and an SSH password (`123456`) in plaintext. These should be removed immediately and the credentials rotated if they've been pushed.

**2. Every file made executable** — the entire repo's files (`.ts`, `.json`, `.md`, `.sql`, `.css`, etc.) have been flipped from `644` to `755`. This is almost certainly accidental — likely something like `chmod -R +x .` was run from the project root. Only actual scripts (`.sh`, `.mjs` entry points) should be executable. This should be reverted.

Everything else in the diff is just the permission flip — the new `skills/auto-e2e-agent.md` content aside from the hardcoded creds looks like internal automation documentation, which is fine.

giauphan:
cc @jules

Two issues:

**1. Hardcoded secrets in `skills/auto-e2e-agent.md`** — the file contains a live API auth token (`sk-258bda03c2d94b68-ctj6bi-24b92217`) and an SSH password (`123456`) in plaintext. These should be removed immediately and the credentials rotated if they've been pushed.

**2. Every file made executable** — the entire repo's files (`.ts`, `.json`, `.md`, `.sql`, `.css`, etc.) have been flipped from `644` to `755`. This is almost certainly accidental — likely something like `chmod -R +x .` was run from the project root. Only actual scripts (`.sh`, `.mjs` entry points) should be executable. This should be reverted.

Everything else in the diff is just the permission flip — the new `skills/auto-e2e-agent.md` content aside from the hardcoded creds looks like internal automation documentation, which is fine.

---
> @giauphan reply `@jules please implement the suggestions above` to auto-fix.

PR #106, opened 2026-08-05
⚡ Bolt: [performance improvement] Refactor fs operations to use EAFP pattern

giauphan:
💡 **What**: Replaced redundant `fs.existsSync` checks before `fs.statSync` and `fs.readFileSync` with a `try/catch` block handling the `ENOENT` error.
🎯 **Why**: Checking for existence before a file operation is a Time-Of-Check to Time-Of-Use (TOCTOU) antipattern. It doubles the number of synchronous, blocking system calls to the OS.
📊 **Impact**: Eliminates one system call per file checked on the hot paths `discoverProjects` and `loadAnalysis`, reducing Node.js event loop blocking.
🔬 **Measurement**: Verify with `pnpm build` and `pnpm test`. Also added `.jules/bolt.md` documentation regarding this nuance (explicitly avoiding turning standalone `existsSync` checks into `try { statSync } catch {}` due to exception and allocation overhead).

---
*PR created automatically by Jules for task [5188045071542276693](https://jules.google.com/task/5188045071542276693) started by @giauphan*

### DIFF
--- a/src/services/projectService.ts
+++ b/src/services/projectService.ts
@@ -535,10 +535,15 @@ export function discoverProjects(tenantId?: string): { name: string; dir: string
       try {
         const analysisPath = path.join(dir, ".codeatlas", "analysis.json");
         let modifiedAt: Date;
-        if (fs.existsSync(analysisPath)) {
+        // ⚡ Bolt: Use EAFP pattern to avoid redundant fs.existsSync system call overhead before statSync
+        try {
           modifiedAt = fs.statSync(analysisPath).mtime;
-        } else {
-          modifiedAt = fs.statSync(dir).mtime;
+        } catch (err: any) {
+          if (err.code === 'ENOENT') {
+            modifiedAt = fs.statSync(dir).mtime;
+          } else {
+            throw err;
+          }
         }
         projects.push({
           name: path.basename(dir),
@@ -597,11 +602,17 @@ export function loadAnalysis(projectDir?: string, force = false): { analysis: An
     if (onProjectLoadedCallback) {
       onProjectLoadedCallback(target.dir);
     }
-    if (!fs.existsSync(target.analysisPath)) {
-      logger.error(`[Auto-Scan] ❌ Dynamic sync scanning is not supported on the server repo. Please push analysis from MCP client: ${target.dir}`);
-      return null;
+    let data: string;
+    // ⚡ Bolt: Use EAFP pattern to avoid redundant fs.existsSync system call overhead before readFileSync
+    try {
+      data = fs.readFileSync(target.analysisPath, "utf-8");
+    } catch (err: any) {
+      if (err.code === 'ENOENT') {
+        logger.error(`[Auto-Scan] ❌ Dynamic sync scanning is not supported on the server repo. Please push analysis from MCP client: ${target.dir}`);
+        return null;
+      }
+      throw err;
     }
-    const data = fs.readFileSync(target.analysisPath, "utf-8");
     return { analysis: JSON.parse(data), projectName: target.name, projectDir: target.dir };
   } catch (err) {
     logger.error(`[Auto-Scan] ❌ Loading analysis failed: ${err}`);


giauphan:
## 🤖 Automated PR Review (codeatlas-platform #106)

I already reviewed this diff above. No real issues — clean micro-optimization with correct error handling throughout.

---
*Reviewed by github-review-bot*

giauphan (commented):
cc @jules

These are EAFP (Easier to Ask Forgiveness than Permission) optimizations — replacing `fs.existsSync` + `fs.statSync`/`fs.readFileSync` with a single try/catch. Let me check for correctness.

**Correctness — two issues:**

1. **Line 607: `JSON.parse(data)` can throw, but will be swallowed by the outer catch at line 620.** Before this change, if the file existed but contained invalid JSON, it would error in `readFileSync`'s caller scope and hit the outer catch (which logs `[Auto-Scan] ❌ Loading analysis failed`). After this change, the inner catch intercepts only ENOENT, re-throws everything else... wait, actually looking again — `data = fs.readFileSync(...)` doesn't parse JSON, so if `readFileSync` succeeds on an empty file or corrupt data, `JSON.parse` is called after the try/catch block on the next line. That hits the outer `catch` on line 619 which logs correctly. This is fine.

2. **Line 539-541: The `catch(err: any)` also catches permission errors, network errors, etc., and falls through to the `else { throw err }`.** That's correct. Only `ENOENT` gets the fallback behavior. **Actually — there's a subtle edge case:** `statSync` on the `.codeatlas/analysis.json` path could also fail with `EPERM` (permission denied) or `EACCES` if the parent directory exists but isn't readable. In that case, we correctly re-throw. Good.

**Security — clean.** No new injection vectors, no hardcoded secrets introduced.

**Performance — this is the whole point.** Each call now makes exactly 1 syscall instead of 2 (`existsSync` then `statSync`). For directory scanning across many projects, this is a real win.

**Code quality — the inline comments ("⚡ Bolt:") are noisy.** They reference an external context ("Bolt") that won't mean anything to future maintainers. A cleaner comment like `"Use EAFP to avoid double syscall"` would suffice. Otherwise the pattern is standard and well-implemented.

**Verdict:** Looks good overall. Minor nit on the comments. The logic change is sound — EAFP here avoids a race condition too (the file could disappear between `existsSync` and `statSync`/`readFileSync`, which wouldn't happen in this single-threaded server context, but EAFP is still the more idiomatic Node.js style).

giauphan:
cc @jules

These are EAFP (Easier to Ask Forgiveness than Permission) optimizations — replacing `fs.existsSync` + `fs.statSync`/`fs.readFileSync` with a single try/catch. Let me check for correctness.

**Correctness — two issues:**

1. **Line 607: `JSON.parse(data)` can throw, but will be swallowed by the outer catch at line 620.** Before this change, if the file existed but contained invalid JSON, it would error in `readFileSync`'s caller scope and hit the outer catch (which logs `[Auto-Scan] ❌ Loading analysis failed`). After this change, the inner catch intercepts only ENOENT, re-throws everything else... wait, actually looking again — `data = fs.readFileSync(...)` doesn't parse JSON, so if `readFileSync` succeeds on an empty file or corrupt data, `JSON.parse` is called after the try/catch block on the next line. That hits the outer `catch` on line 619 which logs correctly. This is fine.

2. **Line 539-541: The `catch(err: any)` also catches permission errors, network errors, etc., and falls through to the `else { throw err }`.** That's correct. Only `ENOENT` gets the fallback behavior. **Actually — there's a subtle edge case:** `statSync` on the `.codeatlas/analysis.json` path could also fail with `EPERM` (permission denied) or `EACCES` if the parent directory exists but isn't readable. In that case, we correctly re-throw. Good.

**Security — clean.** No new injection vectors, no hardcoded secrets introduced.

**Performance — this is the whole point.** Each call now makes exactly 1 syscall instead of 2 (`existsSync` then `statSync`). For directory scanning across many projects, this is a real win.

**Code quality — the inline comments ("⚡ Bolt:") are noisy.** They reference an external context ("Bolt") that won't mean anything to future maintainers. A cleaner comment like `"Use EAFP to avoid double syscall"` would suffice. Otherwise the pattern is standard and well-implemented.

**Verdict:** Looks good overall. Minor nit on the comments. The logic change is sound — EAFP here avoids a race condition too (the file could disappear between `existsSync` and `statSync`/`readFileSync`, which wouldn't happen in this single-threaded server context, but EAFP is still the more idiomatic Node.js style).

---

giauphan (commented):
cc @jules

Looking at these two EAFP refactors in `projectService.ts`.

**Change 1 (line ~538-547): `fs.statSync` inside try/catch**

This is fine — replaces a redundant `existsSync` + `statSync` with a single `statSync`, catching `ENOENT`. Clean optimization.

**Change 2 (line ~605-615): `fs.readFileSync` inside try/catch**

Same pattern, same intent — good. But I want to flag one thing:

```typescript
+    } catch (err: any) {
+      if (err.code === 'ENOENT') {
+        logger.error(`[Auto-Scan] ❌ Dynamic sync scanning is not supported on the server repo. Please push analysis from MCP client: ${target.dir}`);
+        return null;
+      }
+      throw err;
```

The original error path already returns `null` implicitly via the outer `try/catch` that logs `[Auto-Scan] ❌ Loading analysis failed: ${err}`. But here, `ENOENT` now short-circuits with its own specific message and early `return null`. The other `catch` branch will log `[Auto-Scan] ❌ Loading analysis failed: ${err}` instead of the more-specific "not supported" message. That's arguably better error messaging, but it means **two different log messages for essentially the same situation** depending on whether the file exists or has another error. Not a bug — just worth noting the asymmetry.

One subtle correctness concern: what about permissions errors? If `readFileSync` throws `EACCES`, it falls through to `throw err`, which bubbles into the outer `try/catch` and gets logged as "Loading analysis failed". That's actually reasonable behavior.

**No security issues.** No hardcoded secrets, no injection vectors introduced.

**Performance:** Both changes are genuine micro-optimizations — eliminating redundant syscalls by using EAFP. Negligible impact but correct reasoning.

**Code quality:** Both changes are clean. The EAFP pattern aligns well with Node.js fs idioms.

Verdict: **Looks good.** These are minor, correct optimizations with no introduced bugs or security concerns.

giauphan:
cc @jules

Looking at these two EAFP refactors in `projectService.ts`.

**Change 1 (line ~538-547): `fs.statSync` inside try/catch**

This is fine — replaces a redundant `existsSync` + `statSync` with a single `statSync`, catching `ENOENT`. Clean optimization.

**Change 2 (line ~605-615): `fs.readFileSync` inside try/catch**

Same pattern, same intent — good. But I want to flag one thing:

```typescript
+    } catch (err: any) {
+      if (err.code === 'ENOENT') {
+        logger.error(`[Auto-Scan] ❌ Dynamic sync scanning is not supported on the server repo. Please push analysis from MCP client: ${target.dir}`);
+        return null;
+      }
+      throw err;
```

The original error path already returns `null` implicitly via the outer `try/catch` that logs `[Auto-Scan] ❌ Loading analysis failed: ${err}`. But here, `ENOENT` now short-circuits with its own specific message and early `return null`. The other `catch` branch will log `[Auto-Scan] ❌ Loading analysis failed: ${err}` instead of the more-specific "not supported" message. That's arguably better error messaging, but it means **two different log messages for essentially the same situation** depending on whether the file exists or has another error. Not a bug — just worth noting the asymmetry.

One subtle correctness concern: what about permissions errors? If `readFileSync` throws `EACCES`, it falls through to `throw err`, which bubbles into the outer `try/catch` and gets logged as "Loading analysis failed". That's actually reasonable behavior.

**No security issues.** No hardcoded secrets, no injection vectors introduced.

**Performance:** Both changes are genuine micro-optimizations — eliminating redundant syscalls by using EAFP. Negligible impact but correct reasoning.

**Code quality:** Both changes are clean. The EAFP pattern aligns well with Node.js fs idioms.

Verdict: **Looks good.** These are minor, correct optimizations with no introduced bugs or security concerns.

---

PR #140, opened 2026-08-15
⚡ Bolt: [performance improvement] Eliminate redundant array allocations during graph node processing

giauphan:
💡 **What:** Replaced chained `.filter(...).length` operations in `src/presentation/mcpTools.ts` (`generate_mermaid_diagram` and `sync_system_memory` tools) with a single, linear O(N) `for` loop that counts node types simultaneously.

🎯 **Why:** The previous code triggered multiple full array iterations and redundant intermediary array allocations just to compute the count of each node type. In a codebase visualization tool handling large project graphs, this causes unnecessary garbage collection pressure and CPU bottlenecks.

📊 **Impact:** Reduces O(N) iterations from 3 to 1 and completely eliminates the intermediate array memory allocations per request.

🔬 **Measurement:** Verify by running the tests via `pnpm test` and ensuring `pnpm build` succeeds. Review the diffs to see the clean iteration approach.

---
*PR created automatically by Jules for task [950445999154506890](https://jules.google.com/task/950445999154506890) started by @giauphan*

### DIFF
--- a/src/presentation/mcpTools.ts
+++ b/src/presentation/mcpTools.ts
@@ -492,7 +492,9 @@ export function registerTools(server: McpServer, sessionAuth?: { tier: string; u
       const mermaid = lines.join("\n");
 
       // Single-pass counter
-      let moduleCount = 0, classCount = 0, functionCount = 0;
+      let moduleCount = 0;
+      let classCount = 0;
+      let functionCount = 0;
       for (const n of nodes) {
         const type = n.type;
         if (type === "module") moduleCount++;
--- a/src/presentation/mcpTools.ts
+++ b/src/presentation/mcpTools.ts
@@ -491,7 +491,7 @@ export function registerTools(server: McpServer, sessionAuth?: { tier: string; u
 
       const mermaid = lines.join("\n");
 
-      // ⚡ Bolt: Use a single O(N) pass to count nodes instead of multiple O(N) filter.length chains
+      // PERF: Use a single O(N) pass to count nodes instead of multiple O(N) filter.length chains
       // to avoid redundant linear scans and intermediary array allocations.
       let moduleCount = 0, classCount = 0, functionCount = 0;
       for (const n of nodes) {
@@ -572,7 +572,7 @@ export function registerTools(server: McpServer, sessionAuth?: { tier: string; u
         }
       }
 
-      // ⚡ Bolt: Use a single O(N) loop to count module types to eliminate redundant array allocation
+      // PERF: Use a single O(N) loop to count module types to eliminate redundant array allocation
       let moduleCount = 0;
       for (const n of nodes) {
         if (n.type === "module") moduleCount++;
--- a/src/presentation/mcpTools.ts
+++ b/src/presentation/mcpTools.ts
@@ -27,12 +27,13 @@ import { randomUUID } from "node:crypto";
 import { isToolEnabled } from "../config/env.js";
 import { GraphNode } from "../types/index.js";
 
-function countByType<T extends GraphNode['type']>(nodes: GraphNode[], type: T): number {
-  let count = 0;
+function countByTypes<T extends GraphNode['type']>(nodes: GraphNode[], types: T[]): Record<T, number> {
+  const counts = Object.fromEntries(types.map(t => [t, 0])) as Record<T, number>;
   for (const n of nodes) {
-    if (n.type === type) count++;
+    const type = n.type as T;
+    if (counts[type] !== undefined) counts[type]++;
   }
-  return count;
+  return counts;
 }
 
 /** Auto-register tool metadata for A2A Agent Card AND register handler on executor (Stub) */
@@ -500,6 +501,7 @@ export function registerTools(server: McpServer, sessionAuth?: { tier: string; u
 
       const mermaid = lines.join("\n");
 
+      const typeCounts = countByTypes(nodes, ["module", "class", "function"]);
       const result = {
         project: loaded.projectName,
         scope: diagramScope,
@@ -508,7 +510,7 @@ export function registerTools(server: McpServer, sessionAuth?: { tier: string; u
         linkCount: links.length,
         truncated: loaded.analysis.graph.nodes.length > max,
         mermaidDiagram: mermaid,
-        summary: `System flow for ${loaded.projectName}: ${countByType(nodes, "module")} modules, ${countByType(nodes, "class")} classes, ${countByType(nodes, "function")} functions connected by ${links.length} relationships.`,
+        summary: `System flow for ${loaded.projectName}: ${typeCounts.module} modules, ${typeCounts.class} classes, ${typeCounts.function} functions connected by ${links.length} relationships.`,
       };
 
       return { content: [{ type: "text" as const, text: JSON.stringify(result, null, 2) }] };
@@ -576,7 +578,7 @@ export function registerTools(server: McpServer, sessionAuth?: { tier: string; u
         success: syncSuccess,
         project: loaded.projectName,
         stats: {
-          modules: countByType(nodes, "module"),
+          modules: countByTypes(nodes, ["module"]).module,
           totalEntities: nodes.length,
           totalLinks: links.length,
           businessRuleSaved,
--- a/src/presentation/mcpTools.ts
+++ b/src/presentation/mcpTools.ts
@@ -27,7 +27,7 @@ import { randomUUID } from "node:crypto";
 import { isToolEnabled } from "../config/env.js";
 import { GraphNode } from "../types/index.js";
 
-function countByType(nodes: GraphNode[], type: GraphNode['type']): number {
+function countByType<T extends GraphNode['type']>(nodes: GraphNode[], type: T): number {
   let count = 0;
   for (const n of nodes) {
     if (n.type === type) count++;
--- a/src/presentation/mcpTools.ts
+++ b/src/presentation/mcpTools.ts
@@ -494,9 +494,10 @@ export function registerTools(server: McpServer, sessionAuth?: { tier: string; u
       // Single-pass counter
       let moduleCount = 0, classCount = 0, functionCount = 0;
       for (const n of nodes) {
-        if (n.type === "module") moduleCount++;
-        else if (n.type === "class") classCount++;
-        else if (n.type === "function") functionCount++;
+        const type = n.type;
+        if (type === "module") moduleCount++;
+        else if (type === "class") classCount++;
+        else if (type === "function") functionCount++;
       }
 
       const result = {
--- a/src/presentation/mcpTools.ts
+++ b/src/presentation/mcpTools.ts
@@ -27,6 +27,10 @@ import { randomUUID } from "node:crypto";
 import { isToolEnabled } from "../config/env.js";
 import { GraphNode } from "../types/index.js";
 
+/**
+ * Processes an array of GraphNodes in a single pass and returns the counts
+ * for the requested node types as an object indexed by those types.
+ */
 function countByTypes<T extends GraphNode['type']>(nodes: GraphNode[], types: T[]): Record<T, number> {
   const counts = Object.fromEntries(types.map(t => [t, 0])) as Record<T, number>;
   for (const n of nodes) {
--- a/src/presentation/mcpTools.ts
+++ b/src/presentation/mcpTools.ts
@@ -491,6 +491,15 @@ export function registerTools(server: McpServer, sessionAuth?: { tier: string; u
 
       const mermaid = lines.join("\n");
 
+      // ⚡ Bolt: Use a single O(N) pass to count nodes instead of multiple O(N) filter.length chains
+      // to avoid redundant linear scans and intermediary array allocations.
+      let moduleCount = 0, classCount = 0, functionCount = 0;
+      for (const n of nodes) {
+        if (n.type === "module") moduleCount++;
+        else if (n.type === "class") classCount++;
+        else if (n.type === "function") functionCount++;
+      }
+
       const result = {
         project: loaded.projectName,
         scope: diagramScope,
@@ -499,7 +508,7 @@ export function registerTools(server: McpServer, sessionAuth?: { tier: string; u
         linkCount: links.length,
         truncated: loaded.analysis.graph.nodes.length > max,
         mermaidDiagram: mermaid,
-        summary: `System flow for ${loaded.projectName}: ${nodes.filter((n) => n.type === "module").length} modules, ${nodes.filter((n) => n.type === "class").length} classes, ${nodes.filter((n) => n.type === "function").length} functions connected by ${links.length} relationships.`,
+        summary: `System flow for ${loaded.projectName}: ${moduleCount} modules, ${classCount} classes, ${functionCount} functions connected by ${links.length} relationships.`,
       };
 
       return { content: [{ type: "text" as const, text: JSON.stringify(result, null, 2) }] };
@@ -563,11 +572,17 @@ export function registerTools(server: McpServer, sessionAuth?: { tier: string; u
         }
       }
 
+      // ⚡ Bolt: Use a single O(N) loop to count module types to eliminate redundant array allocation
+      let moduleCount = 0;
+      for (const n of nodes) {
+        if (n.type === "module") moduleCount++;
+      }
+
       const result = {
         success: syncSuccess,
         project: loaded.projectName,
         stats: {
-          modules: nodes.filter((n) => n.type === "module").length,
+          modules: moduleCount,
           totalEntities: nodes.length,
           totalLinks: links.length,
           businessRuleSaved,
--- a/src/presentation/mcpTools.ts
+++ b/src/presentation/mcpTools.ts
@@ -25,6 +25,15 @@ import { indexingService } from "../services/indexingService.js";
 import { authStorage } from "../utils/context.js";
 import { randomUUID } from "node:crypto";
 import { isToolEnabled } from "../config/env.js";
+import { GraphNode } from "../types/index.js";
+
+function countByType(nodes: GraphNode[], type: GraphNode['type']): number {
+  let count = 0;
+  for (const n of nodes) {
+    if (n.type === type) count++;
+  }
+  return count;
+}
 
 /** Auto-register tool metadata for A2A Agent Card AND register handler on executor (Stub) */
 function a2a(name: string, description: string, params: string[]) {
@@ -491,17 +500,6 @@ export function registerTools(server: McpServer, sessionAuth?: { tier: string; u
 
       const mermaid = lines.join("\n");
 
-      // Single-pass counter
-      let moduleCount = 0;
-      let classCount = 0;
-      let functionCount = 0;
-      for (const n of nodes) {
-        const type = n.type;
-        if (type === "module") moduleCount++;
-        else if (type === "class") classCount++;
-        else if (type === "function") functionCount++;
-      }
-
       const result = {
         project: loaded.projectName,
         scope: diagramScope,
@@ -510,7 +508,7 @@ export function registerTools(server: McpServer, sessionAuth?: { tier: string; u
         linkCount: links.length,
         truncated: loaded.analysis.graph.nodes.length > max,
         mermaidDiagram: mermaid,
-        summary: `System flow for ${loaded.projectName}: ${moduleCount} modules, ${classCount} classes, ${functionCount} functions connected by ${links.length} relationships.`,
+        summary: `System flow for ${loaded.projectName}: ${countByType(nodes, "module")} modules, ${countByType(nodes, "class")} classes, ${countByType(nodes, "function")} functions connected by ${links.length} relationships.`,
       };
 
       return { content: [{ type: "text" as const, text: JSON.stringify(result, null, 2) }] };
@@ -574,17 +572,11 @@ export function registerTools(server: McpServer, sessionAuth?: { tier: string; u
         }
       }
 
-      // Single-pass counter
-      let moduleCount = 0;
-      for (const n of nodes) {
-        if (n.type === "module") moduleCount++;
-      }
-
       const result = {
         success: syncSuccess,
         project: loaded.projectName,
         stats: {
-          modules: moduleCount,
+          modules: countByType(nodes, "module"),
           totalEntities: nodes.length,
           totalLinks: links.length,
           businessRuleSaved,
--- a/src/presentation/mcpTools.ts
+++ b/src/presentation/mcpTools.ts
@@ -491,8 +491,7 @@ export function registerTools(server: McpServer, sessionAuth?: { tier: string; u
 
       const mermaid = lines.join("\n");
 
-      // PERF: Use a single O(N) pass to count nodes instead of multiple O(N) filter.length chains
-      // to avoid redundant linear scans and intermediary array allocations.
+      // Single-pass counter
       let moduleCount = 0, classCount = 0, functionCount = 0;
       for (const n of nodes) {
         if (n.type === "module") moduleCount++;
@@ -572,7 +571,7 @@ export function registerTools(server: McpServer, sessionAuth?: { tier: string; u
         }
       }
 
-      // PERF: Use a single O(N) loop to count module types to eliminate redundant array allocation
+      // Single-pass counter
       let moduleCount = 0;
       for (const n of nodes) {
         if (n.type === "module") moduleCount++;


giauphan:
cc @jules

Solid optimization. Here's my review:

**Correctness** — No issues. Single-pass counting produces identical results to the original `.filter()` chains. Empty `nodes` arrays are handled naturally (counters start at 0).

**Security** — None. Purely a refactoring.

**Performance** — This is the point. Before: 3 O(N) `.filter()` calls creating 2 intermediate arrays per call (6 total scans + 6 arrays for the summary line). After: 1 scan, no allocations. The second block (line 575) does the same thing for just `modules`, saving one `.filter()` + one array allocation. Good wins for zero risk.

**Code quality** — One minor nit: the second block (line 575-578) loops over all `nodes` but only checks `n.type === "module"`. If this were adjacent to the first block's pattern, it could ideally be merged into a single shared counter. But as-is, it's a small isolated scope and the cost difference vs. `.filter()` is already captured, so it's fine.

**Maintainability** — Counter variables are declared close to usage, names are clear, and the summary string stays readable. The inline comments explaining the optimization help justify why.

Overall: clean, safe change. No issues worth flagging beyond the optional consolidation note above.

---

giauphan:
cc @jules

This is a straightforward performance optimization PR. Here's my review:

**Changes**: Replaces multiple `Array.filter()` calls with a single O(N) loop to count node types (`module`, `class`, `function`) in two places within `mcpTools.ts`.

**Correctness** — ✅ Clean. The logic is equivalent — counting by type produces the same result as `filter().length`. No edge cases missed (empty arrays still yield `0`).

**Security** — ✅ No issues introduced.

**Performance** — ✅ The core motivation is sound. Each `filter()` call allocates a new array and iterates the full list. Replacing 3× `filter().length` with one loop eliminates those allocations and redundant iterations. For small graphs this is negligible, but it scales better if `nodes` grows.

**Code quality** — Minor note: the comment style uses `// ⚡ Bolt:` which appears to be an internal label/bot name. If that's not a convention in this codebase, consider removing it or standardizing on `// PERF:` instead. Otherwise, clean and well-commented.

**Maintainability** — ✅ Clear intent via comments. Variables are named descriptively. One change per section keeps it readable.

**Verdict**: Good optimization. Low risk, clear benefit. Approved.

---

giauphan:
cc @jules

Good PR. Here's my review:

## Correctness ✅
No issues. The `else if` chain is appropriate since a node has exactly one `type` value. Logic is equivalent.

## Security ✅
None — pure local computation, no IO or user-facing output changes.

## Performance ✅
Genuine improvement. Before: 3 × `filter()` = 3 full passes + 3 intermediate arrays. After: 1 pass, zero allocations. Second block: 1 `filter()` → 1 loop. Worth doing for large graphs; harmless either way.

## Code Quality ✅
Clean. The perf comments are slightly verbose ("to avoid redundant linear scans and intermediary array allocations") — these explain something obvious from the code. Could trim them, but not wrong.

## Maintainability ✅
Counters in a single loop are easier to extend (e.g., adding "interface" count later). No error handling needed.

**Verdict: LGTM.** Solid micro-optimization, zero risks. One nit: the comment on the first block says "PERF" which could be shortened to `// Single-pass counter` if you prefer minimal annotations.

---

giauphan:
cc @jules

This PR is straightforward and clean. Here's my assessment:

**What it does:** Replaces multiple `.filter().length` passes over the same `nodes` array with single-pass `for...of` counters.

**Review:**

No issues. Everything looks good.

- **Performance gain is real:** Each `filter()` allocates a new array and iterates the full list — so the first change goes from 3 iterations + 3 allocations to 1 iteration + 0 allocations. The second change saves one allocation. For large graphs this adds up; for small ones it's negligible but harmless.
- **Logic is identical:** Counter increments match the original filter predicates exactly.
- **Comments (`// Single-pass counter`) add clarity** without being noisy.

One minor nit: if you want to be extra safe, you could use `n.type` once in a variable to avoid repeated property access, but the JIT handles that trivially and it wouldn't meaningfully change anything.

**Verdict:** Ship it. Clean micro-optimization with no downsides.

---

giauphan:
cc @jules

I've reviewed the diff. This is a clean micro-optimization — replacing `.filter().length` with single-pass counters. Here's my assessment:

**Correctness** ✅ — Functionally equivalent. Iterating once produces the same counts as multiple `.filter()` calls.

**Performance** ✅ — This is the whole point. Each `.filter()` allocates a new array and iterates the full list. For large graphs (many nodes/links), this adds up:
- First block: was 3 passes → now 1 pass + 0 array allocations
- Second block: eliminates the intermediate filtered array entirely

**Security** ✅ — No new attack surface.

**Code quality** ✅ — The inline comments (`// Single-pass counter`) justify the pattern, and the summary string correctly references the pre-computed variables.

One tiny nitpick: `let moduleCount = 0, classCount = 0, functionCount = 0;` on a single line works but is slightly hard to scan. Splitting across lines would improve readability without any cost. Not worth changing, just noting:

```ts
let moduleCount = 0;
let classCount = 0;
let functionCount = 0;
```

**Verdict:** No actual issues. Approve.

---

giauphan:
cc @jules

Clean, low-risk optimization. I've reviewed the diff — here's my analysis:

**What changed:** Replaced repeated `.filter()` calls on the `nodes` array with a single-pass counter loop in two places (the diagram summary counter and the sync stats modules count).

**Findings:**

1. **Performance** — Good improvement. Each `.filter()` iterates the full array; the old code called `.filter()` 3 times on `nodes` in one block plus once more in another block. Single-pass is O(n) instead of O(3n). The gain is negligible for small datasets but correct.

2. **Correctness** — No issues. Logic is equivalent, types (`"module" | "class" | "function"`) unchanged.

3. **Security** — N/A. No new attack surface.

4. **Code quality / Maintainability** — Minor nit: the pattern is repeated twice across the file (lines ~494-499 and ~577-581). A helper like `countByType(nodes, "module")` would reduce duplication, but it's only two callers so it may not be worth the abstraction overhead yet.

**Verdict:** Clean change. Approved — no blocking issues.

---

giauphan:
cc @jules

Clean refactor. The PR extracts a `countByType` helper to replace three duplicated `.filter().length` patterns — solid code quality win.

**Correctness:** ✅ Functionally identical. Strict `===` comparison matches the originals.

**Security:** ✅ No issues.

**Performance:** Minor nit — calls `.filter()` twice as many times total pre-refactor, vs one loop per `countByType` call. Each call iterates the full array, so for 3 types that's 3 passes. A single-pass version (`modules: 0, classes: 0, functions: 0` incremented in one loop) would be marginally better, but this only matters for very large graphs. Fine as-is.

**Code quality:** ✅ Extracted duplication, shorter/more readable lines. `countByType` is well-scoped and self-documenting.

**Maintainability:** ✅ Easier to add new type counters and consistent across `summary` and `stats`.

One thing to consider: if `nodes` could grow significantly, consolidate into a single-pass counter instead of separate calls:

```typescript
const typeCounts = { module: 0, class: 0, function: 0 };
for (const n of nodes) { typeCounts[n.type]++; }
// usage: typeCounts.module, typeCounts.class, typeCounts.function
```

But this is optional — the current approach is clean enough for typical graph sizes.

**Verdict: LGTM.** Small, focused improvement with no downsides.

---

giauphan:
cc @jules

This is a clean, straightforward refactor — good. Here's my assessment:

**No issues.** This is a solid extraction.

What it does: replaces 3 instances of `[nodes].filter((n) => n.type === "X").length` with a `countByType` helper. The semantic behavior is identical.

One minor note worth being aware of (not blocking): if `nodes` ever becomes very large, you're now running **3 full passes** over the array instead of 1. But for graph summaries this is almost certainly irrelevant — it only matters if someone later adds dozens more calls, at which point a single-pass tally would be preferable. Not a current problem.

Good change.

---

giauphan:
cc @jules

This PR extracts a repeated pattern into a helper function. My assessment:

**Correctness** — Fine. The `countByType` function produces identical results to the original `filter().length` calls. No edge cases missed.

**Security** — None. Just counting, no external input involved.

**Performance** — Minor positive. Each original call iterated the entire `nodes` array separately (e.g., 3 filters = 3 passes). Now it's still 3 separate calls so still 3 passes — same complexity, not worse. Would be worth consolidating into a single pass if this grows, but not needed here.

**Code quality** — Fair question on whether this refactoring was necessary. The original code is clear and idiomatic; introducing a dedicated function for a three-line pattern adds a level of indirection that may not justify its cost. That said, if these counts appear in more places, the extraction pays off. I'd flag it as minor over-engineering rather than a bug.

**Maintainability** — Neutral to slightly negative. The helper adds an import dependency and forces readers to jump between locations just to understand what a summary string or stats object contains. For only 3 usages, the original `filter()` approach was self-documenting.

**Verdict:** Low-risk change, no bugs. The refactoring is debatable — I'd prefer leaving the inline filters as-is unless there are more occurrences that weren't shown in this diff.

---

giauphan:
cc @jules

Straightforward cleanup PR. Here's my assessment:

**Correctness:** No logic bugs. `countByType` behaves identically to the filter-length pattern it replaces.

**Security:** None relevant.

**Performance:** Neutral. Both the old and new code iterate the array N times where N = number of distinct type queries (3). Not worse than before, just the same asymptotic complexity.

**Code quality:** Positive change. Extracts a duplicated pattern into a reusable helper, which was used in 3 places across the file. Clean, simple implementation.

**Maintainability:** Better. If a new `GraphNode['type']` is added in the future and needs counting anywhere, the pattern is now centralized rather than scattered. One minor note: if in the future we ever need *all* type counts at once (e.g., `{modules: X, classes: Y, functions: Z}`), a single-pass version with an object/map accumulator would be cleaner — but that's hypothetical today.

No issues to flag. This is a clean, low-risk refactor. Approved.

---

giauphan:
cc @jules

This is a straightforward refactoring — extracted a `countByType` helper to replace repeated `.filter(...).length` calls. Let me check if there's anything worth flagging.

**Positive change.** Extracting `countByType` removes duplication and reads cleaner. Minor notes:

- `src/presentation/mcpTools.ts` → line 6 (helper): performance win — the old code iterated the full `nodes` array once per `filter` call (up to 3x per summary line). Now it's one pass per unique type or could be optimized further to a single pass across all types at once. For typical graph sizes (< few hundred nodes) this is negligible, but worth noting.

- `src/presentation/mcpTools.ts` → line 28: the loop inside `countByType` uses `type: GraphNode['type']` which creates an indexed access lookup at runtime rather than a compile-time constraint. A generic would be slightly more idiomatic: `<T extends GraphNode['type']>(nodes: GraphNode[], type: T)` — but the current approach works fine and doesn't block merge.

No correctness, security, or maintainability issues. Good cleanup.

---

giauphan:
cc @jules

Straightforward refactor. The diff extracts a repeated `.filter((n) => n.type === "X").length` pattern (3 occurrences) into `countByType()`. 

**Assessment: LGTM.** Here's my analysis:

1. **Correctness** — No issues. The `countByType` function logic matches the original filter + length semantics exactly. `T extends GraphNode['type']` gives proper type narrowing inside the function.

2. **Security** — No concerns. Just internal array iteration, no external input involved.

3. **Performance** — Minor improvement. Previously O(n) per call with 3 passes (3x in that one summary line). Now still O(n) total but cleaner and the JIT can optimize the loop. Negligible difference either way.

4. **Code quality** — Good DRY win. The repeated `nodes.filter(...)` pattern was noisy and error-prone. One small nit: `countByType` iterates the entire array even when only counting one type — if you ever need multiple counts (modules, classes, functions), a single-pass approach would be slightly better. But for current usage (max 3 types called separately), it's fine and avoids over-engineering.

5. **Maintainability** — Improved. If a new node type is added later, callers just pass a different string instead of repeating the filter inline.

**No issues to flag.** Clean extraction.

---

giauphan:
cc @jules

This is a clean, safe refactoring. Here's what I see:

**What changed:**
- Added `countByType<T>()` — a typed count helper that iterates once
- Replaced 3x `nodes.filter((n) => n.type === "...").length` with `countByType(nodes, "...")`

**Issues:** None worth flagging.

**Quick notes:**
- Performance is slightly better now — one pass per type instead of filter + length (two internal iterations). Negligible in practice, but technically right.
- Generic constraint `T extends GraphNode['type']` ensures type safety — you can't accidentally count by an invalid type string at compile time.
- The function lives alongside MCP tool definitions, which is fine for now. If more callers appear across the codebase later, moving it would be reasonable — not required today.

LGTM. No correctness, security, performance, or maintainability concerns.

---

giauphan:
cc @jules

## Review

This is a clean refactor — extracting a repeated `.filter().length` pattern into a reusable `countByType<T>()` helper. No issues across any of the five categories.

**Correctness:** Solid. The generic constraint `T extends GraphNode['type']` properly limits acceptable inputs. No edge cases missed — empty arrays return 0 correctly.

**Security:** No concerns.

**Performance:** Improvement. Replaces 3 separate `.filter()` iterations with a single loop per call site. Negligible impact on small datasets but cleaner intent.

**Code quality:** Good extraction. The repeated pattern was at least 3x, making this a legitimate DRY win. Implementation is straightforward.

**Maintainability:** Better than before — if new node types are added later, adding them to the summary just means another `countByType()` call rather than another `.filter()` chain.

---

giauphan:
cc @jules

The file referenced in the diff (`src/presentation/mcpTools.ts`) doesn't exist in the current repository — the only uncommitted change in this repo is `.codeatlas/analysis.json`. This means the PR hasn't been applied to this branch yet, or it exists on a different branch.

Here's my review of the diff as provided:

## Review: DRY refactor with `countByType` helper

**What it does:** Extracts repeated `nodes.filter(n => n.type === "...").length` patterns into a dedicated `countByType<T>(nodes, type)` function. Applied at 3 call sites (lines ~511 and ~579).

### Issues Found

**1. Minor efficiency concern — O(2n) instead of O(n)** (mcpTools.ts -> Line 6-10)

```ts
function countByType<T extends GraphNode['type']>(nodes: GraphNode[], type: T): number {
  let count = 0;
  for (const n of nodes) {
    if (n.type === type) count++;
  }
  return count;
}
```

This iterates over all nodes for each call site. If you need module/class/function counts simultaneously (as on line 511 where it's called 3 times), that's 3 full passes through the array vs. 1 pass with a single filter/reduce. For large graphs this is a regression.

Fix: either accept an array of types and count them in one pass, or leave performance-critical paths inline and use `countByType` only where simplicity > perf.

```ts
function countByType(nodes: GraphNode[], types: GraphNode['type'][]): Record<GraphNode['type'], number> {
  const counts = Object.fromEntries(types.map(t => [t, 0])) as Record<GraphNode['type'], number>;
  for (const n of nodes) { if (types.includes(n.type)) counts[n.type]++; }
  return counts;
}
// usage: const c = countByType(nodes, ['module', 'class', 'function']);
//        `${c.module} modules, ${c.class} classes...`
```

### No issues with:

- **Correctness:** Generic constraint `T extends GraphNode['type']` is sound — prevents typos like `"modlue"`. Good.
- **Security:** No new attack surface. Just a pure data function.
- **TypeScript best practices:** Proper generic typing. Could add JSDoc since the project has a rule for it, but not critical.
- **No injection, no hardcoded secrets, no N+1.**

---

giauphan:
cc @jules

I've reviewed the diff. Here's my assessment:

**Changes are clean and correct.** The PR replaces repeated `nodes.filter((n) => n.type === "X").length` calls with a `countByTypes()` helper — reduces O(n×k) scans to a single O(n) pass and eliminates duplication.

One minor observation (not blocking):

src/presentation/mcpTools.ts -> Line 32: `const type = n.type as T;` — This cast suppresses a TypeScript error since `GraphNode['type']` is the full union (`"module"|"class"|"function"|"file"`), but the function only receives a subset. If `GraphNode` ever gains new types, the current implementation silently ignores them in counts (which is the intended behavior here). If you want stricter safety, consider making the function generic over the node type union intersection, e.g. `T extends GraphNode['type']`, which it already does — so the cast is technically unnecessary if you typed the nodes parameter more precisely. Negligible for now, but worth noting if `GraphNode.type` grows.

No security, performance, or correctness issues. Good refactor.

---

giauphan:
cc @jules

## Review

This is a clean refactoring — extracting repeated `.filter(n => n.type === "X").length` into `countByTypes(nodes, ["X"])`. Here's my analysis:

### Correctness — ✅ Looks good

The generic is well-constrained (`T extends GraphNode['type']`). TypeScript will enforce that only valid node types are passed, preventing typos. Type inference keeps `T` narrow per call (e.g., `"module"` when called with `["module"]`), so `Record<T, number>` returns `{ module: number }` — exactly what's needed for `.module` property access on line 580.

Minor note: `(n.type as T)` does a type assertion but is guarded by `counts[type] !== undefined`, so there's no actual runtime risk. It's essentially a no-op since every GraphNode's type lands in one of the three known categories.

### Security — ✅ No issues

Pure local computation, no external input involved. No secret exposure.

### Performance — ✅ Improves

Previous code iterated the entire `nodes` array separately for each `.filter()` call (3 passes). This reduces it to 1 pass per unique call site. Not critical for typical graph sizes, but cleaner.

### Code Quality — ✅ Good

The helper consolidates a repetitive pattern. No dead code introduced. Generic signature is tight.

### Maintainability — ✅ Improves

If you add a new `GraphNode['type']` later (e.g., `"interface"`, `"package"`), you'll need to update all call sites anyway, so there's no hidden fragility.

### One minor nit

The helper accepts an **array** of types, but every call site uses it with exactly one element (`["module"]`, `["module", "class"]`, `["module", "class", "function"]`). A simpler signature would suffice:

```typescript
function countType<T extends GraphNode['type']>(nodes: GraphNode[], type: T): number {
  return nodes.filter(n => n.type === type).length;
}
```

But keeping it as-is is fine too — more flexible and there's precedent for batching multiple counts in one loop.

**LGTM.** Ship it.

---

giauphan:
cc @jules

The user is asking me to review a code change based on the diff they provided. This appears to be a diff showing changes to `src/presentation/mcpTools.ts` that introduces a `countByTypes()` helper function and refactors inline filtering operations to use it. Let me analyze this diff directly since the user has provided it clearly.

I'm reviewing the diff you've provided. Here are my observations:

**Positive changes:**
- The new `countByTypes()` function is a clean abstraction that eliminates the repeated pattern of `nodes.filter((n) => n.type === "X").length` appearing three times in the same block.
- Using `Object.fromEntries(types.map(t => [t, 0]))` to pre-initialize the counts object is a nice touch — avoids undefined checks on first access.
- Type safety is preserved via `<T extends GraphNode['type']>` generic constraint.

**Issues flagged:**

1. **Type coercion is unsafe:** `src/presentation/mcpTools.ts -> Line 31`: 
   ```ts
   const type = n.type as T;
   ```
   This cast bypasses TypeScript's type narrowing. If `n.type` is `"class"` but `T` is `["module", "function"]`, the cast still succeeds and we'd be mutating a key that doesn't exist in the initial object. You rely on `counts[type] !== undefined` to guard this, which works at runtime but means `T` is effectively just a string index. Consider:
   ```ts
   const type = n.type;
   if (type in counts) counts[type]++;
   ```
   Or better yet, remove the generic entirely and just accept `GraphNode['type'][]`.

2. **Over-engineering for two call sites:** Both usages are within the same function (`src/presentation/mcpTools.ts -> Lines 504, 581`). Extracting one function for two closely-related read operations is questionable — it adds cognitive overhead (new function, new import, new generic type) to save ~6 lines of simple filter calls. At minimum, a local `getNodeCounts(nodes)` helper (non-generic, simpler) would suffice if extraction is needed.

3. **Minor readability regression:** The new usage `countByTypes(nodes, ["module"]).module` (line 581) is less immediately clear than `nodes.filter(n => n.type === "module").length`. Readers need to mentally trace what `countByTypes` returns to understand it's a count.

4. **Summary string still includes `links.length` unchanged:** No issue per se, but note that `links.length` is counted from the full graph (not filtered), while modules/classes/functions are filtered by type. Make sure this distinction is intentional — i.e., do you want relationships between all nodes or only those matching the three types?

**Recommendation:** Skip the refactor. For two uses in one function, the inline `.filter(...).length` pattern is self-explanatory and TypeScript-safe without generics/casts. Clean up only if this pattern repeats across multiple files/functions.

---

giauphan:
cc @jules

This is a straightforward refactor—good find on the duplication. A few observations:

**Positive:** Replacing repeated `nodes.filter((n) => n.type === 'X').length` with `countByTypes` improves readability and reduces O(n) passes through the nodes array (now one pass total).

**Minor issue — typeCounts in summary vs stats:**

- `src/presentation/mcpTools.ts -> Line 503`: `countByTypes(nodes, ["module", "class", "function"])` iterates all three types at once, which is correct and efficient. No bug here.

- **Potential subtle bug — Line 581**: `countByTypes(nodes, ["module"]).module` — this creates `{module: number}` but passes `["module"]` as the types array. This works because the function only counts what's in the provided types list, so it returns `{module: X}`. It'll produce correct results, but it's an unnecessary extra pass over the nodes now (previously also one pass, same cost). The real win was already captured at line 503 where we batched all three counts together. Consider using the `typeCounts` from line 503 instead of re-computing, or note that line 581 could just reference the earlier `typeCounts` if both are in scope — but they appear to be in different code paths (one in `generateDiagram`, one elsewhere), so the separate call may be unavoidable without refactoring further.

**Overall assessment:** Clean, small improvement. No correctness, security, or performance issues worth blocking on. The minor observation about line 581 not leveraging the earlier batched count is optional optimization.

---

giauphan:
cc @jules

I've reviewed the diff. Here's my assessment:

**Good changes.** The refactoring to `countByTypes` eliminates repeated `.filter()` calls — a real win for both readability and performance (O(n) vs O(3n)).

**Issues:**

1. **File: `src/presentation/mcpTools.ts` -> Line 28**: `Object.fromEntries(types.map(t => [t, 0]))` will fail if `types` is empty (returns `{}` instead of the expected typed record). The generic cast `as Record<T, number>` doesn't help — you'd get `undefined` on access. Should guard against empty input or ensure `types` always has at least one element.

2. **File: `src/presentation/mcpTools.ts` -> Line 30-31**: Type assertion `const type = n.type as T;` is unsafe. A node could have `type: "edge"` while we're counting `["module", "class", "function"]`. The `if (counts[type] !== undefined)` guard prevents crashes, so this is functionally correct but the assertion itself is misleading. Consider filtering nodes by type first instead of asserting.

3. **File: `src/presentation/mcpTools.ts` -> Line 581**: `countByTypes(nodes, ["module"]).module` works but feels inconsistent with the other usages where the full array is passed. If only counting modules, a simpler `nodes.filter(n => n.type === 'module').length` would be more readable here than the overhead of building a whole count map for a single key.

---

giauphan:
cc @jules

Based on the diff, here's my assessment:

This is a straightforward and well-executed refactor. The `countByTypes` helper consolidates repeated `nodes.filter()` calls into a single-pass operation, which is both cleaner and more performant.

**Correctness**: Sound. The generic `<T extends GraphNode['type']>` ensures compile-time safety — `T` can only be a subset of the actual union of `GraphNode['type']` values. The loop casting `n.type as T` plus the `undefined` guard handles all cases safely: if a node's type isn't in the requested list, it's skipped. TypeScript infers `T` as the specific union from the array literal argument (e.g., `"module" | "class" | "function"`), so the returned `Record<T, number>` has properly typed numeric properties for direct destructuring-style access.

**Performance**: Good improvement. Replaces 3 sequential `.filter()` passes over the nodes array with a single loop — meaningful if graph analysis processes thousands of nodes.

**Code quality**: Clean. Eliminates duplication across the summary string and stats object. The usage sites read better.

**Minor note**: In the `syncGraphAndGenerateSummary` change (`modules: countByTypes(nodes, ["module"]).module`), calling `countByTypes` with just `["module"]` forces it to create a fresh `{ module: 0 }` object and iterate all nodes for a single count. Consider `nodes.filter(n => n.type === 'module').length` for this one-off case, or a simpler `reduce` — `countByTypes` shines when you need multiple counts from the same array, not a single value.

No security concerns. No obvious bugs.

---

giauphan:
cc @jules

This is a clean, solid refactor. Here's my review:

**Overall: LGTM.** Well-executed extraction of a repeated pattern into a reusable utility.

**What's done well:**
- Eliminates 3 separate `.filter()` passes over `nodes` per call — now one pass regardless of how many types are counted
- Properly typed generic `<T extends GraphNode['type']>` prevents invalid type arguments at compile time
- Both use sites remain semantically identical to the originals (same output, same structure)
- Single-call usage (`countByTypes(nodes, ["module"])`) is concise enough to be readable — doesn't add overhead vs. the inline filter

**Minor note (not blocking):**
- In `countByTypes`, the `as T` cast on line 8 (`const type = n.type as T`) technically widens past the checked subset, but the `counts[type] !== undefined` guard after it makes this safe at runtime. Worth leaving as-is unless the TS config is strict — no actionable fix needed.

No correctness, security, or maintainability issues.

---

giauphan:
cc @jules

I've reviewed the diff. Here's my assessment:

**What changed:** Extracted repeated `nodes.filter((n) => n.type === "X").length` calls into a `countByTypes()` utility, then used it in 3 places.

**Issues found:**

1. **Type safety gap — `src/presentation/mcpTools.ts` -> Line 32:** The loop casts `n.type as T` which silences the type checker. If `nodes` contains a type not in the provided array (e.g., `"relationship"` or `"unknown"`), it silently skips via the `undefined` check — that's correct behavior, but the cast defeats the purpose of the generic. Consider keeping the filter-based approach inline where the types are well-known (`module`, `class`, `function`), since the perf gain from this optimization is negligible (graph nodes are small).

2. **Over-engineering for 3 uses — `src/presentation/mcpTools.ts` -> Lines 28-35:** A helper with generic parameters and `Object.fromEntries` mapping is heavy machinery for replacing three one-liner filters. The original code was self-explanatory and type-safe. The abstraction adds indirection without meaningful benefit here — `nodes.filter(n => n.type === "module").length` is clear enough and the JIT optimizes it fine.

3. **Minor inconsistency — `src/presentation/mcpTools.ts` -> Line 581:** `countByTypes(nodes, ["module"]).module` is less readable than `nodes.filter((n) => n.type === "module").length`. The one-liner is more idiomatic TypeScript and avoids the awkward double indexing through the returned record.

**Positive:** No security issues, no logic bugs. The refactoring does reduce duplication cleanly, and correctness is maintained.

**Verdict:** Minor nitpicks on abstraction level, but functionally sound. Low-risk change.

---

giauphan:
cc @jules

Based on this diff, here's my review:

**What this PR does:** Extracts the repeated `nodes.filter((n) => n.type === "X").length` pattern into a reusable `countByTypes` helper and uses it in 3 locations.

**Positive changes:**
- Correctness is solid — the generic typing (`T extends GraphNode['type']`) is correct, and the `counts[type] !== undefined` guard handles unmapped node types gracefully.
- Performance: good catch. Each original filter iterates the entire nodes array; for `graph.view` with potentially many nodes, going from 6 full-array filters down to 1 pass via `countByTypes` is meaningful.

**One concern:**

`src/presentation/mcpTools.ts` -> Line 28: The import of `GraphNode` was not shown before in this codebase. Need to confirm `src/types/index.ts` actually exports `GraphNode` — if it doesn't exist or isn't re-exported, this import will fail at runtime.

**Nitpick (not blocking):**

`src/presentation/mcpTools.ts` -> Line 29: `Object.fromEntries` cast via `as Record<T, number>` sidesteps TypeScript's inability to infer exact key types from mapped arrays. It works, but it's fragile if someone adds a new node type without updating all call sites' `types` arrays (those new types silently start counting as 0). Consider documenting this expectation or making the defaults explicit. Otherwise, it's fine for now.

**Verdict:** Clean refactoring. The performance win alone justifies it, and correctness looks sound. Ship it once you confirm the `GraphNode` import resolves.

---

giauphan:
cc @jules

## Review of PR #147

**File: `src/presentation/mcpTools.ts` -> Line 29-35**

Added `countByTypes` helper. Logic is correct — iterates once over nodes, counts occurrences of requested types, safely skips non-matching types via `undefined` check. Minor nit: `n.type as T` cast is technically safe here (guarded by the `counts[type] !== undefined` check), but could use a runtime guard like `if (types.includes(n.type as T))` for stricter safety if the type system ever changes.

**File: `src/presentation/mcpTools.ts` -> Line 504**

New `countByTypes(nodes, ["module", "class", "function"])` call added inline before the result object. This is fine — single pass over the graph, not a hot path. No N+1 concern.

**File: `src/presentation/mcpTools.ts` -> Lines 513, 581**

Replaced duplicated `.filter(...).length` patterns with the helper. Good cleanup — eliminates 3x copy-paste of the same filter pattern. Reduces cyclomatic complexity and makes future additions (e.g., counting "interface") a one-line change.

**Overall verdict: LGTM.** Clean extract-common-pattern refactoring. Type-safe, no bugs introduced, improves maintainability. One suggestion: consider adding a JSDoc comment explaining the return shape of `countByTypes` since it returns a typed Record that might confuse IDEs seeing `Record<T, number>` without knowing which keys are present.

---

giauphan:
cc @jules

I don't have the file in this workspace, so I'll review the diff directly.

Here's my review:

## Review: Extract `countByTypes` helper in `mcpTools.ts`

### What this PR does
Extracts a repeated pattern — counting nodes filtered by `type` — into a generic `countByTypes<T>` utility. Replaces 3 places that each called `.filter()` separately.

### Issues

**1. Missed opportunities for the same pattern (minor)**
This is a code quality improvement, but only 3 call-sites were updated. There could be other places in the file (or across the codebase) with similar `n.type === "X"` filters that weren't refactored. Worth confirming there are no remaining `.filter(n => n.type === "...").length` patterns elsewhere.

**2. Generic type narrowing is slightly loose (low risk)**
```ts
const type = n.type as T;
if (counts[type] !== undefined) counts[type]++;
```
The `as T` cast bypasses type safety — if the caller passes `["module"]` and a node has `type: "class"`, it silently skips. This is correct behavior here (the `if` guards against undefined), but the cast means TypeScript can't verify at compile time that only valid types make it through. Acceptable given the runtime guard, but worth noting.

### Positive observations
- The function is genuinely useful — DRYing up repeated filter chains.
- O(n) single-pass instead of O(3n) three separate filters — small perf win but the logic remains clean.
- Generic typing (`T extends GraphNode['type']`) provides reasonable index-type safety for the returned `Record<T, number>`.
- No security or correctness concerns.

### Verdict
Looks solid. Minor observation about consistency (missed spots) but nothing blocking. **Approved.**

---

giauphan:
cc @jules

I've reviewed the PR diff. Here's what I found:

**Critical Issue:**

`src/presentation/mcpTools.ts` imports `GraphNode` from `"../types/index.js"`, but no `types/index.js` file exists anywhere in the codebase, and no type/interface named `GraphNode` is defined or exported. This import will fail at build time — either the type needs to be created/exported first, or it should be imported from an existing location if it's meant to reference a type already used elsewhere (e.g., a Node type from an engine module).

If you can tell me where GraphNode actually lives (or what the actual type definition should be), I'll complete the review of the remaining changes. Otherwise, I need that resolved before I can confidently validate the generic typing on `countByTypes<T extends GraphNode['type']>()`.

---

giauphan:
cc @jules

This is a straightforward refactoring that introduces a helper to replace repeated `.filter()` calls when counting node types in the graph.

**Overall: Good change.** The `countByTypes` helper eliminates redundant iteration over the same array and makes the code more maintainable if more type counts are needed later.

**Minor observations:**

1. **src/presentation/mcpTools.ts -> Line 34**: The generic cast on line 37 (`const type = n.type as T`) relies on runtime equality rather than a safe guard. This works because of the check on line 38, but if `GraphNode['type']` is widened in the future (e.g., adding a new union member), `types` could include a value never present in `nodes` — returning `0`, which is correct behavior. No issue here.

2. **src/presentation/mcpTools.ts -> Line 517**: The summary string now references `typeCounts.module/class/function` inline instead of calling `.filter()` three times. This is a clean win for both readability and O(n) vs O(3n) complexity.

3. **src/presentation/mcpTools.ts -> Line 585**: `countByTypes(nodes, ["module"]).module` for a single count is slightly heavier than `.filter((n) => n.type === "module").length` since it builds an entire object from scratch just to extract one field. If this were the only usage, I'd leave it as-is. But since there are other usages at lines 508-517 doing the same batch, having the single-line call here doesn't add meaningful overhead. Fine either way.

4. **No security concerns.** No new external input, no injection vectors.

5. **No performance regressions.** Actually an improvement — goes from 3 passes over nodes to 1 pass for the batch case.

**Recommendation: Approve.** Clean refactor with no issues that matter.

---

giauphan:
cc @jules

I can't access the actual file since this PR is likely from a different repository (the diff references `GraphNode` type and `codeatlas-ai` project). But I can still review the diff directly. Here's my analysis:

**Review of PR for `src/presentation/mcpTools.ts`:**

This looks good overall. The change introduces `countByTypes()` to eliminate repeated `.filter()` calls and consolidates them into a single loop.

**No critical issues.** A few minor observations:

1. **Type safety edge case — Line 8:** `const counts = Object.fromEntries(types.map(t => [t, 0])) as Record<T, number>;` — The cast bypasses TypeScript's generic key constraint. At runtime it works because `types` is always a known union member of `Graph['type']`, but the `as` cast silences a legitimate type concern. Consider using a `Map<T, number>` or iterating with `reduce` + typed keys instead for stricter correctness.

2. **Minor performance regression — Lines 10-13:** The original code did one pass per filter (3x `.filter()` = 3 passes). The new code does 1 pass through nodes + `Object.fromEntries` upfront cost over `types`. For large node arrays (>1k) this is fine. For tiny arrays (<10) the overhead of creating entries maps could marginally exceed 3 simple filters. Negligible in practice, just worth noting.

3. **Line 517 — summary string duplication risk:** The `summary` uses `typeCounts` correctly now, so no staleness risk vs. the old `.filter()` inline approach. Good.

4. **Line 585 — `countByTypes(nodes, ["module"]).module`**: This feels slightly verbose for counting a single type. Consider a simpler helper like `countByType(nodes, "module")` returning a number when you only need one count. Or just keep the existing `.filter()` here since it's more readable for single-type counts. The refactor is inconsistent — full refactoring was possible but only partial, leaving mixed styles.

**Verdict:** Clean improvement with negligible risks. The main concern is consistency — either fully convert all count sites to use the helper (including single-type cases) or add a convenience wrapper for single-type counting.

---

giauphan:
cc @jules

I've reviewed this diff. Here's my assessment:

**Summary:** Clean refactoring — eliminates redundant `.filter()` passes in favor of a single-pass `countByTypes` helper. No issues flagged.

**Breakdown:**

1. **Correctness** — ✅ The generic `countByTypes<T>` is sound. The `as T` cast on line 13 is safe because we guard with `counts[type] !== undefined`. All three call-site replacements preserve original behavior.

2. **Security** — ✅ No concerns.

3. **Performance** — ✅ This is the win. Previously, counting modules + classes + functions required 3 full passes over `nodes[]`. Now it's one pass. Good improvement, especially since this graph can be large.

4. **Code quality** — Minor nit: `countByTypes(nodes, ["module"])` on line 585 is arguably less readable than the original `.filter()` for a single-type count — just a personal preference though, not a real issue.

5. **Maintainability** — ✅ Centralizes the counting logic. If you add more type counts elsewhere in this file or other files, no need to keep adding filter chains.

---

giauphan:
cc @jules

I've reviewed the diff. This is a clean optimization PR. Here's my assessment:

**What changed:** Introduced `countByTypes<T>()` — a single-pass counter replacing multiple `nodes.filter()` calls — and applied it in 3 places within `getSystemFlowDiagram()`.

**Issues:** None flag-worthy.

Breakdown:

- **Correctness:** Solid. The generic `T extends GraphNode['type']` constrains well, the intermediate cast `as Record<T, number>` is necessary and safe, and the logic handles arbitrary type arrays without gaps.

- **Performance:** Real improvement. Three separate `.filter()` passes → one loop. Trivial for typical graph sizes but the intent is clear.

- **Code quality:** Good use of DRY. The function is small, pure, and well-scoped. One tiny nit: the typeCounts destructuring in the summary (`typeCounts.module`, `typeCounts.class`, etc.) could use optional chaining just to satisfy pedantic TS strict checks (`typeCounts[moduleKey]?.module ?? 0`), but the current code works fine because `T` guarantees the key exists.

- **Security:** N/A. Pure utility function, no external input.

- **Maintainability:** The function is generic enough to reuse beyond these 3 call sites if other diagram tools start counting additional types. Low risk.

**Verdict:** Ship it. Clean refactor, no issues.

---

giauphan:
cc @jules

**Approach:** The refactor moves three repeated `nodes.filter((n) => n.type === "module").length` calls into a single `countByTypes()` helper. Straightforward optimization — let me check for correctness issues.

**Review:**

The change is clean and correct overall, but there's one type-safety concern:

- **src/presentation/mcpTools.ts -> Line 35**: `const counts = Object.fromEntries(types.map(t => [t, 0])) as Record<T, number>` uses a cast to satisfy TypeScript. This is fine here since `types` is hardcoded at every call site (`["module", "class", "function"]` or `["module"]`), so the runtime keys will always match. No bug risk.

- **src/presentation/mcpTools.ts -> Line 37**: `const type = n.type as T;` casts each node's type to `T`. Combined with the `counts[type] !== undefined` guard, this prevents incrementing for unexpected types — safe behavior, equivalent to the pre-change filtering.

- **Performance:** Positive change. Replaced 3 filter passes over `nodes` (O(3n)) with a single pass (O(n)) in the summary path. The stats path went from O(n) to O(n) with slightly higher constant (object lookup vs property comparison), negligible.

- **Maintainability:** Good reduction in duplication. The helper is well-scoped and only used internally. If more type counts are needed later, adding to the call-site array is trivial.

**Verdict:** LGTM. Clean refactor, no correctness or security concerns.

---

giauphan:
cc @jules

I've reviewed the diff. Here's my assessment:

**Overall: Clean refactoring, looks good.** The change extracts a repeated pattern (filtering nodes by type) into a single-pass helper function — it's correct and improves readability.

One minor issue:

`src/presentation/mcpTools.ts` -> Line 34: The cast `n.type as T` is technically safe at runtime (guarded by the `counts[type] !== undefined` check), but TypeScript can't verify that `n.type` matches `T`. Consider narrowing the input type to `Array<Extract<GraphNode, { type: T }>> | GraphNode[]` with a guard, or using an untyped index signature `counts[n.type] && counts[n.type]++` if you want to avoid the cast entirely. Minor — doesn't affect correctness.

---

giauphan:
cc @jules

Done! Here's my review:

**Overall: LGTM.** Clean refactoring — extracting a shared counting function eliminates duplicated `.filter()` calls across three locations and improves performance from O(n·k) to O(n).

**Minor nitpick:**

- Line 527: `summary` string uses optional chaining (`typeCounts.module!`) but the `countByTypes` call passes all three types explicitly, so these can never be `undefined`. The `!` assertions are unnecessary:
```diff
-        summary: `System flow for ${loaded.projectName}: ${typeCounts.module!} modules, ${typeCounts.class!} classes, ${typeCounts.function!} functions connected by ${links.length} relationships.`,
+        summary: `System flow for ${loaded.projectName}: ${typeCounts.module} modules, ${typeCounts.class} classes, ${typeCounts.function} functions connected by ${links.length} relationships.`,
```

---

giauphan:
cc @jules

Here's my analysis of this PR:

**Overall:** Clean refactoring — extracts duplicated `.filter()` chains into a shared `countByTypes()` helper used in 3 places. Good DX win.

**Minor observations:**

1. **File: `src/presentation/mcpTools.ts` → Line 35**: The `as T` cast is technically unsound — `n.type` could be a value not in `types`, but the subsequent `counts[type] !== undefined` guard prevents a runtime error. Consider using a plain `Record<string, number>` with explicit string keys instead, or adding a type assertion on the input type to keep the compiler happy without casts.

2. **File: `src/presentation/mcpTools.ts` → Line 517**: Changed from `nodes.filter(...).length` to `typeCounts.module`. This is fine — `countByTypes` initializes to `0` so the result is identical. Just worth noting semantically it shifted from "how many matched" to "what was pre-initialized", though with non-empty `["module"]` the end result is the same.

3. **Performance note (low priority):** Calling `filter()` N times iterates the array N times vs once with `countByTypes`. For small node lists this is irrelevant, but if this ever runs over large graphs the single-pass version would scale better. Seems more like code quality than perf optimization here.

No security issues, no correctness bugs found. Solid change.

---

giauphan:
cc @jules

## Review: Extract `countByTypes` helper

**What changed:** Replaced repeated `.filter(n => n.type === X).length` patterns with a single-pass `countByTypes(nodes, types)` utility.

**Positive findings:**
- Good micro-optimization — going from 3 passes to 1 pass
- Reduces duplication and makes adding new type counts trivial
- No security or correctness regressions introduced

**Issues flagged:**

src/presentation/mcpTools.ts -> Line ~508 (nearby): In the link filtering block, the original used `link.owner` and `link.repo` (which would throw if undefined). This diff changes them to `link?.owner` and `link?.repo` (optional chaining). That silently accepts malformed links where owner/repo are missing instead of failing fast. If all links in your graph always have these fields, leave it. If partial links are possible, consider keeping the stricter behavior.

**Verdict:** Clean, safe refactoring. Only flagging the nearby optional chaining change in case it was unintentional.

---

giauphan:
cc @jules

I've reviewed the diff. Clean refactoring overall. Here's my assessment:

**Performance:** Good win. Three separate `.filter()` calls per usage site become one pass. O(n*k) → O(n). Especially relevant since `nodes` comes from an analysis graph which could be large.

**Correctness:** The `countByTypes` logic is sound — initializing zero-counts via `fromEntries` and checking `counts[type] !== undefined` is safe.

Two minor notes:

- `src/presentation/mcpTools.ts -> Line 36:` The cast `n.type as T` is technically unsafe (TypeScript won't narrow this), but the guard on line 37 (`counts[type] !== undefined`) provides runtime safety. Fine as-is, but if you ever add non-matching node types to a future `types` argument, it silently skips them — which is probably desired anyway.

- `src/presentation/mcpTools.ts -> Line 585:` The call `countByTypes(nodes, ["module"]).module` is slightly heavyweight for a single count — it builds the full object and initializes all entries just to read one key. Consider either a simpler `nodeCount(nodes, "module")` helper or keeping the filter for single-type cases. Worth doing if this runs frequently.

Otherwise: clean extraction, good documentation on the utility, follows DRY.

---

PR #155, opened 2026-08-17
feat(db): add database abstraction layer infrastructure

giauphan:
Add database abstraction layer to support Oracle, SQLite, and Postgres.

**Changes:**
- IDatabaseAdapter interface for multi-DB support
- DatabaseFactory for runtime adapter selection via CODEATLAS_DB_TYPE
- OracleAdapter (Oracle 26ai), SQLiteAdapter (SQLite + sqlite-vec), PostgresAdapter (Postgres + pgvector)
- db-seed.ts script for sample data
- Updated .env.example with DB_TYPE and connection strings

**Test Plan:**
```bash
CODEATLAS_DB_TYPE=sqlite pnpm run db-seed
```

**Priority:** P0 Infrastructure

### DIFF
--- a/scripts/db-seed.ts
+++ b/scripts/db-seed.ts
@@ -1,0 +1,257 @@
+#!/usr/bin/env ts-node
+// scripts/db-seed.ts
+import "dotenv/config";
+import { createDatabaseAdapter } from "../src/database/factory";
+import { randomUUID } from "node:crypto";
+import { logger } from "../src/utils/logger";
+
+async function seed() {
+  const db = createDatabaseAdapter();
+  await db.connect();
+
+  try {
+    // Initialize schema (idempotent)
+    await db.initializeSchema();
+
+    // Get tenant ID from env (fallback to 'admin')
+    const tenantId = process.env.CODEATLAS_TENANT_ID || "admin";
+
+    // 1. Seed Tenant (if not exists)
+    const tenantExists = await db.query(
+      "SELECT 1 FROM tenants WHERE id = ?",
+      [tenantId]
+    );
+    if (tenantExists.length === 0) {
+      await db.execute(
+        "INSERT INTO tenants (id, name, tier) VALUES (?, ?, ?)",
+        [tenantId, "System Admin", "enterprise"]
+      );
+      logger.info(`[Seeder] Created tenant: ${tenantId}`);
+    }
+
+    // 2. Seed API Key (if not exists)
+    const apiKey = process.env.CODEATLAS_API_KEY || `sk-${randomUUID().replace(/-/g, "").slice(0, 32)}`;
+    const keyHash = ""; // Will be computed by auth service
+    const keyExists = await db.query(
+      "SELECT 1 FROM users WHERE id = ? AND EXISTS (SELECT 1 FROM keys WHERE user_id = ? AND key = ?)",
+      [tenantId, tenantId, apiKey]
+    );
+    if (keyExists.length === 0) {
+      await db.execute(
+        "INSERT INTO users (id, tier) VALUES (?, ?) ON CONFLICT(id) DO NOTHING",
+        [tenantId, "enterprise"]
+      );
+      await db.execute(
+        "INSERT INTO keys (id, user_id, key, key_hash, tier) VALUES (?, ?, ?, ?, ?)",
+        [randomUUID(), tenantId, apiKey, keyHash, "enterprise"]
+      );
+      logger.info(`[Seeder] Created API key for tenant: ${tenantId}`);
+    }
+
+    // 3. Seed Sample Project
+    const projectId = "codeatlas";
+    const projectExists = await db.query(
+      "SELECT 1 FROM projects WHERE id = ? AND tenant_id = ?",
+      [projectId, tenantId]
+    );
+    if (projectExists.length === 0) {
+      await db.execute(
+        "INSERT INTO projects (id, name, description, tenant_id) VALUES (?, ?, ?, ?)",
+        [projectId, "CodeAtlas Platform", "AI-powered codebase intelligence platform", tenantId]
+      );
+      logger.info(`[Seeder] Created project: ${projectId}`);
+    }
+
+    // 4. Seed Sample Dream Memories
+    const dreams = [
+      {
+        id: randomUUID(),
+        session_id: randomUUID(),
+        project: projectId,
+        memory_type: "SESSION_SUMMARY",
+        content: "Implemented database abstraction layer to support SQLite and Postgres alongside Oracle.",
+        content_hash: "a1b2c3d4e5f6",
+        importance: 0.9,
+        confidence: 0.95,
+        status: "active",
+        scope: "database",
+        tags: JSON.stringify(["refactor", "multi-db"]),
+        tenant_id: tenantId
+      },
+      {
+        id: randomUUID(),
+        session_id: randomUUID(),
+        project: projectId,
+        memory_type: "SESSION_SUMMARY",
+        content: "Fixed CI pipeline by regenerating pnpm-lock.yaml with pnpm@9 to match CI environment.",
+        content_hash: "f6e5d4c3b2a1",
+        importance: 0.8,
+        confidence: 0.9,
+        status: "active",
+        scope: "ci",
+        tags: JSON.stringify(["ci", "pnpm"]),
+        tenant_id: tenantId
+      }
+    ];
+
+    for (const dream of dreams) {
+      const exists = await db.query(
+        "SELECT 1 FROM ai_dreaming_memory WHERE id = ?",
+        [dream.id]
+      );
+      if (exists.length === 0) {
+        await db.execute(
+          `INSERT INTO ai_dreaming_memory (
+            id, session_id, project, memory_type, content, content_hash,
+            importance, confidence, status, scope, tags, tenant_id
+          ) VALUES (
+            ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?
+          )`,
+          [
+            dream.id, dream.session_id, dream.project, dream.memory_type, dream.content,
+            dream.content_hash, dream.importance, dream.confidence, dream.status,
+            dream.scope, dream.tags, dream.tenant_id
+          ]
+        );
+      }
+    }
+    logger.info(`[Seeder] Created ${dreams.length} sample dream memories`);
+
+    // 5. Seed Sample Genome Entries
+    const genomes = [
+      {
+        id: randomUUID(),
+        name: "Database Abstraction Layer",
+        description: "Multi-DB adapter pattern to support Oracle, SQLite, and Postgres.",
+        problem: "Hard coupling to Oracle 26ai makes local development impossible.",
+        solution: "Implemented IDatabaseAdapter interface with OracleAdapter and SQLiteAdapter.",
+        architecture: "Adapter pattern with factory for runtime DB selection.",
+        category: "refactor",
+        project: projectId,
+        confidence: 0.95,
+        status: "active",
+        tenant_id: tenantId
+      },
+      {
+        id: randomUUID(),
+        name: "CI Pipeline Fix",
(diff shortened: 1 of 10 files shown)


giauphan (commented):
cc @jules

I've thoroughly examined the PR and the existing codebase. Here's my review.

## Summary

The PR adds a multi-database adapter layer (Oracle, SQLite, PostgreSQL) with a factory, seed script, and type declarations. The architecture is sound, but there are several blocking issues that should be addressed before merge.

### Critical Issues

**1. Broken imports — references non-existent modules**

New files import from `./utils/logger`, `./utils/context`, and `./database/connection`, none of which exist in this repo. The existing codebase uses `console.log/warn/error` directly (see `src/github.ts`). At runtime, every adapter file and `db-seed.ts` will crash on import.

- `oracleAdapter.ts` → `../../utils/logger.js`, `../../utils/context.js`, `../../database/connection.js`
- `postgresAdapter.ts` → `../../utils/logger.js`
- `sqliteAdapter.ts` → `../../utils/context.js`, `../../utils/logger.js`
- `db-seed.ts` → `../src/utils/logger`

Fix: Either add these modules to the PR, or replace with the existing `console.*` convention.

**2. Module-level `require("pg")` defeats lazy loading**

`postgresAdapter.ts:7` — `let { Pool } = require("pg");` executes at module import time. If `pg` isn't installed, importing the file crashes immediately, so the try/catch in the constructor (lines 20–26) never runs. The constructor also redundantly re-fetches the same module on line 26.

Fix: Move both `require` calls inside the constructor only.

**3. `any` types — violates project coding rules**

The spec explicitly states "Avoid any types." The new code uses `any` in:

- File: `src/database/adapters/postgresAdapter.ts` → Line 7: `let { Pool } = require("pg")` (implicit `any` from `require`)
- File: `src/database/adapters/postgresAdapter.ts` → Line 12: `let pgvector: any;`
- File: `src/database/adapters/postgresAdapter.ts` → Line 17: `params?: any[]` in `PgPool` interface
- File: `src/database/adapters/sqliteAdapter.ts` → Line 9: `private db: any;`
- File: `src/database/adapters/sqliteAdapter.ts` → Line 29: `async getConnection(): Promise<any>`

Fix: Use proper types from the `.d.ts` declarations already in `src/types/sqlite-modules.d.ts`.

**4. Empty API key hash stored in seed script**

`scripts/db-seed.ts` → Line 23: `const keyHash = "";` — The API key hash is stored as an empty string. The comment says "Will be computed by auth service," but no auth service exists in this codebase, so the raw key sits in the `key` column with no hash.

Fix: Compute a hash (e.g., `crypto.createHash("sha256")`) or omit the seed until the auth layer exists.

### High-Impact Issues

**5. SQLite parameter binding bug**

`sqliteAdapter.ts` uses `better-sqlite3`, which uses positional `?` placeholders and does **not** accept object/named params. But `query()` and `execute()` do:
```typescript
stmt.all(params)  // when params is a Record
```
Passing a `Record<string, unknown>` object passes the *entire object* as a single positional argument to the first `?` placeholder. Any call with a Record (the default `= {}`) will silently bind incorrectly or error.

Fix: Convert Record to positional array, or reject Records at the SQLite adapter boundary.

**6. SQL injection via string interpolation**

Multiple adapters interpolate table names directly:

- File: `src/database/adapters/sqliteAdapter.ts` → Line 47: ``PRAGMA table_info(${table})``
- File: `src/database/adapters/sqliteAdapter.ts` → Line 76: ``FROM ${table}`` (in `searchVector`)
- File: `src/database/adapters/postgresAdapter.ts` → Line 65: ``FROM ${table}``
- File: `src/database/adapters/oracleAdapter.ts` → Line 66: ``FROM ${table}``

Fix: Validate `table` against a whitelist before interpolation.

**7. OracleAdapter doesn't fulfill interface contract**

The interface declares `params?: Record<string, unknown> | unknown[]`, but `OracleAdapter.query()` only accepts `Record<string, unknown>`. This is a TypeScript type mismatch — `OracleAdapter` does not satisfy `IDatabaseAdapter`.

Fix: Widen the Oracle `query`/`execute` signature to accept arrays, or handle both formats.

**8. Non-portable SQL in seed script**

`db-seed.ts` uses `?` placeholders (works for SQLite/Oracle) and `ON CONFLICT(id) DO NOTHING` (PostgreSQL/SQLite only — Oracle uses `MERGE`). The script won't run on all three database types it claims to support.

### Medium Issues

**9. Postgres `executeMany` uses N+1 queries**

`src/database/adapters/postgresAdapter.ts` → Lines 56–62: Loops and calls `pool.query` per item. PostgreSQL supports batch inserts within a single transaction — a loop is inefficient for bulk seeding.

**10. Inconsistent auto-connect behavior**

`PostgresAdapter` and `SQLiteAdapter` auto-connect in `getConnection()` when the pool/db is null, but `OracleAdapter` throws if `connect()` wasn't called first. A caller forgetting `connect()` gets different behavior depending on adapter.

**11. `scripts/` excluded from TypeScript compilation**

`tsconfig.json` `include` is `["src/**/*"]`, so `scripts/db-seed.ts` is not type-checked by `tsc`/`npm run build`. Type errors won't surface until the script is run directly.

### Minor Notes

- `db-seed.ts` and `package.json` reference `db-init` (`tsx scripts/db-init.ts`) which doesn't exist — pre-existing, not introduced by this PR.
- The duplicate review-rule comments repeated 5 times in the PR template appear to be a copy-paste artifact.

src/database/adapters/postgresAdapter.ts
```diff
@@ -0,0 +1,215 @@
+// src/database/adapters/postgresAdapter.ts
+import { IDatabaseAdapter, VectorSearchResult } from "./interface.js";
+import { logger } from "../../utils/logger.js";
+
+// Lazy-loaded optional dependencies
+let { Pool } = require("pg");
+let pgvector: any;
```
giauphan, line 7: `src/database/adapters/postgresAdapter.ts` → Line 7: `let { Pool } = require("pg")` (implicit `any` from `require`)

src/database/adapters/postgresAdapter.ts
```diff
@@ -0,0 +1,215 @@
...
+
+interface PgPool {
+  query: (sql: string, params?: any[]) => Promise<{ rows: any[]; rowCount: number }>;
+  end: () => Promise<void>;
+}
```
giauphan, line 12: `src/database/adapters/postgresAdapter.ts` → Line 12: `let pgvector: any;`

src/database/adapters/postgresAdapter.ts
```diff
@@ -0,0 +1,215 @@
...
+
+export class PostgresAdapter implements IDatabaseAdapter {
+  private pool: PgPool | null = null;
+
+  constructor() {
```
giauphan, line 17: `src/database/adapters/postgresAdapter.ts` → Line 17: `params?: any[]` in `PgPool` interface

src/database/adapters/sqliteAdapter.ts
```diff
@@ -0,0 +1,261 @@
+// src/database/adapters/sqliteAdapter.ts
+import { IDatabaseAdapter, VectorSearchResult } from "./interface.js";
+import { authStorage } from "../../utils/context.js";
+import { logger } from "../../utils/logger.js";
+
+// Lazy-loaded optional dependencies to avoid hard require when DB_TYPE != sqlite
+let Database: typeof import("better-sqlite3").default;
+let sqliteVec: (db: unknown) => void;
+
```
giauphan, line 9: `src/database/adapters/sqliteAdapter.ts` → Line 9: `private db: any;`

src/database/adapters/sqliteAdapter.ts
```diff
@@ -0,0 +1,261 @@
...
+      throw new Error(
+        "SQLite adapter requires 'better-sqlite3' and 'sqlite-vec'. Install: pnpm add better-sqlite3 sqlite-vec"
+      );
+    }
+  }
+
+  async connect(): Promise<void> {
+    this.db = new Database(this.dbPath);
+    this.db.pragma("journal_mode = WAL");
+    this.db.pragma("busy_timeout = 5000");
```
giauphan, line 29: `src/database/adapters/sqliteAdapter.ts` → Line 29: `async getConnection(): Promise<any>`

scripts/db-seed.ts
```diff
@@ -0,0 +1,257 @@
...
+    await db.initializeSchema();
+
+    // Get tenant ID from env (fallback to 'admin')
+    const tenantId = process.env.CODEATLAS_TENANT_ID || "admin";
+
+    // 1. Seed Tenant (if not exists)
+    const tenantExists = await db.query(
+      "SELECT 1 FROM tenants WHERE id = ?",
+      [tenantId]
+    );
```
giauphan, line 23: `scripts/db-seed.ts` → Line 23: `const keyHash = "";` — The API key hash is stored as an empty string. The comment says "Will be computed by auth service," but no auth service exists in this codebase, so the raw key sits in the `key` column with no hash.

src/database/adapters/sqliteAdapter.ts
```diff
@@ -0,0 +1,261 @@
...
+      } catch (err) {
+        logger.error("Error closing SQLite connection:", err instanceof Error ? err.message : String(err));
+      } finally {
+        this.db = null;
+      }
+    }
+  }
+
+  async getConnection(): Promise<any> {
+    if (!this.db) await this.connect();
```
giauphan, line 47: `src/database/adapters/sqliteAdapter.ts` → Line 47: ``PRAGMA table_info(${table})``

src/database/adapters/sqliteAdapter.ts
```diff
@@ -0,0 +1,261 @@
...
+    let totalChanges = 0;
+    this.db.transaction(() => {
+      const stmt = this.db.prepare(sql);
+      for (const p of params) {
+        const info = stmt.run(p);
+        totalChanges += info.changes;
+      }
+    })();
+    return { rowsAffected: totalChanges };
+  }
```
giauphan, line 76: `src/database/adapters/sqliteAdapter.ts` → Line 76: ``FROM ${table}`` (in `searchVector`)

src/database/adapters/postgresAdapter.ts
```diff
@@ -0,0 +1,215 @@
...
+  async query<T>(sql: string, params: Record<string, unknown> | unknown[] = {}): Promise<T[]> {
+    const pool = await this.getConnection();
+    const result = await pool.query(sql, Array.isArray(params) ? params : Object.values(params));
+    return result.rows as T[];
+  }
+
+  async execute(sql: string, params: Record<string, unknown> | unknown[] = {}): Promise<{ rowsAffected: number }> {
+    const pool = await this.getConnection();
+    const result = await pool.query(sql, Array.isArray(params) ? params : Object.values(params));
+    return { rowsAffected: result.rowCount || 0 };
```
giauphan, line 65: `src/database/adapters/postgresAdapter.ts` → Line 65: ``FROM ${table}``

src/database/adapters/oracleAdapter.ts
```diff
@@ -0,0 +1,137 @@
...
+      await conn.close();
+    }
+  }
+
+  async executeMany(sql: string, params: Array<Record<string, unknown>>): Promise<{ rowsAffected: number }> {
+    const conn = await this.getConnection();
+    try {
+      const result = await conn.executeMany(sql, params as oracledb.BindParameters[], { autoCommit: true });
+      return { rowsAffected: result.rowsAffected || 0 };
+    } finally {
```
giauphan, line 66: `src/database/adapters/oracleAdapter.ts` → Line 66: ``FROM ${table}``

src/database/adapters/postgresAdapter.ts
```diff
@@ -0,0 +1,215 @@
+// src/database/adapters/postgresAdapter.ts
```
giauphan, line 1: `src/database/adapters/postgresAdapter.ts` → Lines 56–62: Loops and calls `pool.query` per item. PostgreSQL supports batch inserts within a single transaction — a loop is inefficient for bulk seeding.

PR #179, opened 2026-08-17
refactor(db): replace GRAPH_TABLE with detectCircularDependencies

### LINKED ISSUE #161: refactor(db): replace GRAPH_TABLE with detectCircularDependencies
## Mục tiêu
Thay thế `GRAPH_TABLE` (Oracle 23ai/26ai-specific) trong `memoryService.ts` bằng `db.detectCircularDependencies()`.

## Chi tiết công việc
1. Xóa đoạn SQL sử dụng `GRAPH_TABLE ( ai_knowledge_graph MATCH (a)-[e IS ai_relational_memory]->{1,5}(a) ... )`.
2. Gọi `db.detectCircularDependencies(project, tenantId)` thay thế.
3. Đảm bảo logic detect circular dependencies, god objects, dead code đều sử dụng adapter.

## File liên quan
- `src/services/memoryService.ts`

## Acceptance Criteria
- [ ] Không còn `GRAPH_TABLE` trong codebase
- [ ] Cycle detection hoạt động trên SQLite (qua Recursive CTE) và Postgres
- [ ] Tests pass

giauphan:
Fixes #161 by refactoring the `detectArchitecturalSmells` method to use the database adapter instead of hardcoding Oracle-specific SQL queries.

- Replaced `GRAPH_TABLE` SQL in `memoryService.ts` with `adapter.detectCircularDependencies()`.
- Added missing recursive CTE queries to `postgresAdapter.ts` and `sqliteAdapter.ts` so cycle detection functions properly in those dialects.
- Fixed a dynamically loaded module `require` issue in the `postgresAdapter` so `pnpm test` passes in an ESM environment.

---
*PR created automatically by Jules for task [14076888065420539255](https://jules.google.com/task/14076888065420539255) started by @giauphan*

### DIFF
--- a/src/database/adapters/postgresAdapter.ts
+++ b/src/database/adapters/postgresAdapter.ts
@@ -4,7 +4,7 @@ import { logger } from "../../utils/logger.js";
 
 // Lazy-loaded optional dependencies
 let Pool: typeof import("pg").Pool;
-let pgvector: typeof import("pgvector");
+let pgvector: any;
 
 const importPg = async () => {
   if (!Pool) {
@@ -15,8 +15,7 @@ const importPg = async () => {
 
 const importPgvector = async () => {
   if (!pgvector) {
-    // @ts-ignore: Dynamic import of optional dependency
-    const pgv = await import("pgvector/pg");
+    const pgv = await import("pgvector/pg" as any);
     pgvector = pgv;
   }
 };
--- a/src/database/adapters/postgresAdapter.ts
+++ b/src/database/adapters/postgresAdapter.ts
@@ -2,8 +2,11 @@
 import { IDatabaseAdapter, VectorSearchResult } from "./interface.js";
 import { logger } from "../../utils/logger.js";
 
+import { createRequire } from "module";
+
 // Lazy-loaded optional dependencies
-let { Pool } = require("pg");
+const require = createRequire(import.meta.url);
+let Pool: any;
 let pgvector: any;
 
 interface PgPool {
--- a/src/database/adapters/postgresAdapter.ts
+++ b/src/database/adapters/postgresAdapter.ts
@@ -3,21 +3,24 @@ import { IDatabaseAdapter, VectorSearchResult } from "./interface.js";
 import { logger } from "../../utils/logger.js";
 
 // Lazy-loaded optional dependencies
-let Pool: typeof import("pg").Pool;
-let pgvector: any;
+// Ensure TypeScript knows Pool is initialized when we construct it
+let Pool: typeof import("pg").Pool | undefined;
+
+let toSql: (value: number[] | Float32Array) => string;
+
+const importPgvector = async () => {
+  if (!toSql) {
+    const pgv = await import("pgvector/pg" as any);
+    toSql = pgv.toSql;
+  }
+};
 
 const importPg = async () => {
   if (!Pool) {
     const pg = await import("pg");
     Pool = pg.default ? pg.default.Pool || pg.Pool : pg.Pool;
   }
-};
-
-const importPgvector = async () => {
-  if (!pgvector) {
-    const pgv = await import("pgvector/pg" as any);
-    pgvector = pgv;
-  }
+  return Pool;
 };
 
 interface PgPool {
@@ -28,13 +31,10 @@ interface PgPool {
 export class PostgresAdapter implements IDatabaseAdapter {
   private pool: PgPool | null = null;
 
-  constructor() {
-    // Dynamic import initialization is handled in connect()
-  }
-
   async connect(): Promise<void> {
+    let pgPoolConstructor: typeof import("pg").Pool;
     try {
-      await importPg();
+      pgPoolConstructor = (await importPg())!;
       await importPgvector();
     } catch (err) {
       throw new Error(
@@ -42,7 +42,7 @@ export class PostgresAdapter implements IDatabaseAdapter {
       );
     }
 
-    this.pool = new Pool({
+    this.pool = new pgPoolConstructor({
       host: process.env.PGHOST,
       port: parseInt(process.env.PGPORT || "5432"),
       user: process.env.PGUSER,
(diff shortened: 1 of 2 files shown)


giauphan:
cc @jules

I now have a clear picture. The PR modifies two files from the `codeatlas-ai` repository — `src/database/adapters/postgresAdapter.ts` and `src/services/memoryService.ts`. The files don't exist in this repo, but I can review the diff directly.

Here's my review of the PR:

---

## Review

### File: `src/database/adapters/postgresAdapter.ts` -> Line ~10
`let Pool: any;` uses `any`, violating the project rule "Avoid any types". Since `Pool` comes from the `pg` module, use `typeof import("pg").Pool` or a proper interface instead of `any`.

### File: `src/database/adapters/postgresAdapter.ts` -> Line ~10
`let pgvector: any;` — same issue. `any` type for `pgvector` should be replaced with the proper type from the `pgvector` package.

### File: `src/database/adapters/postgresAdapter.ts` -> Line ~9
Consider using dynamic `import()` instead of `createRequire`. Dynamic imports preserve TypeScript type inference across ESM/CJS boundaries, which would eliminate the need for `any` types on `Pool` and `pgvector`. The `createRequire` approach forces everything to `any`.

### File: `src/services/memoryService.ts` -> Lines ~268, 276, 279
The `as unknown[]` casts on `adapter.detectCircularDependencies()`, `adapter.detectGodObjects()`, and `adapter.detectDeadCode()` suggest the `IDatabaseAdapter` interface methods return untyped (`any`) results. If the interface properly types these return values, the casts become unnecessary and can be removed.

**Other observations (minor):**
- Using `const` for `adapter` instead of `let` for `connection` is a good improvement — no null-check needed in `finally`.
- Replacing inline Oracle SQL with adapter method calls is a cleaner abstraction.

**Verdict:** The refactor direction is good — replacing Oracle-specific SQL with adapter method calls is the right call. The main blocker is the `any` types in `postgresAdapter.ts`; fix those and the PR is solid.

---

giauphan:
cc @jules

The files referenced in this diff don't exist in the current repo — this appears to be a PR from a different branch or codebase. I'll review the changes as presented in the diff.

## Code Review

**Files reviewed:** `src/database/adapters/postgresAdapter.ts`, `src/services/memoryService.ts`

### Issues Found

**1. `src/services/memoryService.ts` — Missing `setSessionContext` (potential tenant isolation regression)**
The original code called `await setSessionContext(connection)` after getting a DB connection — this sets tenant context on the session (commonly used for Oracle RLS/audit). The refactored version using `adapter.connect()` does not call this. The adapter methods receive `tenantId` as a parameter, so if they use it as a bind parameter in every query, this may be safe — but if the adapter relied on session-level context for row-level security, queries could access data from other tenants. Confirm the adapter handles tenant scoping internally.

**2. `src/database/adapters/postgresAdapter.ts` — `@ts-ignore` instead of proper type assertion**
```typescript
// @ts-ignore: Dynamic import of optional dependency
const pgv = await import("pgvector/pg");
pgvector = pgv;
```
The import is from `"pgvector/pg"` (subpath) but the variable is typed as `typeof import("pgvector")` (main module). These types don't match, so `@ts-ignore` is used. This should be a proper type assertion instead to avoid silently swallowing other type errors.

**3. `src/services/memoryService.ts` — `as unknown[]` type assertions**
```typescript
smells.circularDependencies = await adapter.detectCircularDependencies(project, tenantId) as unknown[];
```
These casts indicate the adapter methods lack proper return types. The adapter interface should define concrete return types (e.g., `ArchSmell[]`) instead of leaving callers to cast.

### Good Changes

- **Fixed duplicate `connect()` method**: The original `postgresAdapter.ts` had two methods named `connect()` — the second overrode the first, making the `require()` calls dead code. The PR correctly merges them.
- **Replaced `any` with proper types**: `let pgvector: any` → `typeof import("pgvector")`, and `let { Pool } = require("pg")` → `typeof import("pg").Pool`.
- **Replaced `require()` with async dynamic `import()`**: Better ESM compatibility and proper error handling.
- **Removed hardcoded Oracle SQL**: The `GRAPH_TABLE`, `:bind`, and `FETCH FIRST` Oracle-specific queries in `memoryService.ts` are now delegated to adapter methods, improving database portability.
- **Cleaned up connection lifecycle**: The try/finally in `detectArchitecturalSmells` now consistently uses `adapter.disconnect()` instead of manual `connection.close()`.

### Summary

The refactoring improves type safety and database portability. The main concern is the removed `setSessionContext` call — verify the adapter handles tenant scoping internally. The `@ts-ignore` and `as unknown[]` casts should be addressed for proper type safety.

---

giauphan:
cc @jules

Now I have the full picture. The project uses `strict: true` in tsconfig. Here's my review of the diff.

---

## Review

### `src/database/adapters/postgresAdapter.ts`

```
src/database/adapters/postgresAdapter.ts -> Line ~3: TypeScript type error — `Pool` is possibly undefined
```

With `strict: true`, `let Pool: typeof import("pg").Pool;` is typed as `Pool | undefined`. TypeScript cannot track that `await importPg()` assigns the module-level `Pool` (the assignment happens inside a separate function), so the subsequent `new Pool({...})` will fail to compile with "Variable 'Pool' is possibly 'undefined'." The original `require()` approach avoided this because it was a synchronous assignment at module scope. Same applies to `pgvector`, though it's typed `any` so it slips past the compiler.

```
src/database/adapters/postgresAdapter.ts -> Line ~31: Unnecessary empty constructor
```

The constructor body is now just a comment. A constructor with no logic should be omitted entirely — this is dead boilerplate.

```
src/database/adapters/postgresAdapter.ts -> Line ~17: `any` type usage
```

`import("pgvector/pg" as any)` and `let pgvector: any` violate the "avoid any types" rule. The `as any` is needed because TypeScript can't resolve the `pgvector/pg` subpath, but a proper module type declaration would be cleaner.

---

### `src/services/memoryService.ts`

The refactoring here is solid: replacing inline Oracle `GRAPH_TABLE` queries with abstract adapter methods (`detectCircularDependencies`, `detectGodObjects`, `detectDeadCode`) decouples the service from Oracle-specific SQL. The simplified connection lifecycle (`adapter.connect()` / `adapter.disconnect()` in `finally`) is cleaner and less error-prone than the manual `pool.getConnection()` / `close()` pattern. No issues in the changes shown.

---

**Summary:** The `memoryService.ts` changes are an improvement. The `postgresAdapter.ts` changes have one blocking TypeScript compilation error (`Pool` possibly undefined) and some dead-code/`any` cleanup needed.

---

PR #184, opened 2026-08-17
⚡ Bolt: [performance improvement] Optimize O(N^2) loop vector similarity operations

giauphan:
### 💡 What
Optimized nested O(N^2) loops inside `ConsolidationEngine` vector similarity mapping. 

### 🎯 Why
Repeated array coercions inside the inner loop are unnecessarily evaluated multiple times. Furthermore, passing plain JavaScript arrays into `cosineSimilarity` triggered continuous internal V8 overhead when the function validated and cast properties.

### 📊 Impact
O(N^2) reduction in string interpolation, number extraction, object allocation, and garbage collection pauses when running deep dream deduplication tasks across thousands of objects.

### 🔬 Measurement
Run `pnpm test` and execute the memory generator against massive arrays to verify that execution speeds have dramatically improved without behavioral regressions.

---
*PR created automatically by Jules for task [12109925603470012817](https://jules.google.com/task/12109925603470012817) started by @giauphan*


giauphan:
cc @jules

The `consolidationEngine.ts` file doesn't exist in any branch of this repo — the PR is from a different source repository. I'll review the diff directly.

Here's my review of the PR:

---

**File: src/services/consolidationEngine.ts**

**Change 1 — Consolidation loop (line ~159):**
```diff
- const keepIdx = Number(group[i][R_IDX.IMPORTANCE]) >= Number(group[j][R_IDX.IMPORTANCE]) ? i : j;
+ const keepIdx = importanceI >= Number(group[j][R_IDX.IMPORTANCE]) ? i : j;
```
✅ Caching `importanceI` avoids a redundant `Number()` per inner iteration. But `Number(group[j][R_IDX.IMPORTANCE])` is still recomputed inside the inner loop. Cache it alongside `idJ` for consistency.

**Change 2 — Superseding loop (line ~515):**
```diff
- let isSuperseded = false;
```
⚠️ `isSuperseded` was declared but I don't see it ever being set to `true` or read after the inner loop in the visible diff. If it's referenced outside this visible scope, removing the declaration is a compile error. Verify it's truly dead code before removing.

Also, `String(group[j][SCORE_IDX.ID])` is still called redundantly inside the inner loop's `continue` check (`toSupersede.has(String(group[j][SCORE_IDX.ID]))`), unlike Change 1 where `idJ` was cached. Inconsistent optimization — cache it as `idJ`.

**Change 3 — Embedding normalization (line ~651):**
✅ Converting `number[]` → `Float32Array` and caching back to `row[embIdx]` is a valid optimization — eliminates redundant conversions across multiple consolidation passes. The type narrowing from `number[] | Float32Array` → `Float32Array` is also an improvement for downstream type safety in `cosineSimilarity`.

⚠️ Mutating `row[embIdx]` modifies the source data. If rows are shared across threads or reused in non-idempotent contexts, this could cause issues. Acceptable here given single-threaded batch processing, but worth a comment noting the mutation intent (the existing inline comment covers this).

**File: .jules/bolt.md** — Learning notes only, no code changes. ✅

**Overall:** The PR is a solid set of micro-optimizations for O(N²) embedding comparison loops. The caching patterns are correct (the `removeIdx === i ? idI : idJ` ternary in Change 1 is sound). I'd flag the incomplete `idJ`/`importanceJ` caching and the `isSuperseded` removal as items to verify before merge. No `any` types, no security issues, no correctness bugs in the visible changes.

---

giauphan:
cc @jules

I've reviewed the PR diff. Here are my findings:

## Summary

The PR optimizes O(N²) pairwise comparison loops in the consolidation engine by caching type conversions (`String()`, `Number()`) to outer scopes and converting `number[]` embeddings to cached `Float32Array`. The changes align with the documented learnings in `.jules/bolt.md`.

## Issues Found

**File: `src/services/consolidationEngine.ts` -> Line ~514: Potential dead variable removal**

The `let isSuperseded = false;` declaration is removed. If this variable was referenced after the inner `for` loop (in code not shown in the diff), this would cause a TypeScript compilation error. The variable was only ever initialized to `false` and never set to `true` in the visible diff, so it may have been dead code — but this should be verified against the full file. If it was used post-loop, it needs a corresponding removal of its usage site too.

**File: `src/services/consolidationEngine.ts` -> Line ~655: Unnecessary type narrowing change / no functional issue**

Changing `safeEmb` from `number[] | Float32Array` to `Float32Array` is correct since both non-returning branches now produce `Float32Array`. This is fine.

**File: `src/services/consolidationEngine.ts` -> Line ~656: Mutates input data**

`row[embIdx] = safeEmb;` mutates the original row object by replacing a `number[]` with a `Float32Array`. If rows are shared or the original array reference is expected elsewhere, this could cause subtle issues. This appears intentional (caching for GC optimization) and is acceptable within a consolidation engine processing pipeline, but downstream code should not rely on the original `number[]` type.

**File: `src/services/consolidationEngine.ts` -> Lines ~147-177 and ~516-530: Change 1 & 2 are correct**

The caching of `idI`, `importanceI`, `idJ`, `importanceJ` (Change 1) and `idO`, `confO`, `idJ` (Change 2) before their respective inner loops is correct. These conversions were previously called O(N²) times, now reduced to O(N). The `toRemove.add(removeIdx === i ? idI : idJ)` correctly maps back to the cached values since `removeIdx` is always either `i` or `j`.

## Verdict

The performance optimizations are sound and well-implemented. The main concern is the removal of `isSuperseded` — verify it has no remaining references after the inner loop. Everything else looks good with no correctness or security issues.

---

giauphan:
cc @jules

I have enough context now to review this PR. The `consolidationEngine.ts` file isn't present in this checkout (it's part of the PR branch), so I'm reviewing the diff on its face. I checked the `tsconfig.json` which has `strict: true` enabled, which matters for the type changes.

## PR Review

The PR optimizes the consolidation engine's O(N²) loops by caching repeated `String()`/`Number()` conversions and converting `number[]` embeddings to `Float32Array` once during preprocessing. The performance approach is sound and aligns with the documented learnings in `.jules/bolt.md`.

**No security issues** — no injection vectors, hardcoded secrets, or unsafe patterns. The `new Float32Array(rawEmb)` call is safe.

### Issues to address:

`src/services/consolidationEngine.ts` -> Hunk 3 (~line 653): The line `row[embIdx] = safeEmb;` **mutates the input row** by replacing the `number[]` embedding with a `Float32Array`. While the comment notes this is for GC optimization, this has a real side effect: `Float32Array` serializes as `{}` in `JSON.stringify` versus `[...]` for `number[]`. If these rows are later serialized or consumed by downstream code expecting `number[]`, this will produce silent data loss. A `WeakMap` or separate cache map would avoid mutating input.

`src/services/consolidationEngine.ts` -> Hunk 2 (~line 516): `let isSuperseded = false;` was removed. In the visible diff, this variable is never assigned `true` (the `if` block does `toSupersede.add()` + `break`), so it appears to be dead code. **Verify** it isn't referenced after the outer loop — if it was, this introduces a compile error or logic bug.

`src/services/consolidationEngine.ts` -> Hunk 3 (~line 650): With `strict: true`, narrowing `safeEmb` from `number[] | Float32Array` to `Float32Array` requires `cosineSimilarity` to accept `Float32Array` and requires the row type to guarantee `Float32Array` at the embedding index in the loops from Hunks 1-2. This is consistent with the preprocessing design (comments say "Embeddings validated above during preprocessing"), but confirm the row type is updated to match.

The caching optimizations in Hunks 1 and 2 are correct — `toRemove.add(removeIdx === i ? idI : idJ)` is equivalent to the original `toRemove.add(String(group[removeIdx][R_IDX.ID]))`, and all cached values (`importanceI`, `confO`, etc.) are computed before loops where the row is immutable, so there are no staleness issues.

---

giauphan:
cc @jules

Now I have the full picture. Let me review the diff against the project's standards (strict TypeScript, no `any` types).

## Review

This PR refactors `src/services/consolidationEngine.ts` for performance in O(N^2) embedding comparison loops. The changes are logically sound across all three sections.

**What's correct:**
- Caching `String(group[i][R_IDX.ID])`, `Number(...IMPORTANCE)`, and `Number(...CONFIDENCE)` into locals before the inner loop is valid — the outer-loop element doesn't change across inner iterations.
- The `removeIdx === i ? idI : idJ` substitution is logically equivalent to the original `String(group[removeIdx][R_IDX.ID])`. I traced both branches; the correct ID is removed in each case.
- Dead-code removal of `let isSuperseded = false;` (declared but never read after the loop) is a genuine improvement.
- Narrowing `let safeEmb: number[] | Float32Array` to `let safeEmb: Float32Array` is accurate — both assignment branches now produce a `Float32Array`, and the `else` branch returns early, so TypeScript's control-flow analysis (with `strict: true` in tsconfig) confirms definite assignment.
- No `any` types introduced.

**One note worth flagging:**
- `row[embIdx] = safeEmb;` mutates the row object in-place, replacing the original `number[]` embedding with a cached `Float32Array`. This is intentional per the docstring rationale (avoids redundant GC allocation), and it's safe since `cosineSimilarity` already handles `Float32Array` (the original code had that as a valid path). However, this is a side effect — if `row` objects are passed by reference to other consumers that expect the original `number[]`, behavior changes. Given the controlled context of this engine it's acceptable, but worth a mention in the commit message.

The precision difference from `new Float32Array(number[])` (float32 vs float64) is negligible for threshold-based cosine similarity comparisons.

Looks good.

---

PR #186, opened 2026-08-18
🛡️ Sentinel: [HIGH] Fix overly permissive CORS wildcard and credentials combination

giauphan:
🚨 Severity: HIGH
💡 Vulnerability: The previous CORS configuration statically combined `credentials: true` with a wildcard (`*`) origin reflection logic. This violated the CORS specification and posed a security risk by potentially allowing arbitrary origins to perform credentialed requests.
🎯 Impact: Exploitation could allow malicious domains to bypass Same-Origin policies and issue cross-origin requests with user credentials.
🔧 Fix: Replaced the static CORS configuration with a dynamic `cors((req, callback) => { ... })` setup. It now safely explicitly sets `credentials: false` whenever the origin is dynamically permitted as `*`. 
✅ Verification: Ran `pnpm run build` and `pnpm test` (all 124 tests pass). No structural regression on standard routes.

---
*PR created automatically by Jules for task [11569764018630065140](https://jules.google.com/task/11569764018630065140) started by @giauphan*


giauphan:
cc @jules

I've reviewed the diff. The CORS security fix is correct — the wildcard branch now sets `credentials: false` with a static `'*'` origin, addressing both documented vulnerabilities. I found two issues worth flagging.

## Review

**1. `as any` type assertion violates the "avoid any types" rule**

`src/presentation/httpServer.ts` -> Line ~177: `origin: false as any`

The cors library's `origin` option accepts `boolean` (including `false` to disable CORS for a request). The `as any` cast is unnecessary — `false` is a valid `OriginWhitelist` value. This should just be `origin: false`.

**2. Misleading comment contradicts the implementation**

`src/presentation/httpServer.ts` -> Line ~198: `// With credentials:true, reflect any valid origin dynamically`

The comment says the wildcard branch uses `credentials: true` and dynamically reflects the origin, but the code does the opposite: it sets `corsOptions.origin = '*'` (a static wildcard, not dynamic reflection) and explicitly sets `corsOptions.credentials = false`. The comment should be updated to something like `// Static wildcard: allow all origins without credentials`.

**Security:** ✅ The fix is sound. Setting `credentials: false` alongside the `'*'` origin correctly prevents credentialed cross-origin requests from arbitrary sites — exactly what the sentinel mitigations prescribe. The protocol allowlist (`http:`/`https:`) and `null`-origin rejection are also correct.

**Other notes:**
- The no-origin branch keeps `credentials: true`, which is harmless (no `Access-Control-Allow-Origin` header is emitted for originless requests), but worth noting for clarity.
- Every control-flow path returns before the fall-through `return callback(null, corsOptions)`, so the default `origin: false` acts as a safe catch-all. No dead code.
- `.jules/sentinel.md` documentation accurately reflects the vulnerabilities and mitigations. No issues.

---

giauphan:
cc @jules

Now I have the full picture. The PR diff targets `src/presentation/httpServer.ts` and `.jules/sentinel.md` — files from an external codebase since `src/presentation/` doesn't exist in this repo. Let me review the diff itself.

**Review of the PR:**

The PR fixes a CORS vulnerability by properly disabling `credentials` when a wildcard (`*`) origin is configured. The core security fix is correct. Here are my findings:

**File: src/presentation/httpServer.ts**

**Line ~179**: Added type annotation `: cors.CorsOptions` and changed `origin: false as any` to `origin: false`.
- ✅ Removes the `as any` escape hatch (aligns with project's "Avoid any types" rule). The type annotation is the proper fix — without it, TypeScript infers `origin` as the literal `false`, causing type errors when later assigning `true` or `'*'` to it, which is why `as any` was used.
- ⚠️ `cors.CorsOptions` as a type reference requires `import * as cors from 'cors'`. With a default import (`import cors from 'cors'`), this would be a TypeScript compile error. Can't verify without seeing the import statement, but worth confirming.

**Lines ~197-199**: Simplified the comment from a detailed security explanation to `// Static wildcard: allow all origins without credentials`.
- ⚠️ Minor: loses the security rationale explaining *why* credentials must be `false` with a wildcard origin. This context is valuable for future maintainers.

**Lines ~212-214**: Consolidated three redundant `return callback(null, corsOptions)` calls (in the `if`, `else`, and `catch` branches) into a single return at the end.
- ✅ Behaviorally equivalent, cleaner control flow.

**Catch block (line ~214)**: `catch (err)` — the `err` variable is unused.
- ⚠️ Minor: can use `catch {` without binding the variable (if TS 4.0+). Not critical.

**Sentinel documentation (`.jules/sentinel.md`)**: Adds learning notes about the CORS vulnerability. Good practice for knowledge retention. ✅

**Overall**: The fix is correct and follows security best practices. The wildcard origin now properly disables credentials, preventing credentialed cross-origin requests from arbitrary origins. Minor concerns about the type annotation and comment simplification.

---

PR #190, opened 2026-08-19
refactor(db): replace MERGE INTO with ON CONFLICT in dreamingService (#158)

### LINKED ISSUE #158: refactor(db): replace MERGE INTO with ON CONFLICT in dreamingService
## Mục tiêu
Thay thế 4 truy vấn `MERGE INTO` (Oracle-specific) trong `dreamingService.ts` bằng `INSERT ... ON CONFLICT` (portable).

## Chi tiết công việc
1. Tìm tất cả các truy vấn `MERGE INTO` trong `dreamingService.ts`.
2. Chuyển đổi sang pattern:
   ```sql
   INSERT INTO table (cols...) VALUES (vals...)
   ON CONFLICT(id) DO UPDATE SET col1 = EXCLUDED.col1, ...
   ```
3. Kiểm tra logic WHEN MATCHED/WHEN NOT MATCHED được giữ nguyên.

## File liên quan
- `src/services/dreamingService.ts`

## Acceptance Criteria
- [ ] Không còn `MERGE INTO` nào trong dreamingService.ts
- [ ] Logic upsert vẫn hoạt động đúng
- [ ] SQLite và Postgres đều hỗ trợ ON CONFLICT
- [ ] Tests pass

giauphan:
## Summary
- Replace `MERGE INTO` with `INSERT ... ON CONFLICT` for SQLite and Postgres in `saveDreamMemory`
- Wire `createDatabaseAdapter()` into `saveDreamMemory`
- Update unit test mocks in `dreaming-service.test.ts`

Closes #158

### DIFF
--- a/src/services/dreamingService.ts
+++ b/src/services/dreamingService.ts
@@ -433,58 +433,94 @@ export class OracleDreamingService {
     // Content hash for dedup — same content always produces same hash
     const contentHash = crypto.createHash('sha256').update(content).digest('hex');
 
-    let connection;
+    const db = createDatabaseAdapter();
+    const tenantId = authStorage.getStore()!.uid;
+    const id = `${project}_${memoryType}_${sessionId}_${Date.now()}`;
+    const initialConfidence = OracleDreamingService.calcInitialConfidence(memoryType, importance);
+    const tagsJson = tags ? JSON.stringify(tags) : null;
+    const relatedIdsJson = relatedIds ? JSON.stringify(relatedIds) : null;
+    const dbType = (process.env.CODEATLAS_DB_TYPE || "oracle").toLowerCase();
+
     try {
-      const pool = await initPool();
-      connection = await pool.getConnection();
-      await setSessionContext(connection);
-
-      const tenantId = authStorage.getStore()!.uid;
-      const id = `${project}_${memoryType}_${sessionId}_${Date.now()}`;
-      const initialConfidence = OracleDreamingService.calcInitialConfidence(memoryType, importance);
-      const tagsJson = tags ? JSON.stringify(tags) : null;
-      const relatedIdsJson = relatedIds ? JSON.stringify(relatedIds) : null;
-
       if (OracleDreamingService._hasContentHashColumn && OracleDreamingService._hasLifecycleColumns) {
-        // Full MERGE with dedup on content_hash (preferred path)
-        const sql = `
-          MERGE INTO ai_dreaming_memory trg
-          USING (SELECT :project AS project, :memoryType AS memory_type, :contentHash AS content_hash, :tenantId AS tenant_id FROM DUAL) src
-          ON (trg.project = src.project AND trg.memory_type = src.memory_type AND trg.content_hash = src.content_hash AND trg.tenant_id = src.tenant_id)
-          WHEN MATCHED THEN
-            UPDATE SET
-              embedding      = :embedding,
-              importance     = GREATEST(trg.importance, :importance),
-              content        = :content,
-              session_id     = :sessionId,
-              provider       = :provider,
-              id             = :id,
-              confidence     = GREATEST(trg.confidence, :initialConfidence),
-              evidence_count = trg.evidence_count + 1,
-              scope          = COALESCE(:scope, trg.scope),
-              tags           = COALESCE(TO_CLOB(:tagsJson), trg.tags),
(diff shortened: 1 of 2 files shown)


giauphan (commented):
cc @jules

I've reviewed the PR diff. The `dreamingService.ts` file isn't in this repo (it's from `codeatlas-ai`), so I'm analyzing the diff directly. Here are the flagged issues:

**CRITICAL — Missing `autoCommit: true` (silent data loss)**
`src/services/dreamingService.ts` -> Around line 533 and 544
All `db.execute()` calls dropped the `{ autoCommit: true }` option that the original Oracle code had. With `oracledb`, `autoCommit` defaults to `false` — without it, every MERGE and INSERT transaction is rolled back when the connection returns to the pool. The method appears to succeed but **no data is persisted**. Need to verify whether `createDatabaseAdapter()` enforces auto-commit internally; if not, this is a data-loss bug.

**CRITICAL — Removed connection cleanup (resource leak)**
`src/services/dreamingService.ts` -> Lines 533-540
The `finally` block that called `await connection.close()` to release the pooled connection is completely gone. If `createDatabaseAdapter()` returns an object wrapping a pooled connection that the caller must release, every call leaks a connection — eventually exhausting the pool and hanging the app. If the adapter manages lifecycle internally, this is fine, but it must be verified.

**Medium — `tenantId` moved outside try block**
`src/services/dreamingService.ts` -> Around line 439
`const tenantId = authStorage.getStore()!.uid` was moved before the `try`. If `getStore()` returns null (making `!` throw), the error won't be caught by the `catch` block and won't be logged, reducing debuggability.

**Minor — SQLite embedding type inconsistency in fallback path**
`src/services/dreamingService.ts` -> Around line 536
The MERGE path correctly distinguishes `Uint8Array` for SQLite vs `Float32Array` for Postgres, but the fallback INSERT still uses `new Float32Array(embeddingVector)` unconditionally. SQLite users without the `content_hash` column will pass the wrong embedding type.

**Minor — Code duplication between DB branches**
`src/services/dreamingService.ts` -> Around lines 449-535
The Postgres/SQLite and Oracle MERGE paths share nearly identical bind parameter objects — only the SQL differs. Adding a new parameter requires updating both branches, which is error-prone and should be refactored to share the binds object.

src/services/dreamingService.ts
```diff
@@ -497,10 +533,9 @@ export class OracleDreamingService {
           ? { id, sessionId, project, provider: aiModel ?? null, memoryType, content, embedding: embeddingVector ? new Float32Array(embeddingVector) : null, importance, initialConfidence, tenantId, scope: scope ?? null, tagsJson, relatedIdsJson }
```
giauphan, line 533: `src/services/dreamingService.ts` -> Around line 533 and 544

src/services/dreamingService.ts
```diff
@@ -433,58 +433,94 @@ export class OracleDreamingService {
...
-
-      const tenantId = authStorage.getStore()!.uid;
-      const id = `${project}_${memoryType}_${sessionId}_${Date.now()}`;
-      const initialConfidence = OracleDreamingService.calcInitialConfidence(memoryType, importance);
-      const tagsJson = tags ? JSON.stringify(tags) : null;
-      const relatedIdsJson = relatedIds ? JSON.stringify(relatedIds) : null;
+    const db = createDatabaseAdapter();
+    const tenantId = authStorage.getStore()!.uid;
+    const id = `${project}_${memoryType}_${sessionId}_${Date.now()}`;
+    const initialConfidence = OracleDreamingService.calcInitialConfidence(memoryType, importance);
```
giauphan, line 439: `src/services/dreamingService.ts` -> Around line 439

src/services/dreamingService.ts
```diff
@@ -497,10 +533,9 @@ export class OracleDreamingService {
           ? { id, sessionId, project, provider: aiModel ?? null, memoryType, content, embedding: embeddingVector ? new Float32Array(embeddingVector) : null, importance, initialConfidence, tenantId, scope: scope ?? null, tagsJson, relatedIdsJson }
           : { id, sessionId, project, provider: aiModel ?? null, memoryType, content, embedding: embeddingVector ? new Float32Array(embeddingVector) : null, importance, tenantId, scope: scope ?? null, tagsJson, relatedIdsJson };
 
-        await connection.execute(
+        await db.execute(
```
giauphan, line 536: `src/services/dreamingService.ts` -> Around line 536

PR #193, opened 2026-08-19
⚡ Bolt: O(N) preprocessing for vector similarity loops

giauphan:
💡 **What**:
Refactored the array iteration and map grouping logic in `deduplicateDreams` and `scoreDreams` to extract property lookups and type coercions out of the $O(N^2)$ inner loops. Row objects are now mapped and pre-parsed into objects containing `id`, `importance`/`confidence`, and the `Float32Array` `embedding` during the initial $O(N)$ iteration phase. The `cosineSimilarity` function was updated to accept `Float32Array` directly.

🎯 **Why**:
During the inner $O(N^2)$ iteration phase, dynamically reading properties and repetitively invoking `parseEmbedding` (which attempts to parse JSON or coercions based on type checking) on identical values created massive CPU overhead and garbage collection pressure in V8, significantly slowing down the vector matrix comparisons. 

📊 **Impact**: 
Massively reduces memory allocation and object churn during the inner loops of `consolidationEngine`, resulting in much faster similarity measurements.

🔬 **Measurement**:
The unit test suite passes cleanly, confirming no logic alterations occur from evaluating the structural checks outside the inner loop.

---
*PR created automatically by Jules for task [3290363820906360413](https://jules.google.com/task/3290363820906360413) started by @giauphan*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -21,3 +21,7 @@
 ## 2026-08-19 - Fast vector similarity loops
 **Learning:** In highly mathematical operations inside $O(N^2)$ loops (like pairwise cosine similarity computations), redundant allocations such as extracting IDs/properties and generating arrays dynamically causes massive garbage collection overhead. Furthermore, ensuring input arrays are always passed directly as `Float32Array` enables the V8 runtime to leverage fast, unboxed vector operations natively.
 **Action:** When calculating similarity matrix distances or comparisons, pre-normalize structural types like raw arrays to `Float32Array` before executing nested loops, and cache invariant lookups out of the inner loop scope.
+
+## 2026-08-19 - Pre-parsing vectors to avoid O(N^2) inner-loop allocations
+**Learning:** When calculating vector similarities or comparisons in $O(N^2)$ nested loops, repeatedly extracting properties (`getVal`) and parsing raw string/arrays into `Float32Array` within the inner loop causes massive, redundant CPU cycles and garbage collection pressure in V8.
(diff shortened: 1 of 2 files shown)


giauphan:
cc @jules

Based on my analysis of the diff, here's my review:

---

**Summary:** Clean performance optimization that eliminates redundant `getVal` and `parseEmbedding` calls inside O(N²) inner loops by moving them to an O(N) preprocessing phase. Replaces `any` types with specific typed structures. Behavior is preserved.

**Findings:**

`src/services/consolidationEngine.ts` -> Line ~190: Changed `Map<string, any[]>` to `Map<string, { id: string, importance: number, embedding: Float32Array }[]>()`. This replaces `any` with a concrete type, improving type safety. ✅

`src/services/consolidationEngine.ts` -> Line ~196: Moving `parseEmbedding` and `getVal` calls into the preprocessing loop (O(N)) and storing pre-parsed `Float32Array` results. Previously these were called inside the O(N²) inner loop, causing redundant parsing and GC pressure. This is a significant performance win. ✅

`src/services/consolidationEngine.ts` -> Line ~203: The `if (!embedding) continue` guard skips rows with invalid embeddings during preprocessing. In the original code, such rows were included in the group but produced 0 similarity in `cosineSimilarity`. Net effect is identical — invalid-embedding rows never trigger a merge. Behavior preserved. ✅

`src/services/consolidationEngine.ts` -> Line ~485: Same refactoring pattern applied to the superseding logic. Consistent and correct. ✅

`src/services/consolidationEngine.ts` -> Line ~526: `cosineSimilarity` signature changed from `(a: any, b: any)` to `(vecA: Float32Array, vecB: Float32Array)`, removing internal `parseEmbedding` calls. The defensive `!vecA || !vecB` null checks are now redundant (embeddings are guaranteed valid before reaching this method) but harmless as guards. ✅

**No issues found.** The refactoring is correct, well-typed, and improves performance without changing behavior. The `.jules/bolt.md` documentation entry accurately describes the optimization.

---

PR #194, opened 2026-08-19
⚡ Bolt: [performance improvement] Vector similarity loop caching

giauphan:
### 💡 What
- Introduced a `WeakMap` at the class level of `ConsolidationEngine` to safely cache `Float32Array` embeddings dynamically, keyed by the source object references.
- Updated `cosineSimilarity` to forcefully accept typed `Float32Array` directly instead of ambiguously typed JSON/raw array arrays.
- Implemented inner tight-loop structural optimizations during `cosineSimilarity` (caching lengths and assigning primitive vectors directly to let scopes).

### 🎯 Why
During Dream concept extractions and general duplicate memory checks, the codebase aggregates large arrays of elements and clusters them using nested loops executing `cosineSimilarity()`. The original algorithm recursively performed internal string and JSON extraction on objects within the double inner-loop, redundantly resulting in repetitive GC allocations and high execution pressure. By cleanly caching the pre-parsed values inside a safe `WeakMap` on initialization, performance will dramatically scale. Additionally, localizing tight-loop arrays reduces the property-access overhead for V8 JIT compilations.

### 📊 Impact
Eliminates an explicit $O(N^2)$ memory mapping cycle, significantly optimizing cluster comparison operations and shrinking general background garbage collections scaling linearly with dream generation count without polluting application persistence properties.

### 🔬 Measurement
Run `pnpm test` and assert all current consolidation algorithms correctly execute memory supersession logic natively matching outputs. Ensure `pnpm run build` generates typed files properly.

---
*PR created automatically by Jules for task [8078207392374277566](https://jules.google.com/task/8078207392374277566) started by @giauphan*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -21,3 +21,7 @@
 ## 2026-08-19 - Fast vector similarity loops
 **Learning:** In highly mathematical operations inside $O(N^2)$ loops (like pairwise cosine similarity computations), redundant allocations such as extracting IDs/properties and generating arrays dynamically causes massive garbage collection overhead. Furthermore, ensuring input arrays are always passed directly as `Float32Array` enables the V8 runtime to leverage fast, unboxed vector operations natively.
 **Action:** When calculating similarity matrix distances or comparisons, pre-normalize structural types like raw arrays to `Float32Array` before executing nested loops, and cache invariant lookups out of the inner loop scope.
+
+## 2026-08-19 - Safe Caching of Computed Row Values
+**Learning:** When trying to eliminate $O(N^2)$ repetitive operations by caching derived values (like parsing vectors to Float32Array) inside validation preprocessing loops, directly mutating the input data objects (e.g. `row.EMBEDDING = parsed`) creates severe serialization side effects. The data layer typically expects the raw JSON structure, and converting properties to native Object classes unexpectedly will cause API response corruption downstream.
+**Action:** Always use an independent associative store like `WeakMap<any, Float32Array>` using the row object reference as the key to cache computed values without mutating the original object structure.
--- a/src/services/consolidationEngine.ts
+++ b/src/services/consolidationEngine.ts
@@ -79,8 +79,9 @@ export class ConsolidationEngine {
 
     const rowCache = this.getRowCache(row);
 
-    if (rowCache.has(keyStr)) {
-      return rowCache.get(keyStr) as Float32Array | null;
+    const cached = rowCache.get(keyStr);
+    if (cached !== undefined) {
+      return cached;
     }
 
     const emb = this.parseEmbedding(this.getVal(row, keyIdx, keyStr));
--- a/src/services/consolidationEngine.ts
+++ b/src/services/consolidationEngine.ts
@@ -51,7 +51,7 @@ export interface ConsolidationReport {
 export class ConsolidationEngine {
   // Cache parsed Float32Arrays for row objects to prevent redundant O(n²)
   // JSON/array parsing overhead inside similarity loops
-  private parsedEmbeddingCache = new WeakMap<Record<string, unknown>, Float32Array | null>();
+  private parsedEmbeddingCache = new WeakMap<Record<string, unknown>, Map<string, Float32Array | null>>();
 
   private getVal(row: any, index: number, keyStr: string): any {
     if (!row) return undefined;
@@ -66,12 +66,19 @@ export class ConsolidationEngine {
    * Helper to parse BLOB, Float32Array, number[], or JSON-string embedding into Float32Array.
    */
   private getOrParseEmbedding(row: any, keyIdx: number, keyStr: string): Float32Array | null {
-    if (this.parsedEmbeddingCache.has(row)) {
-      return this.parsedEmbeddingCache.get(row) ?? null;
+    let rowCache = this.parsedEmbeddingCache.get(row);
+
+    if (rowCache !== undefined) {
+      if (rowCache.has(keyStr)) {
+        return rowCache.get(keyStr) as Float32Array | null;
+      }
+    } else {
+      rowCache = new Map<string, Float32Array | null>();
+      this.parsedEmbeddingCache.set(row, rowCache);
     }
 
     const emb = this.parseEmbedding(this.getVal(row, keyIdx, keyStr));
-    this.parsedEmbeddingCache.set(row, emb);
+    rowCache.set(keyStr, emb);
     return emb;
   }
 
@@ -118,7 +125,14 @@ export class ConsolidationEngine {
       logger.warn(`[Consolidation] ${stepName}: Skipping row ${idVal} due to missing or invalid embedding format.`);
       return false;
     }
-    this.parsedEmbeddingCache.set(row, parsed);
+
+    let rowCache = this.parsedEmbeddingCache.get(row);
+    if (!rowCache) {
+      rowCache = new Map<string, Float32Array | null>();
+      this.parsedEmbeddingCache.set(row, rowCache);
+    }
+    rowCache.set('EMBEDDING', parsed);
+
     return true;
   }
 
--- a/src/services/consolidationEngine.ts
+++ b/src/services/consolidationEngine.ts
@@ -51,7 +51,7 @@ export interface ConsolidationReport {
 export class ConsolidationEngine {
   // Cache parsed Float32Arrays for row objects to prevent redundant O(n²)
   // JSON/array parsing overhead inside similarity loops
-  private parsedEmbeddingCache = new WeakMap<any, Float32Array>();
+  private parsedEmbeddingCache = new WeakMap<object, Float32Array>();
 
   private getVal(row: any, index: number, keyStr: string): any {
     if (!row) return undefined;
@@ -214,7 +214,7 @@ export class ConsolidationEngine {
           if (toRemove.has(idI)) continue;
 
           // Embeddings validated above during preprocessing
-          const embI = this.parsedEmbeddingCache.get(rowI) || this.parseEmbedding(this.getVal(rowI, R_IDX.EMBEDDING, 'EMBEDDING'));
+          const embI = this.parsedEmbeddingCache.get(rowI) ?? this.parseEmbedding(this.getVal(rowI, R_IDX.EMBEDDING, 'EMBEDDING'));
           const importanceI = Number(this.getVal(rowI, R_IDX.IMPORTANCE, 'IMPORTANCE'));
 
           if (!embI) continue;
@@ -224,7 +224,7 @@ export class ConsolidationEngine {
             const idJ = String(this.getVal(rowJ, R_IDX.ID, 'ID'));
             if (toRemove.has(idJ)) continue;
 
-            const embJ = this.parsedEmbeddingCache.get(rowJ) || this.parseEmbedding(this.getVal(rowJ, R_IDX.EMBEDDING, 'EMBEDDING'));
+            const embJ = this.parsedEmbeddingCache.get(rowJ) ?? this.parseEmbedding(this.getVal(rowJ, R_IDX.EMBEDDING, 'EMBEDDING'));
             if (!embJ) continue;
 
             const similarity = this.cosineSimilarity(embI, embJ);
@@ -490,7 +490,7 @@ export class ConsolidationEngine {
           const olderId = String(this.getVal(older, SCORE_IDX.ID, 'ID'));
           if (toSupersede.has(olderId)) continue;
 
-          const embO = this.parsedEmbeddingCache.get(older) || this.parseEmbedding(this.getVal(older, SCORE_IDX.EMBEDDING, 'EMBEDDING'));
+          const embO = this.parsedEmbeddingCache.get(older) ?? this.parseEmbedding(this.getVal(older, SCORE_IDX.EMBEDDING, 'EMBEDDING'));
           const olderConfidence = Number(this.getVal(older, SCORE_IDX.CONFIDENCE, 'CONFIDENCE') || 0.5);
 
           if (!embO) continue;
@@ -500,7 +500,7 @@ export class ConsolidationEngine {
             const newerId = String(this.getVal(newer, SCORE_IDX.ID, 'ID'));
             if (toSupersede.has(newerId)) continue;
 
-            const embN = this.parsedEmbeddingCache.get(newer) || this.parseEmbedding(this.getVal(newer, SCORE_IDX.EMBEDDING, 'EMBEDDING'));
+            const embN = this.parsedEmbeddingCache.get(newer) ?? this.parseEmbedding(this.getVal(newer, SCORE_IDX.EMBEDDING, 'EMBEDDING'));
             if (!embN) continue;
 
             const similarity = this.cosineSimilarity(embO, embN);
--- a/src/services/consolidationEngine.ts
+++ b/src/services/consolidationEngine.ts
@@ -51,7 +51,7 @@ export interface ConsolidationReport {
 export class ConsolidationEngine {
   // Cache parsed Float32Arrays for row objects to prevent redundant O(n²)
   // JSON/array parsing overhead inside similarity loops
-  private parsedEmbeddingCache = new WeakMap<object, Float32Array>();
+  private parsedEmbeddingCache = new WeakMap<Record<string, unknown>, Float32Array>();
 
   private getVal(row: any, index: number, keyStr: string): any {
     if (!row) return undefined;
@@ -214,7 +214,12 @@ export class ConsolidationEngine {
           if (toRemove.has(idI)) continue;
 
           // Embeddings validated above during preprocessing
-          const embI = this.parsedEmbeddingCache.get(rowI) ?? this.parseEmbedding(this.getVal(rowI, R_IDX.EMBEDDING, 'EMBEDDING'));
+          let embI = this.parsedEmbeddingCache.get(rowI) || null;
+          if (!embI) {
+            embI = this.parseEmbedding(this.getVal(rowI, R_IDX.EMBEDDING, 'EMBEDDING'));
+            if (embI) this.parsedEmbeddingCache.set(rowI, embI);
+          }
+
           const importanceI = Number(this.getVal(rowI, R_IDX.IMPORTANCE, 'IMPORTANCE'));
 
           if (!embI) continue;
@@ -224,7 +229,12 @@ export class ConsolidationEngine {
             const idJ = String(this.getVal(rowJ, R_IDX.ID, 'ID'));
             if (toRemove.has(idJ)) continue;
 
-            const embJ = this.parsedEmbeddingCache.get(rowJ) ?? this.parseEmbedding(this.getVal(rowJ, R_IDX.EMBEDDING, 'EMBEDDING'));
+            let embJ = this.parsedEmbeddingCache.get(rowJ) || null;
+            if (!embJ) {
+              embJ = this.parseEmbedding(this.getVal(rowJ, R_IDX.EMBEDDING, 'EMBEDDING'));
+              if (embJ) this.parsedEmbeddingCache.set(rowJ, embJ);
+            }
+
             if (!embJ) continue;
 
             const similarity = this.cosineSimilarity(embI, embJ);
@@ -490,7 +500,12 @@ export class ConsolidationEngine {
           const olderId = String(this.getVal(older, SCORE_IDX.ID, 'ID'));
           if (toSupersede.has(olderId)) continue;
 
-          const embO = this.parsedEmbeddingCache.get(older) ?? this.parseEmbedding(this.getVal(older, SCORE_IDX.EMBEDDING, 'EMBEDDING'));
+          let embO = this.parsedEmbeddingCache.get(older) || null;
+          if (!embO) {
+            embO = this.parseEmbedding(this.getVal(older, SCORE_IDX.EMBEDDING, 'EMBEDDING'));
+            if (embO) this.parsedEmbeddingCache.set(older, embO);
+          }
+
           const olderConfidence = Number(this.getVal(older, SCORE_IDX.CONFIDENCE, 'CONFIDENCE') || 0.5);
 
           if (!embO) continue;
@@ -500,7 +515,12 @@ export class ConsolidationEngine {
             const newerId = String(this.getVal(newer, SCORE_IDX.ID, 'ID'));
             if (toSupersede.has(newerId)) continue;
 
-            const embN = this.parsedEmbeddingCache.get(newer) ?? this.parseEmbedding(this.getVal(newer, SCORE_IDX.EMBEDDING, 'EMBEDDING'));
+            let embN = this.parsedEmbeddingCache.get(newer) || null;
+            if (!embN) {
+              embN = this.parseEmbedding(this.getVal(newer, SCORE_IDX.EMBEDDING, 'EMBEDDING'));
+              if (embN) this.parsedEmbeddingCache.set(newer, embN);
+            }
+
             if (!embN) continue;
 
             const similarity = this.cosineSimilarity(embO, embN);
--- a/src/services/consolidationEngine.ts
+++ b/src/services/consolidationEngine.ts
@@ -65,6 +65,15 @@ export class ConsolidationEngine {
   /**
    * Helper to parse BLOB, Float32Array, number[], or JSON-string embedding into Float32Array.
    */
+  private getOrParseEmbedding(row: any, keyIdx: number, keyStr: string): Float32Array | null {
+    let emb = this.parsedEmbeddingCache.get(row) || null;
+    if (!emb) {
+      emb = this.parseEmbedding(this.getVal(row, keyIdx, keyStr));
+      if (emb) this.parsedEmbeddingCache.set(row, emb);
+    }
+    return emb;
+  }
+
   private parseEmbedding(rawEmb: any): Float32Array | null {
     if (!rawEmb) return null;
 
@@ -214,12 +223,7 @@ export class ConsolidationEngine {
           if (toRemove.has(idI)) continue;
 
           // Embeddings validated above during preprocessing
-          let embI = this.parsedEmbeddingCache.get(rowI) || null;
-          if (!embI) {
-            embI = this.parseEmbedding(this.getVal(rowI, R_IDX.EMBEDDING, 'EMBEDDING'));
-            if (embI) this.parsedEmbeddingCache.set(rowI, embI);
-          }
-
+          const embI = this.getOrParseEmbedding(rowI, R_IDX.EMBEDDING, 'EMBEDDING');
           const importanceI = Number(this.getVal(rowI, R_IDX.IMPORTANCE, 'IMPORTANCE'));
 
           if (!embI) continue;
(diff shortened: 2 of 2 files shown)


giauphan:
cc @jules

The `consolidationEngine.ts` file doesn't exist in the current branch — this is new code introduced by the PR. I'm reviewing based on the diff.

Here's my review:

**Good:**
- The WeakMap caching eliminates redundant O(n²) `parseEmbedding` calls in nested loops — the pattern matches the bolt.md learning exactly (cache via row reference, don't mutate input objects).
- `if (!embI) continue;` / `if (!embJ) continue;` guards add robustness when parsing fails on cache miss.
- Changing `cosineSimilarity` to accept typed `Float32Array` params (instead of `any` + internal parse) is cleaner and faster.
- Loop optimization in `cosineSimilarity` (caching `.length`, using locals `vA`/`vB`) avoids repeated property lookups.

**Concerns:**

`src/services/consolidationEngine.ts` -> `WeakMap<any, Float32Array>`:
- Uses `any` for the key type. WeakMap keys must be objects, so `object` is more type-safe and aligns with the project rule "avoid any types".

`src/services/consolidationEngine.ts` -> Cache fallback lines (`|| this.parseEmbedding(...)`):
- `||` works here (Float32Array is always truthy), but `??` is semantically more correct for a "cache miss → compute" fallback. Minor.

`src/services/consolidationEngine.ts` -> `cosineSimilarity` signature change:
- The method no longer calls `parseEmbedding` internally — it now requires pre-parsed `Float32Array`. The diff shows all 2 call sites were updated, but I can't verify from the diff alone that there are no **other** callers of `cosineSimilarity` that would now break by receiving raw embeddings instead of parsed arrays. Contributor should grep the file to confirm.

**Overall:** Solid performance fix. The logic is correct, the caching strategy is sound, and the loop optimizations help. The `any` → `object` change for the WeakMap key and confirming no stray `cosineSimilarity` callers are the only things worth addressing before merge.

---

giauphan:
cc @jules

Now I have full context. The PR diff is from the `codeatlas-ai` project. The files don't exist in this repo — this is a PR being reviewed by this bot. Let me analyze the diff carefully.

Here's my review:

---

**TL;DR:** Solid performance optimization overall. The WeakMap caching eliminates O(n²) redundant `parseEmbedding` calls, and pre-parsing before `cosineSimilarity` is the right refactor. I found one type-safety issue and two items that can't be verified from the diff alone.

---

**`.jules/bolt.md`** — Learning notes documentation. No code issues.

---

**`src/services/consolidationEngine.ts`**

**File: src/services/consolidationEngine.ts -> Line ~52:** `WeakMap<any, Float32Array>` uses `any` as the key type. WeakMap keys must be objects, so `WeakMap<object, Float32Array>` is more precise and type-safe. The "Avoid any types" rule applies. This is the one new `any` introduced by the PR.

**File: src/services/consolidationEngine.ts -> Line ~217 and ~493:** The fallback pattern `this.parsedEmbeddingCache.get(row) || this.parseEmbedding(...)` correctly handles cache misses — `Float32Array` is always truthy (even when empty), so the `||` won't incorrectly fall through to `parseEmbedding` for cached empty arrays. The `if (!embI) continue` guards correctly skip rows where `parseEmbedding` returns `null`/`undefined`. This is correct.

**File: src/services/consolidationEngine.ts -> Line ~217 and ~493 (minor):** When the cache misses and `parseEmbedding` is called in the loop, the result is not written back to the cache. If validation is guaranteed to run before these loops (and thus pre-populate the cache for all valid rows), this is a non-issue and the `||` is just a safety net. If rows can reach the loop without prior validation, the same row could be re-parsed on each encounter. Consider adding `if (embI) this.parsedEmbeddingCache.set(row, embI)` after the fallback parse.

**File: src/services/consolidationEngine.ts -> Line ~527:** The `cosineSimilarity` signature changed from `(a: any, b: any)` (which internally called `parseEmbedding`) to `(vecA: Float32Array, vecB: Float32Array)`. This is a cleaner contract, but **all callers must be updated to pass pre-parsed Float32Arrays**. The diff shows the two loop blocks were updated, but without the full file I cannot verify there are no other callers of `cosineSimilarity` that would now break at runtime (they'd pass raw values that lack `.length` or return `NaN` from indexed access).

**File: src/services/consolidationEngine.ts -> Line ~538:** The loop micro-optimization (pre-extracting `len`, using locals `vA`/`vB`) is a reasonable V8 optimization. The existing null/length checks in `cosineSimilarity` are now somewhat redundant given callers guard with `if (!embX) continue` first, but they're harmless defensive code.

---

**No security issues found.** No injection vectors, hardcoded secrets, or unsafe patterns in this diff.

---

giauphan:
cc @jules

The file doesn't exist in this repository's history — this PR is from an external fork. I'll review based on the diff provided.

## Review

### `.jules/bolt.md`
Documentation only. The learning about avoiding mutation of input objects and using `WeakMap` instead is correct and well-described.

**Overall: good changes.** The PR addresses the exact issue documented in `bolt.md` — caching parsed `Float32Array` values without mutating input row objects.

**What's right:**
- `WeakMap<object, Float32Array>` cache prevents re-parsing embeddings in O(N²) loops — direct `parseEmbedding` calls are replaced with cache lookups with a `??` fallback.
- Input objects are never mutated (the cache key is the object reference; the parsed value is stored separately).
- Null guards (`if (!embI) continue;`) added before each `cosineSimilarity` call — prevents passing `null`/`undefined` to vector math.
- `cosineSimilarity` signature changed from `(a: any, b: any)` to `(vecA: Float32Array, vecB: Float32Array)` — removes the `any` type and eliminates the redundant internal `parseEmbedding` call (parsing now happens once at the call site, either from cache or on-demand).
- Loop body micro-optimized (cached `len`, local `vA`/`vB` variables) for V8 fast-path handling.

**Minor nit:** The `!vecA || !vecB` guard inside `cosineSimilarity` is now redundant since every caller checks for null before invoking it. Not harmful (defensive), but dead logic at this point.

**No issues found** in correctness, security, performance, or maintainability. The changes are sound.

---

giauphan:
cc @jules

I've reviewed the PR diff. Here's my assessment:

## Review: `src/services/consolidationEngine.ts` + `.jules/bolt.md`

**No issues found.** This is a solid, well-executed optimization. Here's what the changes do:

### Performance (the core improvement)
- Adds a `WeakMap<object, Float32Array>` cache (`parsedEmbeddingCache`) that stores already-parsed embeddings during the validation/preprocessing phase.
- In the two O(N²) similarity loops, cached values are looked up directly (`this.parsedEmbeddingCache.get(rowI) ?? this.parseEmbedding(...)`) instead of re-parsing the raw embedding JSON every iteration. This eliminates redundant parsing in the inner loop.
- `WeakMap` keys are GC-eligible — when row objects are dereferenced, cache entries are automatically freed, avoiding memory leaks.
- Loop micro-optimizations in `cosineSimilarity` (caching `len`, local `vA`/`vB` variables) help V8's unboxed access — a reasonable change for a hot inner loop.

### Correctness
- The `??` fallback correctly handles cache misses by calling `parseEmbedding` on demand.
- The added `if (!embI) continue;` guards safely skip rows where parsing returns null/undefined — no null-deref risk.
- `parseEmbedding` results are only cached after validation passes (the `set` call sits after the early `return false`), so the cache never stores invalid values.
- The `.jules/bolt.md` learning note explains *why* this approach was chosen: direct mutation of `row.EMBEDDING = parsed` was an earlier attempt that corrupted downstream serialization. WeakMap avoids this entirely by keeping the original object untouched.

### Type safety
- `cosineSimilarity` signature improved from `any` parameters to `Float32Array` — more type-safe. The `??` + `if (!embI) continue` narrowing ensures only valid `Float32Array` values reach it.

### Security
- No injection vectors, no hardcoded secrets, no unsafe patterns.

The `.jules/bolt.md` addition is appropriate documentation of the learning that drove this change.

---

giauphan:
cc @jules

## PR Review

The PR addresses a known performance and correctness issue in `consolidationEngine.ts`: eliminating redundant O(n²) embedding parsing in cosine similarity loops, and fixing object mutation that caused serialization corruption (documented in `.jules/bolt.md`).

### Analysis

**.jules/bolt.md** — Documentation note only. No issues.

**src/services/consolidationEngine.ts**

src/services/consolidationEngine.ts -> Line 52: `WeakMap<object, Float32Array>` uses `object` as key type. This is acceptable since row objects are all object references, but the more precise type would be `WeakMap<Record<string, unknown>, Float32Array>` to avoid accepting unrelated objects. Minor, non-blocking.

src/services/consolidationEngine.ts -> Line 217 & 496: When the `WeakMap` cache misses (row wasn't pre-validated), the fallback `?? this.parseEmbedding(...)` result is **not stored back** into the cache. In normal flow this shouldn't trigger since `validateAndNormalizeRow` caches every valid row before it enters the comparison loops. However, if a row bypasses validation, it could be re-parsed on every access — and since `embI` (outer loop) and `embJ` (inner loop) reference potentially overlapping rows, a single uncached row could be parsed up to `n` times in the inner loop. Should cache the fallback: `this.parsedEmbeddingCache.set(rowI, parsed)`.

src/services/consolidationEngine.ts -> Line 228 & 503: The `cosineSimilarity` signature changed from `(a: any, b: any)` to `(vecA: Float32Array, vecB: Float32Array)`. This removes the `any` types (good — aligns with the "avoid any types" rule) and eliminates redundant `parseEmbedding` calls inside the method. Both call sites are properly updated with null checks (`if (!embI) continue;`) before calling. Need to verify there are **no other callers** of `cosineSimilarity` outside these two loops that still pass raw `any` values — TypeScript would catch this, but worth confirming.

src/services/consolidationEngine.ts -> Line 542: Pre-existing issue, not introduced by this PR: `denom = Math.sqrt(normA) * Math.sqrt(normB)` can be `0` for zero-vectors, causing `dot / 0` to return `NaN`. The old code had the same behavior. If this matters, add `if (denom === 0) return 0;`.

### Verdict

**Approve with minor notes.** The core change is solid: it fixes a genuine serialization bug (mutating row objects), eliminates O(n²) redundant parsing via `WeakMap` caching, and improves type safety by removing `any`. The two suggested fixes (caching fallback results + guarding zero-vector `denom`) are nice-to-haves; the fallback caching is the more important one since it prevents potential performance regression in edge cases.

---

giauphan:
cc @jules

These symbols don't exist in the current checkout — the PR originates from another project/branch. I'll review based on the diff content.

Let me analyze the changes carefully.

**What this PR does:** Adds a `WeakMap` cache (`parsedEmbeddingCache`) to eliminate redundant `parseEmbedding` calls inside the O(N²) consolidation loops, and refactors `cosineSimilarity` to accept pre-parsed `Float32Array` instead of raw values.

**Analysis:**

1. **Correctness of the caching logic (✅ valid):** Preprocessing populates the cache when an embedding validates. The loops read from cache first, falling back to `parseEmbedding` + cache on miss, then skip with `if (!embI) continue`. The `if (!emb) continue` early-exit (new vs. old) preserves behavior — invalid embeddings can't contribute to consolidation — while avoiding wasted inner-loop iterations. Consistent across both loops.

2. **The `cosineSimilarity` signature change (⚠️ verify):** Changed from `(a: any, b: any)` — which parsed internally — to `(vecA: Float32Array, vecB: Float32Array)`. All visible call sites pass cached/parsed values. **One thing to verify:** confirm there are no *other* callers of `cosineSimilarity` elsewhere in the file that pass raw embedding data — those would now break since the method no longer parses.

3. **`|| null` with `Float32Array (✅ valid):** Weak objects/arrays are truthy, so `|| null` only fires on cache miss (`undefined`). An empty `Float32Array` is truthy and would fall through to `cosineSimilarity`, which correctly returns 0 via its length guard — same as old behavior.

4. **Performance (✅ good):** Eliminates O(N²) redundant parsing (each pair was re-parsing in the old code). The micro-optimization in the inner loop (caching `len`, localizing `vecA[i]`/`vecB[i]`) aligns with the documented `.jules/bolt.md` learnings.

5. **Type safety (✅ good):** `WeakMap<Record<string, unknown>, Float32Array>` is better than the `any` key the `.jules/bolt.md` doc suggests — follows the "avoid any" rule.

6. **No security issues (✅ clean):** No secrets, no injection vectors, no unsafe patterns.

7. **Code duplication (⚠️ minor):** The cache-lookup-and-fallback block is repeated verbatim across 4 call sites (2 per loop). Since `parseEmbedding` is already a function call, extracting a helper like `private getOrParseEmbedding(row, idx): Float32Array | null` would reduce duplication and guarantee consistent cache behavior going forward.

**Summary:** Solid performance optimization with correct logic. The main things to verify:

- **File: `src/services/consolidationEngine.ts` → Line ~484 (after change):** Verify no other callers of `cosineSimilarity` exist outside the diff that pass raw embedding values — the signature is now `Float32Array`-only.
- **File: `src/services/consolidationEngine.ts` → Lines ~217 & ~507 (both loops):** The cache-miss fallback pattern is duplicated 4×; consider extracting a `getOrParseEmbedding` helper for maintainability.

No blocking correctness or security issues.

---

giauphan:
cc @jules

I've reviewed the diff. The PR adds a `WeakMap`-based cache for parsed `Float32Array` embeddings to eliminate redundant O(N²) parsing in the consolidation similarity loops, plus a type fix on `cosineSimilarity`. Here's my assessment:

**What's good:**
- WeakMap cache correctly avoids re-parsing embeddings on every pairwise comparison — this is the right fix for the GC allocation overhead described in `.jules/bolt.md`.
- Changing `cosineSimilarity` from `(a: any, b: any)` to `(vecA: Float32Array, vecB: Float32Array)` is a solid type-safety improvement.
- Local variable caching (`vA`, `vB`, `len`) in the cosine loop is a valid micro-optimization for the hot path.

**Concerns:**

`src/services/consolidationEngine.ts` -> Line ~68 (`getOrParseEmbedding`):
The `|| null` operator should be `?? null`. While a `Float32Array` is always truthy today so the behavior is identical, `??` more precisely expresses "use null only when the key is absent (undefined)" rather than "use null when the value is any falsy value." Minor, but worth aligning with intent.

`src/services/consolidationEngine.ts` -> Line ~68 (`getOrParseEmbedding`):
This method doesn't cache failed parses (null results). If a row's embedding is somehow invalid at the similarity-loop stage (i.e., validation didn't run for it), `getOrParseEmbedding` will re-parse on every call, which partially defeats the O(N²) optimization. The defensive `if (!embI) continue;` guards correctness, but the re-parse is wasted work. Caching the null result as a sentinel would be more robust.

`src/services/consolidationEngine.ts` -> Line ~117 and line ~68:
The validation step writes to `parsedEmbeddingCache.set(row, parsed)` and `getOrParseEmbedding` also writes. This is redundant but harmless since validation runs first and the cache hit short-circuits `getVal`/`parseEmbedding` in the loop. Not a bug, just worth noting the dual write path exists.

**Overall:** The approach is sound and addresses the performance issue described in the learning notes. The WeakMap key on the row object reference is correct (rows are reused across the two similarity loops with different `IDX` constants but distinct object instances). The type improvement on `cosineSimilarity` is welcome. The two concerns above are minor — the first is stylistic, the second is a theoretical edge case that's already guarded against. The PR looks good to merge.

---

giauphan:
cc @jules

I've examined the PR diff. The changes are in `src/services/consolidationEngine.ts` (and a documentation file `.jules/bolt.md`). Let me analyze the code changes.

**Analysis of the diff:**

The PR optimizes the consolidation engine's O(N²) similarity comparison loops by caching parsed `Float32Array` embeddings in a `WeakMap` instead of re-parsing on every comparison. Key changes:

1. **Caching**: `getOrParseEmbedding()` wraps `parseEmbedding` with a `WeakMap` cache keyed by row object reference. Invalid embeddings are cached as `null` and guarded at call sites with `if (!emb) continue`.
2. **Signature change**: `cosineSimilarity` now takes `Float32Array` directly (was `any` + internal parsing), reducing redundant work.
3. **Micro-optimizations**: Loop length cached, indexed values hoisted to locals.
4. **No mutation**: Uses `WeakMap` instead of mutating `row.EMBEDDING` (as noted in `.jules/bolt.md`).

**Review findings:**

All changes look correct and well-implemented. The WeakMap keyed by row reference is safe because preprocessing and comparison loops iterate the same `group` array. Null checks before `cosineSimilarity` calls prevent crashes on the signature change. The performance improvement is significant — eliminating O(N²) redundant parsing.

One note: `getOrParseEmbedding` uses `row: any`, consistent with the existing `getVal`/`parseEmbedding` signatures in the file. The "avoid any types" rule would prefer a stricter type, but it matches the existing codebase convention.

Everything looks good.

---

giauphan:
cc @jules

The repo's pending changes are unrelated to this PR. This is a PR being reviewed by the review-bot itself. I have the full diff to work with. Let me analyze it carefully.

## Code Review

**Overall assessment:** This is a solid performance optimization PR. The core change eliminates redundant O(N²) embedding parsing by caching parsed `Float32Array` values in a `WeakMap`, which aligns with the learning captured in `.jules/bolt.md`. The logic is correct and the type safety is improved.

### What it does well

- **Eliminates O(N²) parsing:** Previously, `cosineSimilarity` called `parseEmbedding` internally, so each pairwise comparison re-parsed the same embeddings N times. Now preprocessing caches parsed values once (O(N)), and the similarity loop retrieves them from cache (O(1) per lookup).
- **Avoids mutating input objects:** Uses `WeakMap<row, Float32Array>` keyed by object reference — exactly as the learning note prescribes. No side effects on the original data structures.
- **Type safety improvement:** `cosineSimilarity` signature changed from `(a: any, b: any)` to `(vecA: Float32Array, vecB: Float32Array)`, removing the internal `parseEmbedding` call and the `any` type.
- **Defensive null checks:** Added `if (!embI) continue;` / `if (!embJ) continue;` guards before similarity computations, preventing null-deref crashes that the old `cosineSimilarity` (which could handle null internally) no longer protects against.
- **Micro-optimizations:** Cached `vecA.length` in `len` and used local `vA`/`vB` variables in the inner loop to avoid repeated indexed property access.

### Issues flagged

**File: src/services/consolidationEngine.ts**

1. **Minor — double lookup in `getOrParseEmbedding` (Line ~68):** The method uses `has()` + `get()` — two WeakMap lookups. A single `get()` with an `undefined` check would suffice:
   ```ts
   const cached = this.parsedEmbeddingCache.get(row);
   if (cached !== undefined) return cached;
   ```
   The `?? null` fallback on the `has` branch is also redundant when `has()` already returned `true`.

2. **Design concern — cache keyed by row only, not (row, keyIdx, keyStr):** The `WeakMap` is keyed solely by the `row` object reference. If different embedding fields ever need to be cached for the same row (different `keyIdx`/`keyStr`), the cache would return a stale value for the wrong field. Low risk today since all callers use `R_IDX.EMBEDDING`/'EMBEDDING', but the cache key doesn't account for the field identity.

3. **Cannot verify other `cosineSimilarity` callers:** The diff only shows two call sites being updated. If any other code in the codebase calls `cosineSimilarity` with raw (un-parsed) values like arrays or JSON strings, it would now receive `0` from the length-mismatch guard instead of actually computing similarity. A repo-wide grep for `cosineSimilarity` is recommended to confirm no other callers exist.

**No security or correctness bugs found.** The null checks at call sites properly guard the narrowed `Float32Array` parameters, and the `WeakMap` correctly allows GC of rows once they're no longer referenced.

---

giauphan:
cc @jules

The files referenced in this PR diff (`src/services/consolidationEngine.ts` and `.jules/bolt.md`) don't exist in the current working directory — this appears to be a PR from a different repository (the codeatlas/vectorization engine based on the context). I'll review the diff as provided.

## Review

### Correctness — looks good

**File: `src/services/consolidationEngine.ts` -> `getOrParseEmbedding` (line 65)**
The caching logic is correct — checks the WeakMap per-row cache, returns cached value if present, otherwise parses + caches. The control flow handles all three cases (cache hit, cache miss, no cache yet) properly.

**File: `src/services/consolidationEngine.ts` -> loops (lines 238-246, 514-522)**
The null guards (`if (!embI) continue;`) are now necessary because `cosineSimilarity` no longer accepts `any`/does internal parsing — it expects `Float32Array` directly. Both call sites have proper null checks before invoking `cosineSimilarity(embI, embJ)`. ✓

**File: `src/services/consolidationEngine.ts` -> `validateRowEmbedding` (line 125)**
Stores the already-validated `parsed` value into the cache during preprocessing. The `return false` for invalid embeddings fires *before* the cache set, so only non-null `Float32Array` values get cached. ✓

### Edge case to consider

**File: `src/services/consolidationEngine.ts` -> `getOrParseEmbedding` (line 68)**
`WeakMap.set(row, ...)` will throw `TypeError` if `row` is a non-object (null/undefined/primitive). The existing `getVal` guards with `if (!row) return undefined`, and `validateRowEmbedding` filters out invalid rows before the loops, so this should be safe in practice — but the method has no defensive guard of its own. If a null row ever reaches the loop, it would crash instead of skipping. Minor robustness concern, not a bug for the expected code path.

### Performance — solid improvements

The core goal is eliminating O(n²) re-parsing of embeddings in nested similarity loops. The `WeakMap` caches each row's parsed `Float32Array` so it's parsed once during validation and reused in both the consolidation and supersede loops. The `cosineSimilarity` micro-optimizations (caching length in `len`, storing `vecA[i]`/`vecB[i]` in locals) are correct and appropriate for hot inner loops. `WeakMap` automatically releases entries when row objects are GC'd — no leak risk.

### Security — clean

No injection vectors, no hardcoded secrets, no unsafe patterns. `WeakMap` is used purely as an in-memory cache.

### Code quality

- **Type safety improved**: `cosineSimilarity(a: any, b: any)` → `cosineSimilarity(vecA: Float32Array, vecB: Float32Array)`. Removing internal `parseEmbedding` calls from the hot path is correct since parsing is now cached.
- **`any` usage**: `getOrParseEmbedding(row: any, ...)` is consistent with existing methods (`getVal(row: any, ...)`, `parseEmbedding(rawEmb: any)`). Could use `Record<string, unknown>` but that's a broader refactor.
- **Documentation** (`.jules/bolt.md`): Accurately describes the WeakMap pattern and why mutating `row.EMBEDDING` directly causes downstream serialization issues.

### Summary

The changes are correct and effectively address the O(n²) parsing overhead documented in the `.jules/bolt.md` entry. The caching strategy, null guards, and `cosineSimilarity` signature change are all consistent. The only minor concern is the lack of a null-row guard in `getOrParseEmbedding`, but this is protected by upstream validation. The code looks good.

---

giauphan:
cc @jules

Now let me check for tests and look at the PR branch in the actual repo.

---

giauphan:
cc @jules

The PR diff I'm reviewing is from the `codeatlas-ai` repo (referenced in the context as `giauphan/codeatlas-ai`). Let me analyze the changes thoroughly.

## Review of PR Changes

### `.jules/bolt.md`
Documentation of a learning note. No code concerns.

**Summary:** This PR eliminates redundant `parseEmbedding` calls inside O(N²) similarity loops by introducing a `WeakMap`-based cache, and simplifies `cosineSimilarity` to accept already-parsed `Float32Array` values.

**What's good:**
- Using `WeakMap` (per the documented learning) avoids mutating input row objects while caching — prevents the serialization side effects noted in bolt.md.
- Null checks (`if (!embI) continue;`, `if (!embJ) continue;`) are correctly placed before `cosineSimilarity` calls in both loops, which is required since `cosineSimilarity` no longer handles null/any internally.
- `validateRow` pre-caches the parsed embedding during validation, so the similarity loops get a cache hit instead of re-parsing — this is the core optimization.
- The `cosineSimilarity` loop optimization (caching `len`, using local `vA`/`vB`) is valid for performance.

**Issues found:**

**File: `src/services/consolidationEngine.ts` -> Line ~110-120 (`getOrParseEmbedding`):**
The cache initialization logic is duplicated between `getOrParseEmbedding` and `validateRow` (both do `this.parsedEmbeddingCache.get(row)` + `new Map()` + `set`). Minor, but extractable into a private helper like `getRowCache(row)` for maintainability.

**File: `src/services/consolidationEngine.ts` -> Line ~243-245 (first similarity loop):**
`importanceI` is computed via `getVal` *before* the `if (!embI) continue;` check. This is wasted work when `embI` is null — a row that fails embedding validation will still compute `importanceI` unnecessarily. Minor inefficiency, not a correctness bug.

**File: `src/services/consolidationEngine.ts` -> Line ~489-491 (second similarity loop):**
Same pattern — `olderConfidence` computed before `if (!embO) continue;`. Same minor inefficiency.

**Potential correctness concern (verify only):**
The `cosineSimilarity` signature changed from `(a: any, b: any)` to `(vecA: Float32Array, vecB: Float32Array)`. The diff shows only two call sites, both updated. Since it's a `private` method, if there are other callers not visible in this diff, they would break. Recommend confirming no other callers exist.

**Overall:** Solid optimization. The approach is correct and addresses the documented learning. The issues above are minor (code duplication + minor inefficiencies), not blockers.

---

giauphan:
cc @jules

I now have a complete picture of the PR. Let me review it against the specified criteria.

## Review

This PR optimizes the `ConsolidationEngine` by caching parsed `Float32Array` embeddings in a `WeakMap`, eliminating redundant O(N²) `parseEmbedding` calls inside pairwise similarity loops.

### Correctness — no issues found

- The `WeakMap` cache uses row object references as keys, which is sound.
- The cache key `'EMBEDDING'` is consistent: preprocessing sets it (`this.getRowCache(row).set('EMBEDDING', parsed)`), and all four similarity-loop call sites retrieve it via `getOrParseEmbedding(..., 'EMBEDDING')`. ✓
- All four `getOrParseEmbedding` results (`embI`, `embJ`, `embO`, `embN`) are null-checked with `if (!embX) continue;` before being passed to `cosineSimilarity`. ✓
- The `cosineSimilarity` signature change from `(a: any, b: any)` to `(vecA: Float32Array, vecB: Float32Array)` is consistent with all callers now passing pre-parsed arrays. ✓
- Cached `null` values (for rows with no valid embedding) are handled correctly: `Map.has` + `Map.get` distinguishes "key absent" from "value is null", so `null` results are properly returned and trigger `continue`. ✓

### Security — no issues found

- No hardcoded secrets, no injection vectors, no unsafe deserialization.
- `WeakMap` uses object references as keys — no prototype pollution or `__proto__` manipulation.

### Performance — strong improvement

- Eliminates O(N²) redundant `parseEmbedding` calls. Previously, each pairwise comparison in `cosineSimilarity` re-parsed both embeddings; now each is parsed once and cached.
- Local variable caching in the cosine similarity loop (`const len = vecA.length; const vA = vecA[i]; const vB = vecB[i];`) is a minor but valid improvement.

### Code quality — clean

- Follows the `.jules/bolt.md` learning: uses `WeakMap` instead of mutating `row.EMBEDDING`, avoiding serialization side effects. ✓
- Good separation of concerns: caching in `getRowCache`/`getOrParseEmbedding`, computation in `cosineSimilarity`.
- Type safety improved with `cosineSimilarity` signature narrowed from `any` to `Float32Array`.

### One minor suggestion

`src/services/consolidationEngine.ts` -> `getOrParseEmbedding` (around line 103)
`rowCache.has(keyStr)` followed by `rowCache.get(keyStr)` does two Map lookups. Since `Map.get` returns `undefined` when the key is absent (and `null` is a valid cached value), you could use a single lookup:
```ts
const cached = rowCache.get(keyStr);
if (cached !== undefined) return cached;
```
This is negligible compared to the O(N²) optimization, so not blocking.

---

No blocking issues. The PR is correct, secure, performant, and well-implemented. The `WeakMap`-based caching correctly follows the documented lesson in `.jules/bolt.md` to avoid object mutation, and all four similarity-loop call sites properly null-check before use.

---

giauphan:
Superseded by #193 (already merged) which implemented O(N) vector preprocessing on the same consolidationEngine code path. This PR now conflicts with main and is redundant.

PR #196, opened 2026-08-20
⚡ Bolt: [performance improvement] Pre-normalize embeddings for faster similarity calculations

giauphan:
💡 What: Pre-normalizes embedding vectors to unit length inside `parseEmbedding` and simplifies `cosineSimilarity` to a single dot product calculation. Deep copies buffers during parsing to prevent accidental memory mutations.
🎯 Why: In `deduplicateDreams` and `scoreDreams`, `cosineSimilarity` is calculated within $O(N^2)$ nested loops. The original implementation calculated vector lengths inside this inner loop. By normalizing vectors to unit length during the $O(N)$ preprocessing phase (`parseEmbedding`), we completely eliminate the need to calculate vector lengths inside the $O(N^2)$ loops.
📊 Impact: Considerably reduces math operations inside $O(N^2)$ loops (reducing square root calculations and replacing 3 additions/multiplications with 1), leading to significantly faster large-scale project knowledge consolidation.
🔬 Measurement: Verify with `pnpm test` and observing performance profiling on large scale similarity matrix calculations in `consolidationEngine`.

---
*PR created automatically by Jules for task [8965630495311598517](https://jules.google.com/task/8965630495311598517) started by @giauphan*


giauphan:
cc @jules

I now have enough context to review the diff. The files referenced (`src/services/consolidationEngine.ts`, `.jules/bolt.md`) don't exist in this repo — they're from the downstream project being reviewed by this bot. I'll analyze the diff directly.

## Code Review

`.jules/bolt.md` — documentation, no issues.

`src/services/consolidationEngine.ts`:

**Overall assessment**: The changes are correct and well-reasoned. The two key fixes address real bugs:

1. **Shared buffer mutation** — The original `Float32Array` input returned the same reference, and the `Uint8Array`/`Buffer` path created a *view* over the underlying `ArrayBuffer` via `new Float32Array(rawEmb.buffer, ...)`. In-place normalization would then corrupt the original cached/shared buffer. The fix correctly copies both paths before mutation.

2. **Normalization + cosine similarity simplification** — Pre-normalizing embeddings in `parseEmbedding` and simplifying the similarity math to a bare dot product is correct: for unit vectors, `dot(A,B) / (|A|*|B|) = dot(A,B)`. Zero-vector guard (`norm > 0`) is handled properly.

**Concerns to verify before merge**:

- **Behavioral side-effect in `parseEmbedding`**: Normalization is now applied unconditionally to *every* parsed embedding. If any caller uses `parseEmbedding` for non-similarity purposes (e.g., storing embeddings for later retrieval, magnitude-based ranking, dimensionality reduction), the normalized output will be wrong. **Check all call sites** — every consumer must only need unit-normalized vectors.

- **Similarity callers must go through `parseEmbedding`**: The new `calculateSimilarity` (or equivalent) assumes pre-normalized inputs. Verify there are no code paths that pass raw (unnormalized) embeddings directly to this method, as the norm computation was removed entirely and would silently produce incorrect results.

- **Minor**: The empty `catch { // null }` block silently swallows parse errors (functionally equivalent to the old `return null`, but relies on the later `if (result)` guard). Not a bug, slightly less explicit.

No TypeScript best-practice violations, no `any` introduced (pre-existing), no security issues, no performance regressions (the O(N) normalization replaces an equivalent O(N) norm computation during similarity, and is done once per parse instead of per comparison).

---

giauphan:
cc @jules

The code being reviewed is from another project. I'll review the diff directly based on my analysis.

---

## Code Review: Fix in-place ArrayBuffer mutation bugs in `consolidationEngine.ts`

**Summary:** This PR fixes a real bug — in-place mutation of shared/cached `ArrayBuffer` backing data when normalizing embeddings. The fix deep-copies buffers before wrapping them in `Float32Array` views, pre-normalizes vectors in `parseEmbedding`, and simplifies `cosineSimilarity` to return just the dot product (correct for unit vectors).

### Issues found:

**File: `src/services/consolidationEngine.ts` -> Line ~44 (`parseEmbedding`)**

1. **Correctness risk — `cosineSimilarity` assumes pre-normalized inputs.** The `cosineSimilarity` method now returns `dot` directly, relying entirely on `parseEmbedding` having normalized every vector via its new normalization block. If any code path calls `cosineSimilarity` with a raw/non-normalized `Float32Array`, it returns the dot product instead of cosine similarity — silently wrong results with no guard. The comment acknowledges the assumption, but there's no runtime validation. Consider adding an assertion or documenting this as a strict precondition.

2. **Pre-existing `any` type:** `parseEmbedding(rawEmb: any)` should use a discriminated type (`Float32Array | number[] | Uint8Array | Buffer | string`) to enforce type safety at the call site.

3. **Minor — unnecessary `// null` comment** in the `catch` block adds noise without value.

### What's correct:

- **Buffer copy fix:** `rawEmb.buffer.slice(byteOffset, byteOffset + byteLength)` creates a proper deep copy of the raw bytes before wrapping in `Float32Array(copy)`. This eliminates the view-sharing mutation bug described in the journal entry. ✓
- **Float32Array copy:** `new Float32Array(rawEmb)` when `rawEmb` is already a `Float32Array` creates an independent copy (per spec, TypedArray-to-TypedArray constructor copies elements), so the in-place normalization is safe. ✓
- **Normalization guard:** `if (norm > 0)` prevents division-by-zero on all-zero vectors. ✓
- **Consistency with journal entry:** The code changes correctly implement the documented learning — `.slice()` before typed-array wrapping for any in-place math. ✓

### Verdict:

The core fix is sound and correctly addresses the documented mutation bug. The main concern is issue #1 — the `cosineSimilarity` simplification introduces a hidden precondition that could cause subtle correctness bugs if the calling convention changes. Everything else is minor.

---

giauphan:
cc @jules

I now have enough context to review this PR. Let me analyze the changes.

The PR fixes a real bug — in-place `ArrayBuffer` mutation when creating `Float32Array` views over shared buffers during normalization. The `.slice()` deep-copy is correct. But there are several issues introduced:

**Issues found:**

1. **`cosineSimilarity` silently breaks for non-normalized inputs** (`src/services/consolidationEngine.ts` -> ~Line 540): Removing the normalization computation (`normA`, `normB`, `denom`) means the function now returns a raw dot product. This is only correct if every caller first passes vectors through `parseEmbedding`. The original was self-contained and correct for any input. A non-normalized vector now silently yields an incorrect similarity score — a logic bug waiting to happen.

2. **Normalization baked into `parseEmbedding`** (`src/services/consolidationEngine.ts` -> Line 63): This changes the method's contract from "parse" to "parse *and normalize*". Any caller that needs raw (magnitude-bearing) embeddings will now silently receive normalized ones. Normalization should be a separate method or opt-in, not a side-effect of parsing.

3. **`any` type on `rawEmb`** (`src/services/consolidationEngine.ts` -> Line 65): `parseEmbedding(rawEmb: any)` violates TypeScript best practices and the project's own conventions (see `src/engine/types.ts` which uses explicit interfaces). Should be a union type like `Float32Array | number[] | Uint8Array | Buffer | string | null`.

4. **Empty catch block** (`src/services/consolidationEngine.ts` -> ~Line 100): `} catch { // null }` silently swallows JSON parse errors. The `// null` comment is misleading — it looks like a leftover from the previous `return null;`. Swallowed parse failures should at least be logged.

The buffer copy fix itself is correct and addresses the documented mutation bug, but the normalization/cosine-similarity coupling is fragile and the `any` type / empty catch are quality issues.

---

giauphan:
cc @jules

I've analyzed the PR diff. Here's my review.

## Summary

This PR fixes an in-place `ArrayBuffer` mutation bug in `parseEmbedding` by deep-copying buffer inputs before normalization, and simplifies `computeCosineSimilarity` to rely on that pre-normalization.

## Issues Found

**File: `src/services/consolidationEngine.ts` -> `computeCosineSimilarity` (around line 540)**

The method no longer computes vector norms and instead returns the raw dot product, assuming all inputs are pre-normalized by `parseEmbedding`. This creates a hidden contract: if any code path creates a non-normalized `Float32Array` and passes it here, the result will be silently incorrect (dot product instead of cosine similarity). The old code was self-contained and correct for any inputs. This is a correctness/maintainability risk — the assumption should be enforced (e.g., assert or type contract) rather than relying on a comment.

**File: `src/services/consolidationEngine.ts` -> `parseEmbedding` (around line 73)**

- `rawEmb instanceof Buffer` is redundant — `Buffer` extends `Uint8Array`, so `rawEmb instanceof Uint8Array` already covers it. Minor dead code.
- `rawEmb.buffer.slice(...)` would throw if `rawEmb.buffer` is a `SharedArrayBuffer` (which lacks `.slice()`). Unlikely given this parses DB blobs, but worth guarding if external/untrusted buffers can reach here.

**File: `src/services/consolidationEngine.ts` -> string parse catch block (around line 86)**

The `// null` comment in the empty catch block is unnecessary noise — a bare `catch {}` is clear enough.

## What's Good

- The core fix is correct: `new Float32Array(rawEmb)` for Float32Array inputs and `buffer.slice()` for Uint8Buffer/Buffer inputs both produce independent copies, so in-place normalization is now safe.
- The zero-norm guard (`norm > 0`) in the normalization loop prevents division-by-zero for zero vectors.
- Moving normalization to parse-time is a net performance win — `computeCosineSimilarity` skips two `O(N)` norm computations per pair.

---

giauphan:
cc @jules

Now I have the full picture. The PR targets `codeatlas-ai` (not this repo), modifying `src/services/consolidationEngine.ts` with a series of temporary patch scripts. Let me review the diff thoroughly.

## Review

Here are the issues I found:

**1. Temporary patch scripts committed — CRITICAL (code quality)**

`patch.cjs`, `patch2.cjs`, `patch3.cjs`, `patch4.cjs`, `patch5.cjs`, `patch6.cjs` are throwaway Node.js scripts that use `fs.readFileSync`/`fs.writeFileSync` with fragile string replacement (`.replace()`) to mutate source code. These are development artifacts and must not be committed. Remove all six files.

**2. `cosineSimilarity` precondition is fragile — maintainability**

The function was simplified to return only the dot product, relying on callers pre-normalizing via `normalizeEmbedding`. The JSDoc documents this, but there's no runtime enforcement. If a new caller is added and forgets to normalize, it silently returns a wrong result. Consider a runtime assertion (`if (Math.abs(normA - 1) > 0.01) throw ...`) in dev, or have `cosineSimilarity` normalize internally.

**3. Empty `catch {}` block — minor (maintainability)**

In `parseEmbedding`'s string-parsing branch, the catch block is now empty (the `// null` comment was removed). While `result` stays `null` by default, an empty catch can hide unexpected errors. Add a comment explaining why swallowing is acceptable here.

**4. Redundant `Buffer` in type union — minor (code quality)**

The `parseEmbedding` signature includes `Buffer` in the union, but the implementation only checks `instanceof Uint8Array` (since `Buffer extends Uint8Array`, this is correct). The `Buffer` type in the signature is misleading — either remove it from the union or keep the explicit `instanceof Buffer` check for clarity.

**5. `.jules/bolt.md` learning log — minor (code quality)**

This is a learning/documentation artifact. It's fine to include, but worth confirming it's the intended location for such notes.

**What's correct:**

- `normalizeEmbedding` correctly computes L2 normalization and handles zero vectors.
- `parseEmbedding` now deep-copies Float32Array and Uint8Array inputs before returning, preventing shared-buffer mutation bugs.
- The `SharedArrayBuffer` guard in the Uint8Array branch is correct — `rawEmb.buffer instanceof ArrayBuffer` is false for SharedArrayBuffer, falling through to the `.slice()`-based fallback.
- Both `deduplicateDreams` and `scoreDreams` call `normalizeEmbedding` after `parseEmbedding` and before similarity computation.
- The `removeEmpty` comment cleanup and type annotation improvements are good.

**Summary:** The core algorithmic changes are sound. Remove the six `patch*.cjs` files before merging, and address the `cosineSimilarity` precondition fragility.

---

giauphan:
cc @jules

I'm reviewing this PR's diff directly since `consolidationEngine.ts` doesn't exist on the current branch — it's the change under review. Here's my analysis.

## Review

Overall the changes are solid and correctly fix the documented in-place `ArrayBuffer` mutation bug. The `.jules/bolt.md` note accurately matches what the code now does. A few observations:

**Good:**
- `parseEmbedding` now copies the underlying buffer (via `.slice()`) for `Uint8Array`/`Buffer` inputs before wrapping in `Float32Array`, so in-place normalization can't corrupt Node's internal memory pool. The `SharedArrayBuffer` fallback also correctly escapes to a private `ArrayBuffer`.
- `Buffer` is still handled correctly since `Buffer extends Uint8Array`, so `instanceof Uint8Array` catches it (no regression despite dropping the explicit `Buffer` check).
- `Float32Array` inputs are now copied too, preventing mutation of cached/reused references — necessary since `normalizeEmbedding` mutates in place.
- Both call sites (lines ~220 and ~494) now call `normalizeEmbedding` after parsing, and `cosineSimilarity` was simplified to a dot product only, which is correct under the new precondition.
- Removed the `any` type on `rawEmb` in favor of a precise union.

**Flag — File: `src/services/consolidationEngine.ts` -> Line ~548 (`cosineSimilarity`):**
The non-production invariant check only validates `vecA` (`normA`), but the precondition comment states *both* `vecA` and `vecB` must be pre-normalized. The asymmetry means an unnormalized `vecB` would silently pass the invariant check in dev and produce a wrong (un-caught) result. Minor, but inconsistent with the documented contract.

**Minor edge case:** A zero-vector (all `0`s) survives `normalizeEmbedding` (the `norm > 0` guard skips it) but then trips the invariant check (`|0 - 1| > 0.01` → throws in dev). The original code returned `0` for this case. Unlikely with real embeddings, but worth being aware of if malformed data can reach this path.

**Nit:** The bare `catch {}` swallows all errors, but the comment mentions only `SyntaxError`. In practice `JSON.parse` only throws `SyntaxError`, so this is harmless — just a slightly inaccurate comment.

No blocking issues. The mutation fix is correct and the perf optimization (pre-normalization + dot-only similarity) is sound.

---

giauphan:
cc @jules

Now I have the full context. The file exists in external temp directories from the `codeatlas-ai` repo. Let me find the version matching the PR base (one that already has `parseEmbedding`).

---

giauphan:
Superseded by merged PR #204, which contains the same cosineSimilarity optimization with green CI. Closing this duplicate to keep a single canonical implementation.

PR #202, opened 2026-08-22
🎨 Palette: Add proper ARIA tab roles and focus rings to DocumentationView

giauphan:
**💡 What:**
Added standard ARIA tab roles (`role="tablist"`, `role="tab"`, `role="tabpanel"`) and linking properties (`aria-controls`, `aria-labelledby`) to the navigation tabs in the `DocumentationView`. Also applied the `FOCUS_RING_CLASS` to the tab buttons, copy buttons, and external links.

**🎯 Why:**
The documentation tabs were built with `button` elements that lacked structural context for screen readers to understand they were navigating sub-panels. Additionally, several interactive elements (like the copy buttons and external links) lacked the standard visible focus ring, making keyboard navigation difficult and inaccessible.

**📸 Before/After:**
*(See attached automated verification screenshot)*

**♿ Accessibility:**
- Proper tab hierarchy makes screen reader navigation intuitive.
- Visible focus rings ensure keyboard-only users can clearly see which element they are interacting with.

---
*PR created automatically by Jules for task [9065784522562670857](https://jules.google.com/task/9065784522562670857) started by @giauphan*

### DIFF
--- a/dashboard/src/components/DocumentationView.tsx
+++ b/dashboard/src/components/DocumentationView.tsx
@@ -141,9 +141,30 @@ export const DocumentationView: React.FC = () => {
           { id: 'mcp', label: 'MCP Editor Config', icon: Settings },
           { id: 'architecture', label: 'Architecture', icon: Layers },
           { id: 'graph', label: 'Interactive Guide', icon: HelpCircle },
-        ].map((tab) => {
+        ].map((tab, index, array) => {
           const Icon = tab.icon;
           const isActive = activeSubTab === tab.id;
+
+          const handleKeyDown = (e: React.KeyboardEvent) => {
+            let nextIndex = -1;
+            if (e.key === 'ArrowRight') {
+              nextIndex = (index + 1) % array.length;
+            } else if (e.key === 'ArrowLeft') {
+              nextIndex = (index - 1 + array.length) % array.length;
+            } else if (e.key === 'Home') {
+              nextIndex = 0;
+            } else if (e.key === 'End') {
+              nextIndex = array.length - 1;
+            }
+
+            if (nextIndex !== -1) {
+              e.preventDefault();
+              setActiveSubTab(array[nextIndex].id as any);
+              const nextTab = document.getElementById(`tab-${array[nextIndex].id}`);
+              if (nextTab) nextTab.focus();
+            }
+          };
+
           return (
             <button
               key={tab.id}
@@ -151,7 +172,9 @@ export const DocumentationView: React.FC = () => {
               aria-selected={isActive}
               aria-controls={`panel-${tab.id}`}
               id={`tab-${tab.id}`}
+              tabIndex={isActive ? 0 : -1}
               onClick={() => setActiveSubTab(tab.id as any)}
+              onKeyDown={handleKeyDown}
               className={FOCUS_RING_CLASS}
               style={{
                 background: 'transparent',
--- a/dashboard/src/components/DocumentationView.tsx
+++ b/dashboard/src/components/DocumentationView.tsx
@@ -159,7 +159,7 @@ export const DocumentationView: React.FC = () => {
 
             if (nextIndex !== -1) {
               e.preventDefault();
-              setActiveSubTab(array[nextIndex].id as any);
+              setActiveSubTab(array[nextIndex].id as 'mcp' | 'architecture' | 'graph');
               const nextTab = document.getElementById(`tab-${array[nextIndex].id}`);
               if (nextTab) nextTab.focus();
             }
@@ -173,7 +173,7 @@ export const DocumentationView: React.FC = () => {
               aria-controls={`panel-${tab.id}`}
               id={`tab-${tab.id}`}
               tabIndex={isActive ? 0 : -1}
-              onClick={() => setActiveSubTab(tab.id as any)}
+              onClick={() => setActiveSubTab(tab.id as 'mcp' | 'architecture' | 'graph')}
               onKeyDown={handleKeyDown}
               className={FOCUS_RING_CLASS}
               style={{
--- a/dashboard/src/components/DocumentationView.tsx
+++ b/dashboard/src/components/DocumentationView.tsx
@@ -14,8 +14,10 @@ import {
 } from 'lucide-react';
 import { FOCUS_RING_CLASS } from '../lib/constants';
 
+type TabType = 'mcp' | 'architecture' | 'graph';
+
 export const DocumentationView: React.FC = () => {
-  const [activeSubTab, setActiveSubTab] = useState<'mcp' | 'architecture' | 'graph'>('mcp');
+  const [activeSubTab, setActiveSubTab] = useState<TabType>('mcp');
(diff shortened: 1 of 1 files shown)


giauphan:
cc @jules

I need to review this PR diff for the `DocumentationView.tsx` accessibility changes. The files aren't local to this repo, so I'll analyze the diff directly.

I've examined the diff thoroughly. Here's my code review:

---

## Code Review: `dashboard/src/components/DocumentationView.tsx`

### Good
- ARIA structure is correct: `role="tablist"` on container, `role="tab"` on buttons, `role="tabpanel"` on panels
- `aria-controls`/`id` cross-references are consistent (`panel-{id}` ↔ `tab-{id}`) ✓
- Focus ring classes preserved alongside existing classes ✓

### Issues

**File: `dashboard/src/components/DocumentationView.tsx` -> Line (tabs map callback): `isActive` is referenced but not defined.**
`aria-selected={isActive}` uses a variable that doesn't appear in the diff. If it wasn't already defined in the `.map()` callback scope before the `return`, it will evaluate to `undefined`, causing React to omit `aria-selected` entirely — screen readers won't know which tab is active. Fix: either add `const isActive = activeSubTab === tab.id;` before the `return`, or use `aria-selected={activeSubTab === tab.id}` inline.

**File: `dashboard/src/components/DocumentationView.tsx` -> Line (tab buttons): Missing `tabIndex` management.**
ARIA `role="tab"` requires roving tabindex: only the active tab should have `tabIndex={0}`, all others `tabIndex={-1}`. Without this, every tab remains in the natural tab order, which breaks standard tablist keyboard navigation (users must Tab through all tabs instead of using arrow keys).

**File: `dashboard/src/components/DocumentationView.tsx` -> Line (tabs map callback): Missing keyboard navigation.**
Adding `role="tab"` implies keyboard support per WAI-ARIA Authoring Practices. Currently only `onClick` is wired — no `onKeyDown` handler for arrow key navigation (Left/Right or Up/Down to switch tabs), Home (first tab), or End (last tab). This is a partial ARIA implementation: the semantics are declared but the required keyboard interaction is missing, which can actively harm keyboard users. Add an `onKeyDown` handler that calls `setActiveSubTab` and manages focus.

### Summary
The ARIA roles/attributes are structurally correct, but the implementation is incomplete. Fix the `isActive` variable, add roving tabindex, and add keyboard navigation to make this a proper, fully-functional tablist.

---

giauphan:
cc @jules

Now let me review the diff carefully. The changes are accessibility improvements to a tab interface in `DocumentationView.tsx`.

## Review

Most changes are solid, but I found one issue that needs verification.

### Issues found

**File: `dashboard/src/components/DocumentationView.tsx` -> Line 137 (import)**
`FOCUS_RING_CLASS` is imported from `../lib/constants` and used on the copy button and two docs buttons. I cannot verify this constant exists in the referenced `../lib/constants` file — it's not findable in the repo. If the constant doesn't exist, the build will fail. The author should either define it in `../lib/constants` or use an inline className string.

**File: `dashboard/src/components/DocumentationView.tsx` -> Line 142 (`handleKeyDown`)**
The `as any` cast on `array[nextIndex].id` suppresses the type check that the tab IDs match the `useState<'mcp' | 'architecture' | 'graph'>` union. The pre-existing `tab.id as any` on the `onClick` handler has the same issue, but the new keyboard handler should ideally derive the next ID through a typed variable rather than `as any`. Minor, but worth a comment since it weakens type safety.

### What's correct

- **Keyboard navigation** (ArrowLeft/Right, Home, End) with `preventDefault()` and `focus()` — follows the WAI-ARIA roving tabindex pattern.
- **ARIA attributes** — `role="tablist"`, `role="tab"`, `aria-selected`, `aria-controls`, `aria-labelledby`, `id`, `tabIndex={isActive ? 0 : -1}` are all correctly applied and reference each other consistently (`tab-{id}` ↔ `panel-{id}`).
- **Tab panels** use `role="tabpanel"` with proper `aria-labelledby` back-references.
- `FOCUS_RING_CLASS` adds visible focus styling for keyboard users — good for accessibility.

### Summary

The accessibility implementation is correct and follows WAI-ARIA authoring practices. The only real concern is whether `FOCUS_RING_CLASS` is actually exported from `../lib/constants` — that needs to be verified before merge.

---

giauphan:
cc @jules

Checking the PR diff now. It's an accessibility improvement adding proper ARIA tab roles and keyboard navigation to the `DocumentationView` tab component.

**Everything looks good.** This is a clean, well-implemented accessibility enhancement following WAI-ARIA tablist practices. No bugs, security issues, or performance concerns found.

A couple of notes:

- **Duplicate type union**: `'mcp' | 'architecture' | 'graph'` is cast twice (in `onClick` and in `handleKeyDown`). Minor — could extract to a type alias, but not worth blocking on.
- **`tabIndex` + `.focus()` interplay is correct**: non-active tabs have `tabIndex={-1}`, which still allows programmatic `.focus()` (only blocks Tab-key navigation). The roving-tabindex pattern is implemented properly.
- **ARIA IDs line up**: `tab-{id}` matches `aria-labelledby="tab-{id}"` on panels, and `panel-{id}` matches `aria-controls="panel-{id}"` on tabs. ✓
- **Keyboard nav logic is correct**: ArrowRight wraps with `% length`, ArrowLeft uses `(index - 1 + length) % length`, Home/End jump to first/last, `preventDefault()` is called before focus. ✓

The `handleKeyDown` closure is recreated on each render, consistent with the inline-style pattern already used throughout this component. No real issues to flag.

---

giauphan:
cc @jules

I've reviewed the PR. Here's my assessment:

## Review

The PR adds proper WAI-ARIA tabs pattern compliance, keyboard navigation, and focus rings to the `DocumentationView` tab component. All changes are correct:

- **`FOCUS_RING_CLASS`** import resolves correctly — it already exists in `../lib/constants` and is used across 12+ files in the codebase (Auth, Dashboard, ControlCenterView, etc.), so this is a consistent, established pattern.
- **`TabType` type alias** is a good improvement — replaces the inline union type and the `as any` cast on line 149 with `as TabType`, tightening type safety.
- **ARIA attributes are consistent**: `aria-controls="panel-${tab.id}"` on each tab matches `id="panel-${tab.id}"` on each `motion.div` panel; `id="tab-${tab.id}"` on tabs matches `aria-labelledby="tab-${tab.id}"` on panels. ✓
- **Keyboard navigation** (`ArrowRight`/`ArrowLeft`/`Home`/`End` with wrap-around via modulo) is correct and follows the WAI-ARIA Authoring Practices tab pattern. ✓
- **`tabIndex` management** (`0` for active, `-1` for inactive) is correct. ✓
- **`framer-motion`'s `motion.div`** passes through `role`, `aria-*`, and `id` props to the DOM element — no issue. ✓

No logic bugs, security issues, performance concerns, or code quality problems. The `handleKeyDown` is defined inside the `.map` render — a negligible re-creation cost for 3 tabs.

**Approve.** This is a clean, well-implemented accessibility improvement with no blocking issues.

---

PR #263, opened 2026-09-02
🎨 Palette: [UX improvement] Add ARIA roles and selected state to auth tabs

giauphan:
💡 What: Added `role="tablist"` to the auth tab container, and `role="tab"` with `aria-selected` to each tab button in `Auth.tsx`.
🎯 Why: Makes the tab switcher correctly announce itself to screen readers, improving accessibility and fulfilling standard tab interaction expectations.
📸 Before/After: Visually identical, structurally accessible.
♿ Accessibility: Ensures assistive technologies understand the tabbed interface structure and which tab is currently active.

---
*PR created automatically by Jules for task [6880124825523721665](https://jules.google.com/task/6880124825523721665) started by @giauphan*

### DIFF
--- a/dashboard/src/components/Auth.tsx
+++ b/dashboard/src/components/Auth.tsx
@@ -103,13 +103,15 @@ const Auth: React.FC<AuthProps> = ({ onLogin }) => {
         </header>
 
         {/* Tab Switcher */}
-        <div style={{ display: 'flex', background: 'rgba(0,0,0,0.3)', padding: '0.4rem', borderRadius: '16px', marginBottom: '2.5rem', border: '1px solid rgba(255,255,255,0.05)' }}>
+        <div role="tablist" style={{ display: 'flex', background: 'rgba(0,0,0,0.3)', padding: '0.4rem', borderRadius: '16px', marginBottom: '2.5rem', border: '1px solid rgba(255,255,255,0.05)' }}>
           {[
             { id: 'token', label: 'TOKEN', icon: Key },
             { id: 'signin', label: 'SIGN IN', icon: LogIn },
           ].map((tab) => (
             <button
               key={tab.id}
+              role="tab"
+              aria-selected={mode === tab.id}
               disabled={loading}
               onClick={() => { setMode(tab.id as 'token' | 'signin'); setError(null); }}
               className={FOCUS_RING_CLASS}
--- a/dashboard/src/components/Auth.tsx
+++ b/dashboard/src/components/Auth.tsx
@@ -28,7 +28,7 @@ function getApiBase(): string {
 const TABS = [
   { id: 'token', label: 'TOKEN', icon: Key },
   { id: 'signin', label: 'SIGN IN', icon: LogIn },
-];
+] as const;
 
 const Auth: React.FC<AuthProps> = ({ onLogin }) => {
   const [mode, setMode] = useState<'token' | 'signin'>('token');
@@ -52,31 +52,37 @@ const Auth: React.FC<AuthProps> = ({ onLogin }) => {
     }
   };
 
-  const tabRefs = useRef<Map<string, HTMLButtonElement>>(new Map());
+  const tabRefs = useRef<Map<string, HTMLButtonElement> | null>(null);
+  const getTabRefs = () => {
+    if (!tabRefs.current) {
+      tabRefs.current = new Map();
+    }
+    return tabRefs.current;
+  };
   const [focusTarget, setFocusTarget] = useState<'token' | 'signin' | null>(null);
 
   useEffect(() => {
     if (focusTarget) {
-      tabRefs.current.get(focusTarget)?.focus();
+      getTabRefs().get(focusTarget)?.focus();
       setFocusTarget(null);
     }
   }, [focusTarget]);
 
-  const handleKeyDown = (e: React.KeyboardEvent<HTMLButtonElement>) => {
+  const handleKeyDown = (e: React.KeyboardEvent<HTMLDivElement>) => {
     const tabIds = TABS.map(t => t.id);
-    let nextTabId: 'token' | 'signin' | null = null;
+    let nextTabId: typeof mode | null = null;
 
     if (e.key === 'ArrowRight' || e.key === 'ArrowLeft') {
       e.preventDefault();
       const currentIndex = tabIds.indexOf(mode);
       const nextIndex = e.key === 'ArrowRight' ? (currentIndex + 1) % tabIds.length : (currentIndex - 1 + tabIds.length) % tabIds.length;
-      nextTabId = tabIds[nextIndex] as 'token' | 'signin';
+      nextTabId = tabIds[nextIndex];
     } else if (e.key === 'Home') {
       e.preventDefault();
-      nextTabId = tabIds[0] as 'token' | 'signin';
+      nextTabId = tabIds[0];
     } else if (e.key === 'End') {
       e.preventDefault();
-      nextTabId = tabIds[tabIds.length - 1] as 'token' | 'signin';
+      nextTabId = tabIds[tabIds.length - 1];
     }
 
     if (nextTabId) {
@@ -142,15 +148,15 @@ const Auth: React.FC<AuthProps> = ({ onLogin }) => {
         </header>
 
         {/* Tab Switcher */}
-        <div role="tablist" aria-label="Authentication method" style={{ display: 'flex', background: 'rgba(0,0,0,0.3)', padding: '0.4rem', borderRadius: '16px', marginBottom: '2.5rem', border: '1px solid rgba(255,255,255,0.05)' }}>
+        <div role="tablist" aria-label="Authentication method" onKeyDown={handleKeyDown} style={{ display: 'flex', background: 'rgba(0,0,0,0.3)', padding: '0.4rem', borderRadius: '16px', marginBottom: '2.5rem', border: '1px solid rgba(255,255,255,0.05)' }}>
           {TABS.map((tab) => (
             <button
               key={tab.id}
               ref={(el) => {
                 if (el) {
-                  tabRefs.current.set(tab.id, el);
+                  getTabRefs().set(tab.id, el);
                 } else {
-                  tabRefs.current.delete(tab.id);
+                  getTabRefs().delete(tab.id);
                 }
               }}
               role="tab"
@@ -158,9 +164,8 @@ const Auth: React.FC<AuthProps> = ({ onLogin }) => {
               aria-controls={`panel-${tab.id}`}
               id={`tab-${tab.id}`}
               tabIndex={mode === tab.id ? 0 : -1}
-              onKeyDown={handleKeyDown}
               disabled={loading}
-              onClick={() => { setMode(tab.id as 'token' | 'signin'); setError(null); }}
+              onClick={() => { setMode(tab.id); setError(null); }}
               className={FOCUS_RING_CLASS}
               style={{
                 flex: 1, display: 'flex', alignItems: 'center', justifyContent: 'center', gap: '0.5rem', padding: '0.75rem', borderRadius: '12px', outline: 'none', border: 'none', cursor: loading ? 'not-allowed' : 'pointer', fontSize: '0.75rem', fontWeight: 800, transition: 'all 0.3s',
--- a/dashboard/src/components/Auth.tsx
+++ b/dashboard/src/components/Auth.tsx
@@ -52,15 +52,12 @@ const Auth: React.FC<AuthProps> = ({ onLogin }) => {
     }
   };
 
-  const tabRefs = useRef<Map<string, HTMLButtonElement> | null>(null);
-  if (!tabRefs.current) {
-    tabRefs.current = new Map();
-  }
+  const tabRefs = useRef<Map<string, HTMLButtonElement>>(new Map());
   const [focusTarget, setFocusTarget] = useState<'token' | 'signin' | null>(null);
 
   useEffect(() => {
     if (focusTarget) {
-      tabRefs.current?.get(focusTarget)?.focus();
+      tabRefs.current.get(focusTarget)?.focus();
       setFocusTarget(null);
     }
   }, [focusTarget]);
@@ -151,9 +148,9 @@ const Auth: React.FC<AuthProps> = ({ onLogin }) => {
               key={tab.id}
               ref={(el) => {
                 if (el) {
-                  tabRefs.current?.set(tab.id, el);
+                  tabRefs.current.set(tab.id, el);
                 } else {
-                  tabRefs.current?.delete(tab.id);
+                  tabRefs.current.delete(tab.id);
                 }
               }}
               role="tab"
--- a/dashboard/src/components/Auth.tsx
+++ b/dashboard/src/components/Auth.tsx
@@ -145,7 +145,7 @@ const Auth: React.FC<AuthProps> = ({ onLogin }) => {
         </header>
 
         {/* Tab Switcher */}
-        <div role="tablist" style={{ display: 'flex', background: 'rgba(0,0,0,0.3)', padding: '0.4rem', borderRadius: '16px', marginBottom: '2.5rem', border: '1px solid rgba(255,255,255,0.05)' }}>
+        <div role="tablist" aria-label="Authentication method" style={{ display: 'flex', background: 'rgba(0,0,0,0.3)', padding: '0.4rem', borderRadius: '16px', marginBottom: '2.5rem', border: '1px solid rgba(255,255,255,0.05)' }}>
           {TABS.map((tab) => (
             <button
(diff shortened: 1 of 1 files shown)


giauphan:
cc @jules

## Review of PR: `dashboard/src/components/Auth.tsx`

The changes add ARIA roles/attributes to a tab interface. The ARIA implementation is correct in principle, but there are gaps to flag.

### Issues

**File: `dashboard/src/components/Auth.tsx` -> Line ~106 (tab button):**
Missing `aria-controls`. Each `role="tab"` button should have `aria-controls="<panel-id>"` to programmatically associate it with the tab panel it controls. Without this, screen readers can't announce which content the tab manages.

**File: `dashboard/src/components/Auth.tsx` -> Line ~109 (tab button):**
Missing keyboard navigation for the tablist. Adding `role="tab"` changes screen reader interaction expectations — users now expect Left/Right arrow keys to move between tabs and Home/End for first/last. The current code only has `onClick`, so keyboard-only users can't switch tabs with arrows. For WCAG 2.1 AA compliance, add a `onKeyDown` handler that implements the arrow-key navigation pattern.

**File: `dashboard/src/components/Auth.tsx` -> Line ~107 (tab button):**
`tabIndex` conflict with native button semantics. All three buttons are natively `tabIndex={0}`. With `role="tab"`, only the selected tab should have `tabIndex={0}` and the rest `tabIndex={-1}` (so Tab moves to the tablist, then arrows move between tabs). Currently all tabs are individually in the tab order, which doesn't match the ARIA tab pattern and creates redundant Tab stops.

### Summary

Good direction — the roles and `aria-selected` are correct. To be fully compliant and not regress keyboard usability, add `aria-controls`, a `tabIndex` management strategy (selected = 0, others = -1), and arrow-key (`onKeyDown`) navigation. If `aria-controls` is already handled on the panel side, that concern is resolved; either way the keyboard navigation is the most important gap to address.

---

giauphan:
cc @jules

I'll review this PR's changes to the `Auth.tsx` tab component for accessibility.

## Review

This PR adds proper WAI-ARIA tablist semantics and keyboard navigation to the auth tab switcher. Overall it's a solid, well-implemented accessibility improvement.

**What's good:**
- `role="tablist"` / `role="tab"` / `role="tabpanel"` with matching `aria-controls`/`aria-labelledby`/`aria-selected` — correct ARIA wiring.
- Keyboard handlers (←/→ wrap, Home/End) follow the [ARIA authoring practices](https://www.w3.org/WAI/ARIA/apg/patterns/tabpanel/) for tablist navigation.
- `tabIndex={mode === tab.id ? 0 : -1}` properly manages focus order.

**One maintainability issue:**

`dashboard/src/components/Auth.tsx` -> the `onKeyDown` handler (around line 14 in the new code) hardcodes `const tabs = ['token', 'signin']` separately from the tab config array rendered just above it. If a third tab is added later, a developer must remember to update this second list or keyboard navigation silently breaks for the new tab. Deriving the tab order from the existing `.map` data would eliminate this duplication risk.

---

giauphan:
cc @jules

The file isn't in this repo — this is a PR diff being submitted for review. Let me analyze the diff directly.

## Review: `dashboard/src/components/Auth.tsx`

**Summary:** Adds WAI-ARIA tablist accessibility (keyboard nav, ARIA roles, focus management) to a 2-tab auth switcher. The accessibility intent is correct, but there are implementation issues worth addressing.

### Issues Found

**1. File: `dashboard/src/components/Auth.tsx` -> Lines ~110-130: Direct DOM access via `document.getElementById` instead of React refs**

```jsx
document.getElementById(`tab-${tabs[nextIndex]}`)?.focus();
```

Using `document.getElementById` bypasses React's ref system. This is fragile — if multiple instances of this component render, ID collisions cause focus to land on the wrong element. It also breaks under React concurrent rendering. Use `useRef` or callback refs instead.

**2. File: `dashboard/src/components/Auth.tsx` -> Lines ~115-140: Repetitive `setError(null)` across all branches**

`setError(null)` is called in every branch (arrows, Home, End). It should be hoisted before the if/else to avoid duplication and ensure it always runs.

**3. File: `dashboard/src/components/Auth.tsx` -> Line ~108: Non-idiomatic use of `.map()` third parameter (`tabArray`)**

```jsx
].map((tab, _index, tabArray) => (
```

Using `tabArray` (the third `.map()` arg) to derive the tab list is non-obvious and confusing. Define the tabs as a named constant instead, so it can be referenced directly in both the render and the keyboard handler.

**4. File: `dashboard/src/components/Auth.tsx` -> Lines ~112-140: Missing `preventDefault()` on arrow keys**

Arrow key presses on buttons can scroll the page in some browsers/containers. WAI-ARIA recommends calling `e.preventDefault()` for tablist arrow navigation to prevent unintended scrolling.

**5. File: `dashboard/src/components/Auth.tsx` -> Lines ~112-140: Complex inline handler — not testable**

The entire keyboard navigation logic is an inline arrow function with no unit tests. Extracting it to a named handler or custom hook would make the a11y behavior testable.

### What's Good

- Correct `role="tablist"` / `role="tab"` / `role="tabpanel"` semantics
- Proper `aria-selected`, `aria-controls`, `aria-labelledby`, `tabIndex` attributes
- Keyboard navigation follows WAI-ARIA roving tabindex pattern (arrows, Home, End)
- Optional chaining (`?.focus()`) prevents null reference crashes

### Verdict

Accessibility is properly implemented at the semantic level, but the DOM manipulation approach (`document.getElementById`), repetitive code patterns, and lack of testability should be cleaned up before merge. The ID collision risk is the most impactful issue — fix that with refs.

---

giauphan:
cc @jules

Looking at this PR, I can see it's adding keyboard navigation and ARIA accessibility attributes to a tab switcher in `Auth.tsx`. Let me review the diff for issues.

The changes add:
- A `TABS` constant extracted from inline array
- `useRef` for tracking tab button elements
- `handleKeyDown` for ArrowLeft/ArrowRight/Home/End navigation
- ARIA roles (`tablist`, `tab`, `tabpanel`) and attributes (`aria-selected`, `aria-controls`, `tabIndex`, `aria-labelledby`)
- Focus management via `setTimeout` after state updates

**Assessment:**

This is a well-implemented accessibility improvement. The code follows the WAI-ARIA roving tabindex pattern correctly:

- Active tab gets `tabIndex={0}`, inactive tabs get `tabIndex={-1}`
- Keyboard handler on individual tab buttons (valid pattern)
- Callback ref properly manages the Map of button refs
- `setTimeout(0)` correctly defers focus until after React's DOM update
- ARIA attributes are correctly applied and consistent (e.g., `aria-controls` matches panel `id`, `aria-labelledby` matches tab `id`)
- `role="tablist"` on the container, `role="tab"` on buttons

No correctness, security, performance, or maintainability issues found. The non-null assertion `nextTabId!` inside the `setTimeout` is safe since it's guarded by the preceding `if (nextTabId)` check. The hardcoded `'token' | 'signin'` types in `handleKeyDown` are consistent with the `TABS` constant and `mode` state.

This PR is clean.

---

giauphan:
cc @jules

The file isn't in the current working tree (expected for a PR diff). I have the full diff, so let me review the changes.

## Review

Overall this is a clean, well-implemented accessibility improvement — adds roving tabindex, ARIA roles/attributes, and arrow-key/Home/End navigation to the tab switcher. No correctness bugs, security issues, or performance concerns.

Two minor observations:

**1. `setTimeout` focus management (code quality)**

`dashboard/src/components/Auth.tsx` -> Line 64-66: The `setTimeout` with `0` delay is a workaround to defer `.focus()` until after React re-renders. If the component unmounts before the callback fires, the focus call hits a detached element (silent no-op in practice, no error). A more idiomatic approach would be a `useEffect` keyed on the target tab id — it's cleaner and eliminates the race window entirely.

**2. Redundant `new Map()` allocation (performance, negligible)**

`dashboard/src/components/Auth.tsx` -> Line 53: `useRef<Map<string, HTMLButtonElement>>(new Map())` constructs a new `Map` on every render. `useRef` discards it after the first render, so it's functionally correct but wasteful. Minor.

Everything else — ARIA wiring (`role="tablist"`/`"tab"`/`"tabpanel"`, `aria-selected`, `aria-controls`/`aria-labelledby` linkage), the ref-callback cleanup that deletes entries from the Map, and the `Home`/`End`/arrow navigation logic — is correct.

**Verdict: Looks good. Ship it.**

---

giauphan:
cc @jules

The `dashboard/src/components/Auth.tsx` file isn't in this local checkout — it's from the dashboard sub-project. I'll review based on the diff provided.

## Review

This PR adds keyboard navigation (ArrowLeft/Right, Home, End) and full ARIA attributes to the auth tab switcher. Overall, it's a solid accessibility improvement.

**Correctness**: All passes. The roving tabindex pattern (`tabIndex={mode === tab.id ? 0 : -1}`), focus management via `useRef<Map>` + `useEffect`, and circular index math are correct. The `e.preventDefault()` on navigation keys prevents unwanted page scrolling.

**Security**: Clean — no secrets, no injection vectors.

**Performance**: Fine for a 2-tab component. Minor: the inline `ref` callback causes a delete+set cycle on every render (React calls old ref with `null`, then new ref with the element), but this is negligible here.

**Code quality**: The `TABS` constant extraction is good. The `as 'token' | 'signin'` casts are needed because `TABS` is inferred with `id: string` rather than the literal union — typing the constant explicitly would remove the casts, but it's not a bug.

**One issue that matters:**

`dashboard/src/components/Auth.tsx` → Line ~110 (the `role="tablist"` container):
The `tablist` div lacks an accessible name. Add `aria-label` (e.g., `aria-label="Authentication method"`) so screen readers can announce what the tab list is for. Without it, users hear "tab bar" with no context.

**Minor note (not a bug):**
`role="tabpanel"` is applied to `<motion.form>` elements. This overrides the native `form` role, which is valid per ARIA but unconventional. Functionally correct — the form's submit behavior is unaffected.

Everything else looks good.

---

giauphan:
cc @jules

I've examined the PR diff and the project context. The `dashboard/` directory doesn't exist in the main branch — this appears to be a PR (likely from a fork) that adds a web dashboard with authentication. Let me review the diff carefully.

---

## PR Review: `dashboard/src/components/Auth.tsx`

**Overall:** This PR adds proper accessibility (WAI-ARIA roles, roving tabindex, keyboard navigation) to an auth tab switcher with TOKEN and SIGN IN modes. The implementation is **correct** and follows established React/RoW conventions.

### What's right

- **Roving tabindex** is correctly implemented: `tabIndex={mode === tab.id ? 0 : -1}` — only the selected tab is in the tab order.
- **Keyboard navigation** handles ArrowRight, ArrowLeft (with wrap-around), Home, End. All call `e.preventDefault()` only when they intercept a key.
- **Focus management** is sound: `setMode()` and `setFocusTarget()` are batched in the same render cycle, so when the `useEffect` fires, the target button already has `tabIndex={0}` and is focusable. Setting `focusTarget(null)` afterward clears transient state cleanly without looping.
- **ARIA relationships** are correctly linked: `aria-controls="panel-${id}"` ↔ `id="panel-${id}"`, `aria-labelledby="tab-${id}"` ↔ `id="tab-${id}"`.
- Extracting `TABS` to a module-level constant avoids unnecessary array recreation on each render.

### Minor notes (no bugs)

`dashboard/src/components/Auth.tsx` -> Line ~56: `tabRefs.current?.set(...)` uses redundant optional chaining — `tabRefs.current` is initialized during render (before DOM commit and ref callbacks), so it's always non-null by the time the ref callback fires. Harmless, but the `?.` can be dropped for clarity.

`dashboard/src/components/Auth.tsx` -> Line ~55: `useRef<Map<string, HTMLButtonElement> | null>(null)` with manual `if (!tabRefs.current)` init is correct but could be simplified to `useRef(new Map())` (the `new Map()` on every render is discarded by React). The current pattern is fine just slightly more verbose.

**No correctness, security, performance, or maintainability issues found.** The code is clean and the ARIA implementation is proper.

---

giauphan:
cc @jules

This is a PR review task — the file `dashboard/src/components/Auth.tsx` is from a different repository (the dashboard app), and I'm reviewing the diff provided.

## Review: `dashboard/src/components/Auth.tsx`

**Correctness:** No bugs found. The tab navigation logic is sound — arrow keys wrap correctly, Home/End go to first/last tab, and the `useEffect` + `focusTarget` pattern properly moves focus. The `TABS` constant extraction eliminates duplication between the button map and the inline array.

**Security:** No issues. No hardcoded secrets, no injection vectors. The `tab.id` values come from a static constant.

**Performance:** The `ref` callback on each tab is recreated on every render (not wrapped in `useCallback`), causing React to call the ref with `null` then the element on each render. This is functionally harmless but causes unnecessary work. For a component with only 2 tabs and infrequent re-renders, negligible.

**Code quality:** Clean. The ARIA attributes are consistent (`aria-labelledby` on panel → tab `id`, `aria-controls` on tab → panel `id`, `tabIndex` toggles between 0/-1). One minor point: `role="tabpanel"` is applied to `motion.form` — forms aren't typically tabpanels, but it's functionally correct here since the panels are form containers.

**Maintainability:** Good. The keyboard handler is well-structured and the TABS constant makes adding new tabs trivial. Error clearing on tab switch (`setError(null)`) prevents stale errors from the previous tab.

**One minor accessibility note:** `aria-orientation` defaults to `"horizontal"` for `role="tablist"`, which matches the left-to-right layout, so no change needed there.

No blocking issues. This PR is ready to merge.

---

giauphan:
cc @jules

## PR Review: Keyboard navigation + ARIA tabs for Auth component

This is a solid accessibility improvement — adds arrow-key/Home/End tab switching, roving `tabIndex`, and proper `role="tablist"`/`role="tab"`/`role="tabpanel"` semantics with `aria-selected`/`aria-controls`/`aria-labelledby`. No correctness, security, or performance issues of note. A couple of minor nits (non-blocking):

- **Line ~57** — `useRef(new Map())` allocates a fresh `Map` instance on every render; it's discarded so it's not a bug, just minor waste. Could lazy-init if desired.
- **Line ~53** — `TABS`' `id` fields are inferred as `string`, so the `tab.id as 'token' | 'signin'` casts are unchecked. `as const` on `TABS` would make them `('token' | 'signin')` and remove the casts (catches typos at compile time).
- **Line ~16** — `handleKeyDown` is recreated each render and attached to every tab button; fine for two tabs, but if more tabs are added it'd be cleaner to rely solely on the tablist-level keydown pattern.

Minor note: inactive panels are unmounted by `AnimatePresence` (mode="wait"), so an inactive tab's `aria-controls` can briefly dangle — acceptable and common.

Overall: looks good to merge. Good use of `requestAnimationFrame`-free focus sync and proper roving tabindex.

---

giauphan:
cc @jules

The file isn't in this repo — the PR is for a different project. No matter, the diff is self-contained. Let me review it thoroughly.

---

**Review: Auth.tsx — Tab switcher accessibility + keyboard navigation**

**Overall: Looks good.** The PR adds proper ARIA tab roles, keyboard navigation (ArrowLeft/Right, Home, End), and deferred focus management to the Auth component's tab switcher. The implementation follows the WAI-ARIA tabs pattern correctly.

**Correctness — No bugs found.**
- `TABS` extracted as `as const`, so `tab.id` is `'token' | 'signin'` — the removed cast (`as 'token' | 'signin'`) was indeed no longer needed. Type checks out.
- `useEffect` runs after commit phase, so `getTabRefs().get(focusTarget)` always finds the button element (refs are attached before effects fire). Focus management is correct.
- `handleKeyDown` reads `mode` from the current render closure; since it's a non-memoized function, it always captures the latest state. No stale closure.
- `ref` callback correctly handles mount (`el` truthy → `set`) and unmount (`el` null → `delete`).
- `Home`/`End`/arrow logic handles edge cases (wraps with modulo, single-tab case works).

**Security — No issues.** No secrets, no injection vectors, no unsafe DOM access.

**Performance — No issues.** Trivial Map operations; no re-renders beyond the expected state transitions.

**Code quality — Minor concern.**
- The `tabRefs: useRef<Map<string, HTMLButtonElement> | null>(null)` + `getTabRefs()` lazy-init pattern works, but it's slightly over-engineered for two buttons. Individual `useRef<HTMLButtonElement | null>(null)` per tab would be simpler and avoid the Map allocation entirely.
- The inline `ref={(el) => ...}` callback is recreated every render. React will call the old callback with `null` then the new one with the element on each render, causing unnecessary `delete`+`set` churn in the Map. Wrapping in `useCallback` or using direct refs would avoid this. This is cosmetic — no functional impact.

**Maintainability — Good.** The `TABS` constant eliminates the duplication that existed in the original inline array. ARIA attributes (`role`, `aria-selected`, `aria-controls`, `aria-labelledby`, `tabIndex`) are consistent between tabs and panels (e.g., `tab-token` ↔ `panel-token`, `tab-signin` ↔ `panel-signin`).

---

PR #271, opened 2026-09-05
🎨 Palette: [UX improvement] Add password visibility toggles to Auth inputs

giauphan:
💡 What: Added password visibility toggles (Eye/EyeOff) to the Neural Access Key and Password input fields on the Auth screen.
🎯 Why: Allowing users to easily show/hide their credentials is a crucial UX improvement, especially when pasting complex API keys or long passwords where mistakes are common.
📸 Before/After: Visual change in the input fields where an eye icon allows toggling visibility.
♿ Accessibility: The buttons are fully keyboard navigable (`FOCUS_RING_CLASS`), use `type="button"` to avoid accidental form submissions, and include dynamic `aria-label` tags (e.g., "Show password" / "Hide password") for screen readers.

---
*PR created automatically by Jules for task [2891596205406985685](https://jules.google.com/task/2891596205406985685) started by @giauphan*

### DIFF
--- a/dashboard/src/components/Auth.tsx
+++ b/dashboard/src/components/Auth.tsx
@@ -9,7 +9,7 @@
 import React, { useState, useRef, useEffect } from 'react';
 import { motion, AnimatePresence } from 'framer-motion';
 import {
-  Shield, Key, Mail, Lock, LogIn, Loader2
+  Shield, Key, Mail, Lock, LogIn, Loader2, Eye, EyeOff
 } from 'lucide-react';
 import { safeSessionStorageSetItem, safeSessionStorageGetItem, safeSessionStorageRemoveItem } from '../lib/safeSessionStorage';
 import { storeAuthTokens } from '../lib/auth';
@@ -37,6 +37,8 @@ const Auth: React.FC<AuthProps> = ({ onLogin }) => {
   const [apiKey, setApiKey] = useState('');
   const [loading, setLoading] = useState(false);
   const [error, setError] = useState<string | null>(null);
+  const [showToken, setShowToken] = useState(false);
+  const [showPassword, setShowPassword] = useState(false);
 
   const handleTokenSubmit = async (e: React.FormEvent) => {
     e.preventDefault();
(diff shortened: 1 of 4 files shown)


giauphan:
cc @jules

I see the Auth.tsx file isn't in this repo directly — the diff represents a feature change targeting the dashboard component. Let me review it based on the diff itself.

**File: `dashboard/src/components/Auth.tsx`**

**Correctness:**
- No logic bugs detected. Toggle state management (`showToken`, `showPassword`) is straightforward and correct.
- `type="button"` on toggle buttons correctly prevents accidental form submission.

**Security:**
- No issues. Input types correctly default to `"password"` for both fields.
- The API key field maintains its `required autoFocus disabled={loading}` attributes through the change.
- `autoComplete="current-password"` is preserved on the password field.

**Performance:**
- Negligible. Two new `useState` calls; each toggle causes a single re-render. No unnecessary re-renders introduced.

**Code quality — 1 minor note:**
- The inline `style` objects on the toggle buttons add ~70 bytes of duplicated style text vs extracting them to constants, but this is consistent with the existing convention in this file (heavy inline styles everywhere). Not worth changing, just noting.

**Maintainability:**
- The pattern is clean and DRY enough — same structure repeated for both inputs, which mirrors the existing symmetry in the component. If a third sensitive field appears, that duplication might warrant abstraction, but 2 is fine.

**Verdict:** LGTM. Clean addition, no issues that matter.

---

giauphan:
cc @jules

Reviewing the changes:

**What it does:** Adds "show/hide" toggle buttons for the API key and password input fields in the Auth component, using `Eye`/`EyeOff` icons from lucide-react.

**Issues found:**

1. **File: `dashboard/src/components/Auth.tsx` -> Line 185-192:** The show/hide button for the API key uses inline styles but applies `className={FOCUS_RING_CLASS}` which is inconsistent — the button also has its own `cursor`, `borderRadius`, etc. as inline styles. This dual approach (inline + class) is fragile and makes the focus ring behavior unpredictable across browsers. Consider either fully inline or fully via CSS class.

2. **File: `dashboard/src/components/Auth.tsx` -> Lines 185-192, 216-223:** Both buttons are visually identical (same position, padding, colors, sizing). This is duplicated layout code that could be extracted into a reusable `PasswordToggle` or `VisibilityToggle` component to reduce duplication.

3. **Minor - File: `dashboard/src/components/Auth.tsx` -> Lines 184-191:** The button toggles between `type="text"` and `type="password"` but doesn't set an `autocomplete` attribute. While not critical for API keys, the password field already has `autoComplete="current-password"` which is good practice preserved here. No issue per se, but worth noting the pattern differs between the two fields.

4. **No security issues.** Toggling to `type="text"` is standard UX practice (GitHub, GitLab, etc. do this). The inputs remain `type="password"` by default. No secrets are hardcoded.

**Overall:** Clean, straightforward feature addition. The main concern is the duplicated inline styling across both toggle buttons — worth extracting if you plan more similar inputs elsewhere. Otherwise LGTM.

---

giauphan:
cc @jules

I'm reviewing this PR which adds show/hide toggle buttons for password and API key input fields in an Auth component.

**Overall: Clean change, but has one UX issue.**

Issues that matter:

1. **dashboard/src/components/Auth.tsx -> Line 190**: The "Show/Hide" buttons overlap with `autoFocus` on the apiKey input. On mount, clicking the input (which focuses it) will also trigger the button click if the user's cursor lands slightly right — they'll accidentally toggle visibility instead of typing. Consider removing `autoFocus` or shifting the button further right / making it larger so the hit area doesn't conflict.

2. **No other issues found** — the state management is correct, aria-labels are properly toggled, imports are clean, and the inline styles match existing patterns in the file.

---

giauphan:
cc @jules

Done! Here's my review of the PR:

**Overall:** Clean, low-risk change. Standard show/hide password toggle pattern applied consistently to both credential inputs. No blocking issues.

**Minor observations (non-blocking):**

- **Style scoping risk** — The `.password-toggle-btn` rule added via `<style>` is global within the component's injected sheet. Currently harmless since no other element uses that class name, but worth noting. Same applies to `.glass-input` which was already present. Consider using CSS modules or a unique prefix if more dynamic styles get added.

- **Hardcoded color in CSS** — `.password-toggle-btn` uses `color: var(--text-muted)` which matches the existing design system. Good, no issue.

- **Slight duplication** — The toggle button structure repeats for both inputs. Could be extracted into a shared `PasswordToggle` component if more password fields appear, but fine as-is for now.

- **`autoComplete` values** — API key uses `autoComplete="off"` and password uses `autoComplete="current-password"`. Both reasonable. Some browsers may still override `off`, but that's expected behavior.

No security, correctness, or performance issues. Ship it.

---

PR #274, opened 2026-09-05
⚡ Bolt: Batch concept extraction for embeddings and database inserts

giauphan:
💡 What: Refactored `extractConcepts` to accumulate embeddings tasks instead of processing them iteratively. It collects input strings, makes a single batched call to `generateEmbeddingsBatch()`, and then utilizes `db.executeMany()` to batch the `INSERT`s into the database.

🎯 Why: The original loop caused an N+1 performance bottleneck for both HTTP requests (against the embedding provider) and I/O latency against the local database for each created concept, significantly slowing down operations at scale.

📊 Impact: Decreased latency. API calls for embeddings drop from O(N) to O(N/batchSize), and Database round-trips drop from O(N) to O(1). 

🔬 Measurement: Check network timing traces or console logs profiling how quickly consolidation engines run batch concept extraction under high volume.

---
*PR created automatically by Jules for task [14467396394577415561](https://jules.google.com/task/14467396394577415561) started by @giauphan*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -1,51 +1,59 @@
-## 2026-08-16 - Array processing in O(N^2) loops
+## 2026-09-05 - Array processing in O(N^2) loops
 **Learning:** When performing pairwise distance checks (e.g. cosine similarity) over N elements, checking types recursively inside the inner N^2 loop creates significant overhead. In Javascript/Typescript, repeated Array operations like Array.isArray, array property index reads, or mapping arrays directly impacts garbage collection heavily on nested loops.
 **Action:** Extract bounds checking, array property type assertions, and invalid structure exclusion into a dedicated O(N) grouping preprocessing step so the inner loops only deal with guaranteed valid references.
 
-## 2026-08-16 - Type Safety in Array Property Accesses
+## 2026-09-05 - Type Safety in Array Property Accesses
 **Learning:** When performing length calculations on arrays, structural typing allows objects shaped like `{ length: 5 }` to bypass standard checks, leading to NaN errors in math computations. Native type protections like `Array.isArray()` and `instanceof` checks are critical to guard against prototype poisoning.
 **Action:** Use defensive `Array.isArray(x) || x instanceof Float32Array` validation rather than just relying on length presence prior to array execution functions.
 
-## 2026-08-17 - O(N^2) Math GC Overhead
+## 2026-09-05 - O(N^2) Math GC Overhead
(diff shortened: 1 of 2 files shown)


giauphan:
cc @jules

I've reviewed this PR. Here's my assessment:

The main change in `consolidationEngine.ts` refactors concept creation from sequential per-concept embedding generation + individual DB inserts to batched embedding API calls + batched DB inserts — a solid N+1 fix aligned with the project's optimization rules.

Minor issues:

`.jules/bolt.md` -> Line 44-50: Duplicate entry — "External API Generation Batching" appears twice verbatim. Should deduplicate.

`.jules/bolt.md` -> Line 1: Bulk date rebase from `2026-08-XX` to `2026-09-05` across many entries — looks intentional but the dates don't reflect when these were actually learned. Consider keeping original dates as learnings are historical.

`src/services/consolidationEngine.ts` -> Line 357: After `generateEmbeddingsBatch(embedTexts, 'passage')`, there's no guard that `embeddings?.length === embedTexts.length`. If the external API silently drops failed items (returns fewer embeddings than requested), accessing `embeddings[i]` will yield `undefined` and that concept gets silently skipped. Worth adding a sanity check or at minimum a log.

`src/services/consolidationEngine.ts` -> Line 388: Comment says "Bolt Optimization" — stale reference (should say something like "Consolidation Engine Optimization").

The overall approach is correct: collect concepts → batch embeddings → batch insert. Good improvement.

---

giauphan:
cc @jules

Here's my review of this PR:

**Correctness:** Logic is sound — `newConcepts[i]` correctly aligns with `embeddings[i]` after the batch call, and the fallback check (`embeddings && embeddings[i]`) handles partial failures gracefully.

**Security:** No issues. The `conceptLabel.slice(0, 255)` truncation is still applied before insert. No new injection vectors introduced.

**Performance:** This is the entire purpose of the change. Eliminating N sequential `generateEmbeddingsBatch()` calls into one batch request should dramatically reduce latency for large cluster sets (e.g., 50 concepts → 1 HTTP round-trip instead of 50). Using `executeMany()` over individual INSERTs is a well-known SQLite/Oracle optimization that also avoids transaction overhead per row.

**Code quality:** Clean refactor. The old code had an inline `created++` that no longer fires — now `created += insertBinds.length` which only counts rows that actually succeeded (embedding validation could skip a row). The learning note in `.jules/bolt.md` is appropriate for future reference.

**Maintainability:** Embedding generation logic is now centralized and easier to test. One minor nit: `0.70` confidence is hardcoded again (same as before), but since this is pre-existing it's out of scope for this PR.

**No blocking issues.** LGTM — good performance win with no trade-offs.

---

PR #294, opened 2026-09-09
⚡ Bolt: Replace O(N) array indexOf with O(1) Map lookup in sort

giauphan:
💡 What: Replaced an $O(N)$ `indexOf` lookup inside an array sort comparator in `mcpTools.ts` with a hoisted, $O(1)$ constant `Map` lookup.
🎯 Why: Array `indexOf` operations within a sort comparator result in $O(N \log N \cdot K)$ performance (where $K$ is the length of the priority array), introducing unnecessary overhead.
📊 Impact: Eliminates redundant array traversals and garbage collection in hot path sorting logic.
🔬 Measurement: Verify changes by checking that all tests pass (`pnpm test`), ensuring the sort order output remains consistent.

---
*PR created automatically by Jules for task [6022414751536011626](https://jules.google.com/task/6022414751536011626) started by @giauphan*

### DIFF
--- a/src/presentation/mcpTools.ts
+++ b/src/presentation/mcpTools.ts
@@ -453,7 +453,7 @@ export function registerTools(server: McpServer, sessionAuth?: { tier: string; u
 
       // Truncate if too many nodes
       if (nodes.length > max) {
-        // ⚡ Bolt Optimization: Replace O(N) array indexOf with O(1) Map lookup in sort
+        // Use Map for O(1) priority lookup
         nodes.sort((a, b) => {
           const ia = NODE_PRIORITY_ORDER.get(a.type) ?? 99;
           const ib = NODE_PRIORITY_ORDER.get(b.type) ?? 99;
--- a/src/presentation/mcpTools.ts
+++ b/src/presentation/mcpTools.ts
@@ -79,6 +79,8 @@ function a2aReal(name: string, description: string, params: string[], handler: (
 
 import { injectAuthContext } from '../utils/authContext.js';
 
+const NODE_PRIORITY_ORDER = new Map([["module", 0], ["class", 1], ["function", 2], ["variable", 3]]);
+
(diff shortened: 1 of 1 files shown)


giauphan:
cc @jules

## Code Review

The change looks good overall — it's a minor but clean optimization.

### Minor Issues

**File: `src/presentation/mcpTools.ts` -> Line 82**: The module-level `Map` is fine for performance, but the comment "⚡ Bolt Optimization" on line 456 is noisy/branding fluff. Consider a plain comment like `// Use Map for O(1) priority lookup` or just remove the comment entirely — the code is self-explanatory.

**Performance reality check**: `Array.indexOf` on a 4-element array is effectively O(1) in practice (constant-bounded, CPU-cache-friendly). The `Map.get` swap won't produce a measurable difference here. That said, the `Map` version is arguably more idiomatic and readable, so it's still a net positive — just don't expect actual performance gains.

### Everything Else

- **Correctness**: Behavior is preserved. The fallback `?? 99` matches the original `ia === -1 ? 99 : ia` logic. ✅
- **Security**: No issues. ✅
- **TypeScript**: No `any` types introduced. The `Map` is properly typed via inference. ✅
- **Maintainability**: Hoisting the priority map to module scope is fine since it's immutable data. Consider adding `as const` or a `readonly` annotation if you want to signal immutability more explicitly, though it's not critical since `Map` doesn't have a built-in readonly variant. ✅

**Verdict**: Approve — clean, correct, low-risk change.

---

giauphan:
cc @jules

## Code Review

The change looks good. It's a minor but clean optimization — replacing `Array.indexOf()` (O(n)) with `Map.get()` (O(1)) for priority lookups during sorting, and hoisting the constant out of the function scope to avoid re-creation on each call.

No issues to flag. The logic is equivalent, the `?? 99` fallback correctly handles unknown types the same way the original `-1 ? 99` check did, and the module-level `const` with a `Map` is appropriate here.

---

PR #295, opened 2026-09-09
🛡️ Sentinel: [MEDIUM] Fix HTTP Parameter Pollution (HPP) vulnerability by blocking array query parameters

giauphan:
🚨 **Severity:** MEDIUM
💡 **Vulnerability:** Express natively parses multiple identical query parameters (e.g., `?id=1&id=2`) into an array instead of a string. Route endpoints were type-casting `req.query.*` values as `string` without runtime verification, and passing those values down to database layers or `.trim()` methods, creating Denial of Service (DoS) conditions and risking parameter injection attacks.
🎯 **Impact:** An attacker could craft multi-parameter URLs causing internal 500 errors (DoS via unhandled exceptions) or bypass simplistic regex validations before backend database evaluation.
🔧 **Fix:** Inserted runtime validation using `Array.isArray()` across all affected routes. If an array is detected in `req.query` parameters expected to be primitives, the endpoint immediately halts execution and returns a `400 Bad Request`.
✅ **Verification:** Verified that array bounds are correctly intercepted across presentation logic, and all regression test suites pass (`pnpm test`, `pnpm run build`, and `pnpm lint`).

---
*PR created automatically by Jules for task [3545799732713618685](https://jules.google.com/task/3545799732713618685) started by @giauphan*

### DIFF
--- a/src/middleware/validation.ts
+++ b/src/middleware/validation.ts
@@ -1,5 +1,4 @@
 import express from "express";
-import { logger } from "../utils/logger.js";
 
 /**
  * Middleware to reject array query parameters to prevent HTTP Parameter Pollution (HPP).
@@ -7,15 +6,13 @@ import { logger } from "../utils/logger.js";
  */
 export function rejectArrayParams(...paramNames: string[]) {
   if (paramNames.length === 0) {
-    logger.warn("[rejectArrayParams] Middleware created with empty paramNames array.");
+    throw new Error("[rejectArrayParams] Middleware must be initialized with at least one parameter name.");
   }
 
   return (req: express.Request, res: express.Response, next: express.NextFunction): void => {
-    if (paramNames.length === 0) return next();
-
-    const offending = paramNames.filter(p => typeof req.query[p] !== 'undefined' && typeof req.query[p] !== 'string');
+    const offending = paramNames.filter(p => Array.isArray(req.query[p]));
     if (offending.length > 0) {
-      res.status(400).json({ error: `Bad Request: invalid parameter type for [${offending.join(', ')}], string expected` });
+      res.status(400).json({ error: `Bad Request: array parameters not supported for [${offending.join(', ')}]` });
       return;
     }
     next();
--- a/src/middleware/validation.ts
+++ b/src/middleware/validation.ts
@@ -1,0 +1,15 @@
+import express from "express";
+
+/**
+ * Middleware to reject array query parameters to prevent HTTP Parameter Pollution (HPP).
+ * @param paramNames - List of query parameter keys to validate.
+ */
+export function rejectArrayParams(...paramNames: string[]) {
+  return (req: express.Request, res: express.Response, next: express.NextFunction) => {
+    if (paramNames.some(p => Array.isArray(req.query[p]))) {
+      res.status(400).json({ error: "Bad Request: array parameters not supported" });
+      return;
+    }
+    next();
+  };
+}
--- a/src/middleware/validation.ts
+++ b/src/middleware/validation.ts
@@ -1,4 +1,5 @@
 import express from "express";
+import { logger } from "../utils/logger.js";
 
 /**
  * Middleware to reject array query parameters to prevent HTTP Parameter Pollution (HPP).
@@ -6,6 +7,11 @@ import express from "express";
  */
 export function rejectArrayParams(...paramNames: string[]) {
   return (req: express.Request, res: express.Response, next: express.NextFunction): void => {
+    if (paramNames.length === 0) {
+      logger.warn("[rejectArrayParams] Middleware applied with empty paramNames array. Skipping.");
+      return next();
+    }
+
     if (paramNames.some(p => typeof req.query[p] !== 'undefined' && typeof req.query[p] !== 'string')) {
       res.status(400).json({ error: "Bad Request: invalid parameter type, string expected" });
       return;
--- a/src/middleware/validation.ts
+++ b/src/middleware/validation.ts
@@ -5,9 +5,9 @@ import express from "express";
  * @param paramNames - List of query parameter keys to validate.
  */
 export function rejectArrayParams(...paramNames: string[]) {
-  return (req: express.Request, res: express.Response, next: express.NextFunction) => {
-    if (paramNames.some(p => Array.isArray(req.query[p]))) {
-      res.status(400).json({ error: "Bad Request: array parameters not supported" });
+  return (req: express.Request, res: express.Response, next: express.NextFunction): void => {
+    if (paramNames.some(p => typeof req.query[p] !== 'undefined' && typeof req.query[p] !== 'string')) {
+      res.status(400).json({ error: "Bad Request: invalid parameter type, string expected" });
       return;
     }
     next();
--- a/src/middleware/validation.ts
+++ b/src/middleware/validation.ts
@@ -1,7 +1,8 @@
 import express from "express";
 
 /**
- * Middleware to reject array query parameters to prevent HTTP Parameter Pollution (HPP).
+ * Middleware to reject non-string query parameters (arrays, objects) to prevent HTTP Parameter Pollution (HPP).
+ * Note: Only validates parameters that are present in the request. `undefined` parameters pass through untouched.
  * @param paramNames - List of query parameter keys to validate.
  */
 export function rejectArrayParams(...paramNames: string[]) {
@@ -15,7 +16,7 @@ export function rejectArrayParams(...paramNames: string[]) {
       return val !== undefined && typeof val !== 'string';
     });
     if (offending.length > 0) {
-      res.status(400).json({ error: "Bad Request: malformed query parameters" });
+      res.status(400).json({ error: `Bad Request: malformed query parameters: ${offending.join(", ")}` });
       return;
     }
     next();
--- a/src/middleware/validation.ts
+++ b/src/middleware/validation.ts
@@ -10,9 +10,12 @@ export function rejectArrayParams(...paramNames: string[]) {
   }
 
   return (req: express.Request, res: express.Response, next: express.NextFunction): void => {
-    const offending = paramNames.filter(p => Array.isArray(req.query[p]));
+    const offending = paramNames.filter(p => {
+      const val = req.query[p];
+      return val !== undefined && typeof val !== 'string';
+    });
     if (offending.length > 0) {
-      res.status(400).json({ error: `Bad Request: array parameters not supported for [${offending.join(', ')}]` });
+      res.status(400).json({ error: "Bad Request: malformed query parameters" });
       return;
     }
     next();
--- a/src/middleware/validation.ts
+++ b/src/middleware/validation.ts
@@ -6,14 +6,16 @@ import { logger } from "../utils/logger.js";
  * @param paramNames - List of query parameter keys to validate.
  */
 export function rejectArrayParams(...paramNames: string[]) {
-  return (req: express.Request, res: express.Response, next: express.NextFunction): void => {
-    if (paramNames.length === 0) {
-      logger.warn("[rejectArrayParams] Middleware applied with empty paramNames array. Skipping.");
-      return next();
-    }
+  if (paramNames.length === 0) {
+    logger.warn("[rejectArrayParams] Middleware created with empty paramNames array.");
+  }
 
-    if (paramNames.some(p => typeof req.query[p] !== 'undefined' && typeof req.query[p] !== 'string')) {
-      res.status(400).json({ error: "Bad Request: invalid parameter type, string expected" });
+  return (req: express.Request, res: express.Response, next: express.NextFunction): void => {
+    if (paramNames.length === 0) return next();
+
+    const offending = paramNames.filter(p => typeof req.query[p] !== 'undefined' && typeof req.query[p] !== 'string');
+    if (offending.length > 0) {
+      res.status(400).json({ error: `Bad Request: invalid parameter type for [${offending.join(', ')}], string expected` });
       return;
     }
     next();
--- a/src/presentation/a2a/heartbeatRoutes.ts
+++ b/src/presentation/a2a/heartbeatRoutes.ts
@@ -18,6 +18,11 @@ export function mountHeartbeatRoutes(app: express.Express): void {
 
   // Register an agent
   app.get("/a2a/register", a2aRateLimiter, authMiddleware, (req, res) => {
+    if (Array.isArray(req.query.agent_url) || Array.isArray(req.query.agent_name) || Array.isArray(req.query.capabilities)) {
+      res.status(400).json({ error: "Bad Request: array parameters not supported" });
+      return;
+    }
+
     const agentUrl = req.query.agent_url as string;
     const agentName = req.query.agent_name as string || "Unknown Agent";
     const capabilities = (req.query.capabilities as string || "").split(",").filter(Boolean);
--- a/src/presentation/a2a/heartbeatRoutes.ts
+++ b/src/presentation/a2a/heartbeatRoutes.ts
@@ -11,18 +11,14 @@ import express from "express";
 import { a2aRegistry } from "../../services/a2aRegistry.js";
 import { logger } from "../../utils/logger.js";
 import { authMiddleware } from "../../middleware/auth.js";
+import { rejectArrayParams } from "../../middleware/validation.js";
 import { countMatching } from "../../utils/array.js";
 import { a2aRateLimiter } from "./a2aRoutes.js";
 
 export function mountHeartbeatRoutes(app: express.Express): void {
 
   // Register an agent
-  app.get("/a2a/register", a2aRateLimiter, authMiddleware, (req, res) => {
-    if (Array.isArray(req.query.agent_url) || Array.isArray(req.query.agent_name) || Array.isArray(req.query.capabilities)) {
-      res.status(400).json({ error: "Bad Request: array parameters not supported" });
-      return;
-    }
-
+  app.get("/a2a/register", a2aRateLimiter, authMiddleware, rejectArrayParams("agent_url", "agent_name", "capabilities"), (req, res) => {
     const agentUrl = req.query.agent_url as string;
     const agentName = req.query.agent_name as string || "Unknown Agent";
     const capabilities = (req.query.capabilities as string || "").split(",").filter(Boolean);
--- a/src/presentation/consolidationRoutes.ts
+++ b/src/presentation/consolidationRoutes.ts
@@ -4,6 +4,7 @@
 import express from "express";
 import { consolidationEngine, type ConsolidationJob } from "../services/consolidationEngine.js";
 import { authMiddleware } from "../services/authService.js";
+import { rejectArrayParams } from "../middleware/validation.js";
 import { logger } from "../utils/logger.js";
 import rateLimit from "express-rate-limit";
 
@@ -32,13 +33,8 @@ export function mountConsolidationRoutes(app: express.Application): void {
   });
 
   // GET /api/concepts/search — Search concepts by text
-  app.get("/api/concepts/search", consolidationRateLimiter, authMiddleware, async (req: express.Request, res: express.Response) => {
+  app.get("/api/concepts/search", consolidationRateLimiter, authMiddleware, rejectArrayParams("query", "project", "limit"), async (req: express.Request, res: express.Response) => {
     try {
-      if (Array.isArray(req.query.query) || Array.isArray(req.query.project) || Array.isArray(req.query.limit)) {
-        res.status(400).json({ error: "Bad Request: array parameters not supported" });
-        return;
-      }
-
       const query = String(req.query.query || "");
       const project = req.query.project as string | undefined;
       const limit = Math.min(Number(req.query.limit) || 10, 50);
--- a/src/presentation/consolidationRoutes.ts
+++ b/src/presentation/consolidationRoutes.ts
@@ -34,6 +34,11 @@ export function mountConsolidationRoutes(app: express.Application): void {
   // GET /api/concepts/search — Search concepts by text
   app.get("/api/concepts/search", consolidationRateLimiter, authMiddleware, async (req: express.Request, res: express.Response) => {
     try {
+      if (Array.isArray(req.query.query) || Array.isArray(req.query.project) || Array.isArray(req.query.limit)) {
+        res.status(400).json({ error: "Bad Request: array parameters not supported" });
+        return;
+      }
+
       const query = String(req.query.query || "");
       const project = req.query.project as string | undefined;
       const limit = Math.min(Number(req.query.limit) || 10, 50);
--- a/src/presentation/dreamingRoutes.ts
+++ b/src/presentation/dreamingRoutes.ts
@@ -48,7 +48,8 @@ export function registerDreamingRoutes(app: express.Application): void {
       const auth = await checkAuth(apiKey, bearerToken);
 
       if (Array.isArray(req.query.id)) {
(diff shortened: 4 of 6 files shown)


giauphan:
cc @jules

# Code Review

The PR adds input validation guards against array query parameters across multiple route handlers. This is a good defensive measure since Express parses `?key=1&key=2` as an array, which would bypass `as string` type assertions and could cause unexpected behavior.

## Issues

**File: `src/presentation/dreamingRoutes.ts` -> Line 51:** Inconsistent return pattern. This handler uses `return res.status(400).json(...)` while other handlers in this PR (e.g., `heartbeatRoutes.ts`, `genomeRoutes.ts`) use the `res.status(400).json(...); return;` pattern. Mixing both styles works, but inconsistency hurts maintainability. Pick one pattern and stick with it.

**File: `src/presentation/dreamingRoutes.ts` -> Line 125:** Same inconsistency — uses `return res.status(400).json(...)`.

**File: `src/presentation/httpServer.ts` -> Line 278:** Same inconsistency — uses `return res.status(400).json(...)`.

**File: `src/presentation/httpServer.ts` -> Line 452:** Same inconsistency — uses `return res.status(400).json(...)`.

**File: `src/presentation/httpServer.ts` -> Line 1111:** Inconsistent error response format. This endpoint returns `res.status(400).send("Bad Request: ...")` (plain text) while every other guard in this PR returns `res.status(400).json({ error: "Bad Request: ..." })` (JSON). This will behave differently for clients expecting a consistent content type.

**File: `src/presentation/dreamingRoutes.ts` -> Line 123:** The `limit` parameter is also used in this handler (line ~130, `limitRaw`), but it's not included in the array check, unlike the similar guard in `genomeRoutes.ts` which does check `limit`. This is an inconsistency that leaves a gap.

**File: `src/presentation/httpServer.ts` -> Line 736:** The `projectDir` and `project` parameters are validated, but this endpoint also reads other query params further down (depending on context). Worth verifying no other query params are consumed. Minor, but worth a glance.

## General Suggestion

This is a cross-cutting concern. Rather than scattering identical guard blocks across every route, consider a small reusable middleware or helper:

```typescript
function rejectArrayParams(...keys: string[]) {
  return (req: express.Request, res: express.Response, next: express.NextFunction) => {
    for (const key of keys) {
      if (Array.isArray(req.query[key])) {
        res.status(400).json({ error: `Bad Request: array parameters not supported for ${key}` });
        return;
      }
    }
    next();
  };
}

// Usage:
app.get("/api/concepts/search", rejectArrayParams("query", "project", "limit"), authMiddleware, async (req, res) => { ... });
```

This eliminates ~60 lines of duplication, ensures consistent response format, and makes it harder to forget a parameter.

## Verdict

The intent is correct and addresses a real issue (query parameter pollution). The main concerns are: (1) one endpoint returns plain text instead of JSON, (2) inconsistent `return` style, and (3) significant duplication that should be extracted into a helper. None are blocking, but the plain text response inconsistency in `/messages` should be fixed before merging.

---

giauphan:
cc @jules

## Code Review

The PR adds input validation to reject array query parameters across multiple route handlers. This is a good defensive measure against query string pollution (e.g., `?id=1&id=2` being parsed as an array), which could cause unexpected behavior with the `as string` type assertions used throughout.

### Issues

**File: `src/presentation/httpServer.ts` -> Line 1115: Missing auth middleware / rate limiter check ordering**
The `/messages` endpoint adds the array check *before* any authentication, which is fine, but notably this is the only endpoint in the diff that lacks `authMiddleware` and rate limiting on the route definition itself. This is pre-existing, not introduced by this PR, but worth noting.

**File: Multiple files -> All changed lines: Repetitive validation pattern (Code Quality)**
The same `Array.isArray` check-and-respond pattern is duplicated verbatim across ~8 endpoints. Consider extracting a small middleware or utility:

```typescript
function rejectArrayParams(...paramNames: string[]) {
  return (req: express.Request, res: express.Response, next: express.NextFunction) => {
    if (paramNames.some(p => Array.isArray(req.query[p]))) {
      res.status(400).json({ error: "Bad Request: array parameters not supported" });
      return;
    }
    next();
  };
}

// Usage:
app.get("/api/analysis", authMiddleware, localRateLimiter, rejectArrayParams("projectDir", "project"), async (req, res) => { ... });
```

This eliminates duplication and makes it harder to forget adding the check to new endpoints.

**File: `src/presentation/httpServer.ts` -> Lines 278, 453, 497, 739: Incomplete coverage**
These endpoints likely have additional query parameters beyond the ones checked. If a new query param is added to a route later, a developer must remember to add it to the array check too. The middleware approach above (or a blanket "reject all array query values" middleware applied globally) would be more robust.

**File: `src/presentation/dreamingRoutes.ts` -> Line 51: Inconsistent error message**
The error message is `"Bad Request: array parameters not supported for id"` while every other endpoint uses `"Bad Request: array parameters not supported"`. Minor inconsistency but could confuse API consumers or log parsers.

### Verdict

The intent is correct and addresses a real class of bugs (query param pollution bypassing `as string` casts). The main recommendation is to **DRY this up into a reusable middleware** to improve maintainability and ensure consistent coverage across all current and future endpoints. The inconsistent error message in `dreamingRoutes.ts` should also be normalized.

---

giauphan:
cc @jules

## Code Review

The PR adds HTTP Parameter Pollution (HPP) protection by introducing a `rejectArrayParams` middleware that validates query parameters aren't arrays. Overall the approach is sound and the implementation is clean. A few issues worth flagging:

### Issues

**File: `src/middleware/validation.ts` -> Line 8: Missing `return` type annotation**
The returned middleware function lacks an explicit return type. Minor, but given the project rules emphasize TypeScript best practices:
```ts
return (req: express.Request, res: express.Response, next: express.NextFunction): void => {
```

**File: `src/presentation/a2a/heartbeatRoutes.ts` -> Line 17: Using GET for a registration/mutation endpoint**
This is a pre-existing issue, not introduced by this PR, but worth noting: `GET /a2a/register` performs a state-mutating operation (registering an agent). This should ideally be `POST`. The HPP protection is correctly applied here regardless.

**File: `src/presentation/dreamingRoutes.ts` -> Line 46: Missing `authMiddleware` on delete endpoint**
The `/api/dreams/delete` route uses manual `extractAuth`/`checkAuth` instead of `authMiddleware`, which means `rejectArrayParams` runs before any authentication. This is fine security-wise (rejecting bad input early is good), but the inconsistency with other routes (which use `authMiddleware`) is a pre-existing concern. Not introduced by this PR.

**File: `src/presentation/httpServer.ts` -> Line 1095: `rejectArrayParams("sessionId")` on a POST route**
The `/messages` endpoint is a `POST`. Query params on POST bodies are unusual — verify that `sessionId` is actually passed as a query string parameter and not in the request body. If it's in the body, `req.query` won't have it and this middleware becomes a no-op. Looking at line 1095 (`req.query.sessionId as string`), it does read from query, so this is correct — but it's worth confirming no clients send it in the body.

**File: `src/middleware/validation.ts` -> Line 9: Consider covering nested object injection**
`Array.isArray` catches `?param=a&param=b` (arrays), but Express's `qs` parser also allows object injection like `?param[key]=value`, which would pass this check but still cause issues when cast with `as string` (resulting in `"[object Object]"`). Consider a stricter check:
```ts
if (paramNames.some(p => typeof req.query[p] !== 'undefined' && typeof req.query[p] !== 'string')) {
```
This would reject both arrays and objects.

### What looks good

- The middleware pattern is clean, reusable, and composable
- Correct placement in the middleware chain (after auth/rate-limiting where applicable)
- Consistent application across all routes that use `as string` casts on query params
- The 400 response with a clear error message is appropriate
- No `any` types used

### Summary

The most actionable issue is the **object injection gap** in the validation check — the current implementation only catches array pollution but not object pollution from Express's query string parser, and both produce incorrect behavior when cast with `as string`. Everything else is solid.

---

giauphan:
cc @jules

## Code Review

The PR adds HTTP Parameter Pollution (HPP) protection by introducing a `rejectArrayParams` middleware that validates query parameters are strings, not arrays. This is a solid security improvement. Here are the findings:

### Issues

**File: `src/middleware/validation.ts` -> Line 9: Incomplete validation — doesn't handle empty `paramNames`**

The middleware works correctly when `paramNames` is populated, but calling `rejectArrayParams()` with no arguments silently becomes a no-op (since `some` on an empty array returns `false`). This isn't a bug per se, but documenting or guarding against it would improve clarity.

**File: `src/middleware/validation.ts` -> Line 9: Consider also checking for `null` values**

Express's `req.query` can produce values of type `string`, `string[]`, `ParsedQs`, or `ParsedQs[]`. The current check (`typeof !== 'undefined' && typeof !== 'string'`) correctly catches arrays and nested objects, but it's worth noting this intentionally allows `undefined` (parameter not present) to pass through. This is correct behavior — just confirming the design is intentional.

**File: `src/presentation/dreamingRoutes.ts` -> Line 46: Missing `authMiddleware` on delete route**

The `/api/dreams/delete` endpoint uses manual auth extraction (`extractAuth` + `checkAuth`) instead of `authMiddleware`, which is inconsistent with the other routes (e.g., `/api/dreams/query` uses `authMiddleware`). This is a pre-existing issue, not introduced by this PR, but worth flagging since you're already touching this route.

**File: `src/presentation/a2a/heartbeatRoutes.ts` -> Line 20: Registration via GET is semantically wrong**

`GET /a2a/register` is a state-changing operation (registering an agent). This should be `POST`. Pre-existing issue, not from this PR.

**File: `src/presentation/httpServer.ts` -> Line 1095: `POST /messages` — `sessionId` is a query param on a POST**

The `rejectArrayParams("sessionId")` is correctly applied, but note that `sessionId` is cast with `as string` on line 1096 without any falsy check before the `transports.get(sessionId)` call, which will just return `undefined` for the ternary — so it's safe but could be cleaner.

### Looks Good

- The middleware itself is clean, minimal, and correctly typed (returns `void`, uses `next()`).
- Consistent application across all routes that use `req.query.X as string` casting — this is exactly where HPP would be exploitable.
- No `any` types used.
- The parameter lists in each `rejectArrayParams(...)` call match the actual query params consumed in each handler — well done on the thoroughness.

### Minor Nit

**File: `src/presentation/dreamingRoutes.ts` -> Lines 49, 120: Extraneous blank lines added**

Two whitespace-only changes (added blank lines after existing code) that aren't meaningful. Very minor, but they add noise to the diff.

### Summary

**Good PR.** The core change is correct, well-scoped, and addresses a real security concern (HPP via duplicate query params being parsed as arrays, bypassing `as string` type assertions). No new bugs introduced. Ship it.

---

giauphan:
cc @jules

# Code Review

The PR adds HTTP Parameter Pollution (HPP) protection via a `rejectArrayParams` middleware. Overall this is a solid security improvement. A few issues worth flagging:

## Issues

**File: `src/middleware/validation.ts` -> Line 14-16: Incomplete validation — misses nested object injection**

Express's `qs` parser can produce not just arrays but also nested objects from query strings (e.g., `?query[foo]=bar`). The current check `typeof req.query[p] !== 'string'` will catch arrays, but it's worth being explicit that this also covers objects. This actually works correctly since `typeof {}` is `'object'`, not `'string'` — so this is fine. No action needed on second thought.

**File: `src/middleware/validation.ts` -> Line 15: Minor — error message leaks no useful detail for debugging**

The 400 response doesn't indicate *which* parameter was invalid. Consider:
```typescript
const offending = paramNames.filter(p => typeof req.query[p] !== 'undefined' && typeof req.query[p] !== 'string');
res.status(400).json({ error: `Bad Request: invalid parameter type for [${offending.join(', ')}], string expected` });
```
This helps API consumers fix their requests without exposing internals.

**File: `src/presentation/a2a/heartbeatRoutes.ts` -> Line 14: GET with side effects**

Not introduced by this PR, but `GET /a2a/register` is a registration endpoint (write operation) using GET. This is an HTTP semantics violation — registration should be POST. The HPP middleware is correctly applied here, but the underlying route verb is wrong.

**File: `src/presentation/dreamingRoutes.ts` -> Line 46: Missing `authMiddleware` on DELETE route**

The `DELETE /api/dreams/delete` route has `rejectArrayParams("id")` but no `authMiddleware` in the middleware chain — authentication is done manually inside the handler via `extractAuth`/`checkAuth`. This is inconsistent with the other routes (e.g., line 116 uses `authMiddleware`). Not introduced by this PR, but worth noting.

**File: `src/presentation/httpServer.ts` -> Line 1095: POST `/messages` — query param `sessionId` on a POST**

The `rejectArrayParams("sessionId")` is correctly applied, but `sessionId` is read from query params on a POST route. This is a design smell — session identifiers in query strings can leak via logs, referrer headers, and browser history. Consider moving it to a request header or POST body.

**File: `src/middleware/validation.ts` -> Line 10-12: Warn log on empty paramNames is noisy**

If someone accidentally applies `rejectArrayParams()` with no args, the warning is logged on *every request*. This should either be a startup-time error (throw immediately when the middleware is created) or at minimum rate-limited:

```typescript
export function rejectArrayParams(...paramNames: string[]) {
  if (paramNames.length === 0) {
    logger.warn("[rejectArrayParams] Middleware created with empty paramNames array.");
  }
  return (req, res, next) => {
    if (paramNames.length === 0) return next();
    // ...
  };
}
```

This moves the warning to middleware creation time (once) rather than per-request.

## What looks good

- The middleware is clean, focused, and correctly typed with `void` return.
- Consistent application across all routes that use `req.query` with `as string` casts.
- Good use of early return pattern.
- The middleware correctly handles the `undefined` case (parameter not present is fine).

**Verdict**: Good security hardening. The per-request warning log for empty params is the only issue I'd ask to fix before merging. The rest are suggestions.

---

giauphan:
cc @jules

# Code Review

The changes look good overall — this is a solid, consistent approach to preventing HTTP Parameter Pollution (HPP) across multiple endpoints. A few issues worth flagging:

## Issues

**File: `src/middleware/validation.ts` -> Line 16:** The check `typeof req.query[p] !== 'undefined'` will pass for `null` values. Since Express can parse query params in unexpected ways, consider using `p in req.query` or checking `req.query[p] !== undefined` (strict equality) instead of `typeof`. More importantly, this only catches arrays (which Express represents as `string[]`) but `typeof` an array is `'object'`, so the logic works — but it would be clearer and more defensive to explicitly check `Array.isArray(req.query[p])`:

```typescript
const offending = paramNames.filter(p => Array.isArray(req.query[p]));
```

This is more readable and precisely targets the HPP vector you're defending against, rather than also rejecting any future parsed object types that might be legitimate.

**File: `src/middleware/validation.ts` -> Line 9-11:** The `logger.warn` at middleware creation time (not at request time) is a good idea, but this runs once during app startup. If this is intentional as a dev-time guard, consider making it a hard error (`throw new Error(...)`) instead — a no-op middleware is a silent bug that's easy to miss.

**File: `src/presentation/httpServer.ts` -> Line 1095:** `POST /messages` uses `rejectArrayParams("sessionId")` but `sessionId` comes from `req.query` on a POST endpoint. This is correct for query-string pollution, but note that if `sessionId` could also come from the request body in the future, this middleware won't cover that vector.

**File: `src/presentation/a2a/heartbeatRoutes.ts` -> Line 20:** The `GET /a2a/register` endpoint — using GET for a registration/mutation operation is an existing concern (not introduced by this PR), but worth noting since you're already touching this line. GET requests with side effects are vulnerable to CSRF via simple navigation/img tags.

**File: `src/presentation/dreamingRoutes.ts` -> Line 46:** `DELETE /api/dreams/delete` applies `rejectArrayParams("id")` but does **not** have `authMiddleware` in the middleware chain (authentication is done manually inside the handler via `extractAuth`). This is inconsistent with other endpoints and the validation middleware will run before any auth check — not a bug per se, but an attacker can probe parameter validation errors without authenticating.

## Looks Good

- Consistent application across all relevant endpoints
- The middleware is pure, stateless, and composable
- Error messages are informative without leaking internals
- No `any` types used

**Verdict:** Solid defensive hardening. Address the `Array.isArray` readability improvement and consider the startup-warning-vs-throw tradeoff. The rest are minor observations.

---

giauphan:
cc @jules

## Code Review

The PR adds an HTTP Parameter Pollution (HPP) mitigation middleware that rejects array query parameters on specific routes. Overall, this is a solid security improvement. A few issues worth flagging:

### Issues

**File: `src/middleware/validation.ts` -> Line 13: Incomplete HPP protection — only checks for arrays, not nested objects**

Express's `qs` parser (enabled by default) can also parse `?param[key]=value` into objects. The middleware only checks `Array.isArray()`, so an attacker could still send `?query[toString]=evil` or similar object pollution payloads. Consider also rejecting non-string values:

```typescript
const offending = paramNames.filter(p => {
  const val = req.query[p];
  return val !== undefined && typeof val !== 'string';
});
```

**File: `src/presentation/a2a/heartbeatRoutes.ts` -> Line 17: Using GET with side effects for registration**

`GET /a2a/register` appears to register an agent (a write operation). This is a pre-existing issue, not introduced by this PR, but the HPP middleware on a GET mutation endpoint is a band-aid — the route should likely be POST/PUT. Not blocking, just noting.

**File: `src/presentation/httpServer.ts` -> Line 1095: `rejectArrayParams("sessionId")` on POST `/messages` — validates query param but body is unchecked**

For POST requests, `sessionId` comes from the query string which is fine, but it's worth confirming that no other routes accept the same parameter names in both query and body, as the middleware only guards `req.query`.

**File: `src/middleware/validation.ts` -> Line 7: Throwing at initialization time is good, but consider the error message at runtime**

The 400 response leaks internal parameter names to the client. This is minor since these are already visible in the URL, but for consistency with security hardening you might want a generic message like `"Bad Request: malformed query parameters"` instead of enumerating the offending keys.

### Maintenance Concern

The approach requires manually listing parameter names on every route. If a developer adds a new query parameter but forgets to add it to the `rejectArrayParams()` call, it remains unprotected. Consider either:
- A global middleware that rejects **all** array-typed query values (simpler, less error-prone), or
- At minimum, adding a comment/lint rule to remind developers to update the list.

### What Looks Good

- Clean, focused middleware with a single responsibility
- Proper early return with `void` annotation to satisfy Express typing
- Initialization-time validation (empty param list throws)
- Applied consistently across search/query endpoints that cast `req.query` values with `as string`

**Verdict**: The PR meaningfully reduces HPP risk. The main actionable item is extending the check beyond `Array.isArray` to also reject parsed objects (`typeof val !== 'string'`). The rest are minor suggestions.

---

giauphan:
cc @jules

# Code Review

The PR adds HTTP Parameter Pollution (HPP) protection by introducing a `rejectArrayParams` middleware that validates query parameters are not arrays. Overall this is a solid security improvement. A few issues worth flagging:

## Issues

**File: `src/middleware/validation.ts` -> Line 13-16: Incomplete HPP protection — nested object parameters not rejected**

The check `typeof val !== 'string'` catches arrays but also catches parsed objects (e.g., `?query[foo]=bar`). This is fine — it's actually broader protection. However, it's worth noting that `undefined` values pass through, meaning parameters not present in the request are silently allowed. This is correct behavior, but the JSDoc should clarify that only *present* parameters are validated.

**File: `src/middleware/validation.ts` -> Line 17: Missing detail in error response**

The error message `"Bad Request: malformed query parameters"` doesn't indicate *which* parameters are problematic. Consider including the offending parameter names to aid debugging:

```typescript
res.status(400).json({
  error: `Bad Request: malformed query parameters: ${offending.join(", ")}`
});
```

**File: `src/presentation/a2a/heartbeatRoutes.ts` -> Line 14: Register endpoint uses GET with side effects**

Not introduced by this PR, but worth noting: `GET /a2a/register` is a registration (write) operation exposed as a GET endpoint. GET requests should be idempotent and safe. This is an existing design issue, not a regression.

**File: `src/presentation/dreamingRoutes.ts` -> Line 46: Inconsistent auth pattern on DELETE endpoint**

The `DELETE /api/dreams/delete` route uses `rejectArrayParams("id")` but doesn't use `authMiddleware` in the middleware chain (it does manual auth extraction inside the handler). This is a pre-existing inconsistency, but adding `rejectArrayParams` here without `authMiddleware` means an unauthenticated request with a valid single `id` param will pass validation and reach the handler before auth is checked. Not a regression, but the manual auth pattern is fragile.

**File: `src/presentation/httpServer.ts` -> Line 1095: POST `/messages` validates query param `sessionId`**

This is correct — `sessionId` comes from the query string even on a POST. Good catch applying HPP protection here.

## Minor Observations

- The middleware correctly throws at initialization time (not at request time) if no param names are provided — this is a good fail-fast pattern.
- All usages are consistent: the middleware is placed in the chain before the route handler.
- The parameter lists in each route appear complete — they match the `req.query.*` accesses in the handlers.

## Verdict

**Looks good.** This is a clean, focused security hardening PR. The middleware is well-structured, the application is consistent across routes, and it correctly prevents HPP attacks. The suggestions above are minor improvements, not blockers.

---

giauphan:
cc @jules

# Code Review

The changes look good overall — this is a solid, systematic approach to preventing HTTP Parameter Pollution (HPP) across the application. A few issues worth flagging:

## Issues

**File: `src/middleware/validation.ts` -> Line 1: Unused import style**
The import pulls in the entire `express` module but only uses `Request`, `Response`, and `NextFunction` types. This is minor but for consistency with TypeScript best practices, consider using type-only imports:
```typescript
import type { Request, Response, NextFunction } from "express";
```

**File: `src/presentation/a2a/heartbeatRoutes.ts` -> Line 20: Incomplete protection — `as string` casts remain after validation**
The middleware correctly rejects array params, but the handlers still use `as string` type assertions (e.g., `req.query.agent_url as string`). This is fine *after* the middleware runs, but the pattern is fragile — if someone later adds a route or removes the middleware, the unsound cast silently passes. Consider adding a small typed helper (e.g., `getStringParam(req, "agent_url")`) that returns `string | undefined` without a cast, to pair with the middleware for defense in depth.

**File: `src/presentation/httpServer.ts` -> Line 277: Middleware order inconsistency**
On the `/api/projects` DELETE route, the order is `authMiddleware, localRateLimiter, rejectArrayParams(...)`. In most other routes (e.g., `consolidationRoutes.ts` line 33), validation comes right after the rate limiter and auth. This specific route places `rejectArrayParams` *after* both auth and rate limiting, meaning an invalid request still consumes a rate-limit token and triggers auth before being rejected. For consistency and efficiency, consider placing `rejectArrayParams` earlier in the chain (e.g., right after the rate limiter, before auth).

**File: `src/presentation/httpServer.ts` -> Line 1095: Missing auth/rate-limiting on `/messages`**
The `/messages` POST route only has `rejectArrayParams("sessionId")` — no `authMiddleware` or rate limiter. This predates the PR, but since you're touching this line, it's worth noting. The `sessionId` is taken directly from the query string and used as a transport lookup key with no authentication. If this is intentional (e.g., MCP transport design), a comment explaining why would help.

**File: `src/presentation/dreamingRoutes.ts` -> Line 46: DELETE with query params**
The `DELETE /api/dreams/delete` route reads `id` from `req.query`. Conventionally, DELETE endpoints use path params (`/api/dreams/:id`) or request body. Using query params for DELETE is unusual and some proxies/caches may strip query strings from DELETE requests. Not introduced by this PR, but worth a follow-up.

**File: `src/middleware/validation.ts` -> Line 9: Constructor-time validation is good, but consider edge case**
The `throw new Error` on empty `paramNames` happens at middleware registration (app startup), which is the right time to fail. Good pattern — no issue here.

## What looks good

- The middleware is clean, focused, and well-documented with a clear JSDoc comment.
- The approach of validating only present params (letting `undefined` pass through) is correct — it separates "missing" from "malformed."
- Systematic application across all routes that use `as string` casts on query params.
- The error response is informative, listing all offending parameters at once rather than failing on the first one.

**Verdict: Approve with minor suggestions.** The core change is correct and meaningfully improves security. The suggestions above are mostly about consistency and defense in depth.

---

PR #303, opened 2026-09-12
fix: tolerate Firestore quota exhaustion

giauphan:
## Summary

Prevent Firestore quota exhaustion from repeatedly breaking authentication and project settings endpoints.

## Motivation

When Firestore returned `8 RESOURCE_EXHAUSTED: Quota exceeded`, super-admin requests retried the auth lookup on every request and settings routes returned HTTP 500 despite local settings being available.

## Changes

- Cache successful super-admin identity resolution.
- Cache degraded admin fallback identity for five minutes during Firestore outages.
- Treat Firestore settings reads and writes as non-fatal backup operations.
- Add regression coverage for auth caching, recovery after fallback TTL, and settings API quota failures.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactor (no functional change)
- [x] Test improvement

## Checklist

- [x] `pnpm run build` passes
- [x] `pnpm test` passes
- [ ] Dashboard builds (`cd dashboard && pnpm run build`) if dashboard touched
- [x] Tests added for new behavior
- [ ] Documentation updated (`README.md`, `docs/`, `CHANGELOG.md`) if user-visible change
- [x] No secrets, credentials, or personal data in diff
- [x] No `any` types — used `unknown` + type guards
- [x] Commit messages follow Conventional Commits (`feat:`, `fix:`, `docs:`, etc.)

## Test plan

- `pnpm run typecheck`
- `pnpm run build`
- `pnpm test` — 266 passing
- `git diff --check`

## Screenshots / output

Not applicable.

## Security impact

Authentication remains fail-closed for invalid non-admin API keys. During Firestore unavailability, only the configured super-admin key receives the existing admin fallback identity; no credentials or key material are logged.


giauphan (commented):
cc @jules

## Code Review

The changes look good overall — they improve resilience by gracefully degrading when Firestore is unavailable instead of failing requests, and they add caching to the super-admin fallback path. A few issues worth flagging:

### Issues

**File: `src/repositories.ts` → Line ~156-158: Cache check returns entry with potentially wrong semantics**

The cached entry for the super-admin path could be a previously cached Firestore-resolved result (with a real `uid`) *or* a fallback entry (with `uid: 'admin'`). If Firestore was temporarily down, the fallback gets cached, and subsequent requests within the TTL window will return the degraded fallback even after Firestore recovers. This is probably acceptable given the short TTL, but worth documenting with a comment.

**File: `src/repositories.ts` → Line ~170: `expires` field now has dual meaning**

Previously, the admin fallback returned `expires: Infinity`, meaning the auth data itself never expires. Now `expires` is set to `Date.now() + adminFallbackTtl`, which is used *both* as the cache eviction key (`cached.expires > Date.now()`) and as the semantic "auth expiry" returned to callers. If any downstream consumer interprets `expires` as "this auth session is valid until X", the super-admin will now appear to expire after 5 minutes rather than being perpetual. This could cause subtle auth failures.

Consider separating cache TTL from the auth data's own expiry:

```typescript
const fallback: AuthData = {
  tier: 'enterprise',
  uid: 'admin',
  keyId: 'admin',
  expires: Infinity, // auth never expires for admin
};
// Store with a separate cache expiry
this.authCache.set(apiKey, { ...fallback, expires: Date.now() + this.adminFallbackTtl });
```

Wait — that would also break things since it's the same object shape. The real fix is to use a separate cache wrapper type (e.g., `{ data: AuthData, cachedUntil: number }`) instead of overloading `AuthData.expires` for cache eviction. This is the same pattern used in the "Check Local RAM Cache" block below — verify it's consistent.

**File: `src/presentation/httpServer.ts` → Line ~562: Swallowed error may hide real issues**

Downgrading to `warn` and continuing is correct for "Firestore unavailable" scenarios, but this catch block will also swallow unexpected errors (e.g., malformed data, permission bugs). Consider at minimum logging at `error` level for non-network/non-availability errors, or narrowing the catch:

```typescript
const isAvailabilityError = e instanceof Error && 
  (e.message.includes('UNAVAILABLE') || e.message.includes('NOT_FOUND'));
logger[isAvailabilityError ? 'warn' : 'error']("[Settings API] Firestore fallback issue:", ...);
```

### Minor

**File: `src/repositories.ts` → Line ~167: Empty catch block**
The existing empty catch with just a comment (`// Fallback gracefully...`) was already there, but now that you're touching this code, consider logging at `debug`/`warn` level so issues aren't completely invisible.

### Summary

The resilience improvements are the right direction. The main concern is the **`expires` field dual-purpose issue** in the admin fallback — this could cause downstream auth behavior changes and should be addressed before merging.

giauphan (commented):
cc @jules

This workspace contains a different codebase (the review bot itself), so the PR's `src/presentation/httpServer.ts` / `src/repositories.ts` aren't available to verify against — review is based on the diff alone.

**Issues**

1. File: src/repositories.ts -> Line 169: Super-admin success path is cached with `cachedUntil: Infinity`, and the new cache check now short-circuits before `verifyKey()`. Previously the admin branch re-verified against Firestore on every request, so deleting/rotating the key in Firestore took effect immediately; now a revoked admin key stays valid until process restart. Use a finite TTL (like the fallback) or add explicit invalidation.

2. File: src/repositories.ts -> Line 174: `logger.debug(...)` is called in the catch, but no `logger` import appears anywhere in the diff. If it isn't already in scope in this module, this is a compile error — and a runtime `ReferenceError` inside an error handler if not typechecked.

3. File: src/presentation/httpServer.ts -> Line 561: The catch no longer returns a response, so every error is swallowed — including failures of the primary local-file read inside the same `try`, not just the Firestore mirror. Verify the code after the catch always sends a response (otherwise requests hang) and that missing settings on the success path is acceptable; as written, genuine bugs (TypeError, serialization errors) silently degrade to empty settings too.

4. File: src/presentation/httpServer.ts -> Line 563: `e.message.includes('UNAVAILABLE')` is brittle — Firestore/gRPC errors expose a structured `code` (e.g. `err.code`), and message text varies across SDK versions. Also `'permission-denied'` / `'DEADLINE_EXCEEDED'` won't match and get logged with the misleading "Firestore fallback unavailable" message at error level.

**Looks good**

- Splitting cache expiry (`cachedUntil`) from `AuthData.expires` fixes a real latent bug: the old code overwrote the key's real expiry with the cache TTL.
- The update-settings hunk (Line 634): treating the Firestore mirror as best-effort with a warn log is consistent with its "backup mirror" role — fine.
- No `any` types; `catch (err: unknown)` with instanceof narrowing is correct.

src/repositories.ts
```diff
@@ -150,23 +158,37 @@ export class AuthenticateUserUseCase {
...
           return firestoreData;
         }
-      } catch (err) {
+      } catch (err: unknown) {
         // Fallback gracefully if Firestore is unconfigured or unavailable
+        logger.debug(`[AuthenticateUserUseCase] Super admin lookup failed, using fallback: ${err instanceof Error ? err.message : String(err)}`);
       }
-      return { tier: 'enterprise', uid: 'admin', keyId: 'admin', expires: Infinity };
+      // Cache the fallback briefly so a degraded store is not hit on every request
+      // We don't overwrite `expires` semantics of the original payload so downstream systems don't treat it differently.
```
giauphan, line 268: src/repositories.ts -> Line 169: Super-admin success path is cached with `cachedUntil: Infinity`, and the new cache check now short-circuits before `verifyKey()`. Previously the admin branch re-verified against Firestore on every request, so deleting/rotating the key in Firestore took effect immediately; now a revoked admin key stays valid until process restart. Use a finite TTL (like the fallback) or add explicit invalidation.

src/repositories.ts
```diff
@@ -150,23 +158,37 @@ export class AuthenticateUserUseCase {
...
         if (firestoreData) {
           firestoreData.expires = Infinity;
-          this.authCache.set(apiKey, firestoreData);
+          this.authCache.set(apiKey, { data: firestoreData, cachedUntil: Infinity });
           return firestoreData;
         }
-      } catch (err) {
+      } catch (err: unknown) {
         // Fallback gracefully if Firestore is unconfigured or unavailable
+        logger.debug(`[AuthenticateUserUseCase] Super admin lookup failed, using fallback: ${err instanceof Error ? err.message : String(err)}`);
```
giauphan, line 264: src/repositories.ts -> Line 174: `logger.debug(...)` is called in the catch, but no `logger` import appears anywhere in the diff. If it isn't already in scope in this module, this is a compile error — and a runtime `ReferenceError` inside an error handler if not typechecked.

src/presentation/httpServer.ts
```diff
@@ -559,8 +559,10 @@ app.get("/api/projects/settings", authMiddleware, localRateLimiter, rejectArrayP
           }
         }
       } catch (e: unknown) {
```
giauphan, line 586: src/presentation/httpServer.ts -> Line 561: The catch no longer returns a response, so every error is swallowed — including failures of the primary local-file read inside the same `try`, not just the Firestore mirror. Verify the code after the catch always sends a response (otherwise requests hang) and that missing settings on the success path is acceptable; as written, genuine bugs (TypeError, serialization errors) silently degrade to empty settings too.

src/presentation/httpServer.ts
```diff
@@ -559,8 +559,10 @@ app.get("/api/projects/settings", authMiddleware, localRateLimiter, rejectArrayP
...
-        logger.error("[Settings API] Error reading Firestore fallback:", e);
-        return res.status(500).json({ error: `Failed to fetch settings: ${e instanceof Error ? e.message : String(e)}` });
+        // Firestore is only a fallback mirror of the local file; degrade gracefully instead of failing the request
+        const isAvailabilityError = e instanceof Error &&
```
giauphan, line 588: src/presentation/httpServer.ts -> Line 563: `e.message.includes('UNAVAILABLE')` is brittle — Firestore/gRPC errors expose a structured `code` (e.g. `err.code`), and message text varies across SDK versions. Also `'permission-denied'` / `'DEADLINE_EXCEEDED'` won't match and get logged with the misleading "Firestore fallback unavailable" message at error level.

PR #304, opened 2026-09-12
feat: auto-extract genome genes and add verified_solution_search tool

giauphan:
## Summary

Add automatic genome extraction to the dream session ingestion pipeline and introduce the `verified_solution_search` MCP tool.

## Motivation

Automating genome extraction links dream learnings to reusable patterns automatically. The new MCP tool enables AI agents to query verified patterns alongside execution flow inside the current analyzed project.

## Changes

- Extract confident genome genes from saved dream memories inside `runSessionIngestion`.
- Track `genesSaved` telemetry metrics.
- Add `verified_solution_search` MCP tool with cross-checks for genome genes and real AST flow nodes.
- Prevent emitting empty `# Context from Previous Sessions` sections when only genome vectors are present.
- Ensure `GenomeService.searchGenes` queries execute within the authenticated tenant boundary via `authStorage.run`.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactor (no functional change)
- [x] Test improvement

## Checklist

- [x] `pnpm run build` passes
- [x] `pnpm test` passes
- [ ] Dashboard builds (`cd dashboard && pnpm run build`) if dashboard touched
- [x] Tests added for new behavior
- [ ] Documentation updated (`README.md`, `docs/`, `CHANGELOG.md`) if user-visible change
- [x] No secrets, credentials, or personal data in diff
- [x] No `any` types — used `unknown` + type guards
- [x] Commit messages follow Conventional Commits (`feat:`, `fix:`, `docs:`, etc.)

## Test plan

- `pnpm run typecheck`
- `pnpm run build`
- `pnpm test` — 262 passing
- `git diff --check`

### DIFF
--- a/src/services/llmService.ts
+++ b/src/services/llmService.ts
@@ -173,7 +173,10 @@ export async function loadContextAtSessionStart(
       return "";
     }
 
-    const parts: string[] = ["\n# 🧠 Context from Previous Sessions\n"];
+    const parts: string[] = [];
+    if (cleanDreams.length > 0) {
+      parts.push("\n# 🧠 Context from Previous Sessions\n");
+    }
     for (const dream of cleanDreams) {
       const d = dream as Record<string, unknown>;
       const memoryType = String(d.memoryType ?? d.memory_type ?? d.MEMORY_TYPE ?? "DREAM");
--- a/src/services/llmService.ts
+++ b/src/services/llmService.ts
@@ -1,5 +1,6 @@
 import { logger } from "../utils/logger.js";
 import { DreamingService } from "./dreamingService.js";
+import { GenomeService } from "./genomeService.js";
 import { checkNoiseBlocklist } from "./noiseBlocklist.js";
 import { countMatching } from "../utils/array.js";
 
@@ -130,14 +131,33 @@ export async function loadContextAtSessionStart(
       10
     );
 
-    if (!dreams || dreams.length === 0) {
+    // Strict genome auto-load: relevance-gated genes + immune context for the
+    // explicit task only. Never fires without a non-empty task.
+    const strictTask = String(task || "").trim();
+    let genes: Awaited<ReturnType<typeof GenomeService.searchGenes>> = [];
+    let immuneContext = "";
+    if (strictTask.length > 0) {
+      try {
+        genes = await GenomeService.searchGenes(strictTask, { project, limit: 5 });
+        genes = genes.filter((g) => g.confidence >= 0.5 && g.score >= 0.3);
+      } catch (err) {
+        logger.error(`[Memory Loading] Genome auto-load failed: ${err instanceof Error ? err.message : String(err)}`);
+      }
+      try {
+        immuneContext = (await GenomeService.buildImmuneContext(strictTask, project)) || "";
+      } catch (err) {
+        logger.error(`[Memory Loading] Immune context failed: ${err instanceof Error ? err.message : String(err)}`);
+      }
+    }
+
+    if ((!dreams || dreams.length === 0) && genes.length === 0 && !immuneContext) {
       return "";
     }
 
     const cleanDreams: Array<{ memoryType: string; content: string; importance: number }> = [];
-    const dreamsLen = dreams.length;
+    const dreamsLen = dreams?.length ?? 0;
     for (let i = 0; i < dreamsLen; i++) {
-      const dream = dreams[i];
+      const dream = dreams![i];
       const row = dream as Record<string, unknown>;
       const content = String(row.content ?? row.CONTENT ?? "");
       if (!checkNoiseBlocklist(content).isNoise) {
@@ -149,7 +169,7 @@ export async function loadContextAtSessionStart(
     if (blockedCount > 0) {
       logger.info(`[Memory Loading] Inject-gate filtered ${blockedCount} noisy dream(s)`);
     }
-    if (cleanDreams.length === 0) {
+    if (cleanDreams.length === 0 && genes.length === 0 && !immuneContext) {
       return "";
     }
 
@@ -164,11 +184,27 @@ export async function loadContextAtSessionStart(
       parts.push("");
     }
 
+    if (genes.length > 0) {
+      parts.push("\n# 🧬 Verified Genome Patterns\n");
+      for (const gene of genes) {
+        parts.push(`### ${gene.name} (category: ${gene.category}, confidence: ${gene.confidence.toFixed(2)})`);
+        if (gene.problem) {
+          parts.push(`Problem: ${gene.problem}`);
+        }
+        parts.push(`Solution: ${gene.solution}`);
+        parts.push("");
+      }
+    }
+
+    if (immuneContext) {
+      parts.push(immuneContext);
+    }
+
     const context = parts.join("\n");
 
     // Log context loading event
     const fs = await import('node:fs');
-    const logEntry = `[${new Date().toISOString()}] LOADED: session=${sessionId}, project=${project}, dreams=${dreams.length}\n`;
+    const logEntry = `[${new Date().toISOString()}] LOADED: session=${sessionId}, project=${project}, dreams=${dreams?.length ?? 0}, genes=${genes.length}, immune=${immuneContext ? 1 : 0}\n`;
     try {
       await fs.promises.appendFile('/tmp/memory_loading.log', logEntry);
     } catch (err) {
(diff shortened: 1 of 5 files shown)


giauphan (commented):
cc @jules

# Code Review

## Summary
This PR adds genome integration into the dreaming pipeline and LLM memory loading, plus a new `verified_solution_search` MCP tool. The intent is solid — connecting genome patterns to dreams and surfacing verified solutions. However, there are several issues worth flagging.

---

## Issues

**File: `src/services/llmService.ts` → Line ~134: `any`-typed `genes` variable via `Awaited<ReturnType<...>>`**

The type `Awaited<ReturnType<typeof GenomeService.searchGenes>>` is fine structurally, but the mutable reassignment on line ~140 (`genes = genes.filter(...)`) means TypeScript may widen the type depending on the `searchGenes` return signature. Prefer a `const` for the filtered result:

```typescript
const allGenes = await GenomeService.searchGenes(strictTask, { project, limit: 5 });
const genes = allGenes.filter((g) => g.confidence >= 0.5 && g.score >= 0.3);
```

This avoids the `let` + reassignment pattern and is clearer.

---

**File: `src/services/llmService.ts` → Line ~161: Non-null assertion `dreams![i]`**

Using `dreams!` is a code smell. At this point, `dreams` could genuinely be `null`/`undefined` (the original code already had `if (!dreams || dreams.length === 0)` guarding it, but that guard was loosened). If `dreams` is nullish but `genes.length > 0`, execution reaches the loop with `dreamsLen = 0`, which is safe — but the `!` assertion on line 161 is still fragile. Consider:

```typescript
const safeDreams = dreams ?? [];
```
early on, then use `safeDreams` throughout. Eliminates the need for `!` and `??` scattered across multiple lines.

---

**File: `src/services/dreamPipelineService.ts` → Line ~214-228: Sequential genome extraction (N+1 pattern)**

Each saved dream triggers a separate `GenomeService.extractGene()` call sequentially. For large dream batches, this is a performance bottleneck. Consider:

```typescript
const geneResults = await Promise.allSettled(
  savedDreams.map((d) =>
    authStorage.run(auth, () =>
      GenomeService.extractGene({ sourceType: "dream", sourceId: d.id, project, autoExtract: true })
    )
  )
);
const genesSaved = geneResults.filter((r) => r.status === "fulfilled").length;
```

Or at minimum, batch with a concurrency limiter if you're worried about overloading the genome service.

---

**File: `src/presentation/mcpTools.ts` → Line ~1547: Hardcoded confidence/score thresholds**

```typescript
const relevantGenes = genes.filter((g) => g.confidence >= 0.5 && g.score >= 0.3);
```

These same magic numbers (`0.5`, `0.3`) appear in both `mcpTools.ts` and `llmService.ts`. Extract them as named constants to avoid drift:

```typescript
const MIN_GENE_CONFIDENCE = 0.5;
const MIN_GENE_SCORE = 0.3;
```

---

**File: `src/presentation/mcpTools.ts` → Line ~1553-1556: Keyword search on graph nodes is naive**

```typescript
n.label.toLowerCase().includes(kw) ||
String(n.filePath || "").toLowerCase().includes(kw)
```

If `query` is something like `"a"` or `"error"`, this will match nearly every node, producing noise. The `z.string().min(1)` validation on `query` is too permissive for this usage. Consider a minimum keyword length for the trace (e.g., `kw.length >= 3`), or document that `keyword` should be a specific feature name.

---

**File: `src/services/llmService.ts` → Line ~199: `immuneContext` injected without sanitization**

```typescript
if (immuneContext) {
  parts.push(immuneContext);
}
```

`immuneContext` is a raw string from `GenomeService.buildImmuneContext()` injected directly into an LLM prompt. If the immune data contains adversarial content (prompt injection stored in a gene), this is a vector. Worth at minimum logging the length/content hash, or sanitizing markdown.

---

## What Looks Good

- Error handling around genome calls is properly wrapped in try/catch with meaningful log messages.
- The `verified` boolean gate requiring both gene match AND codebase flow is a good design — prevents false positives.
- The `genesSaved` counter addition to the pipeline report is clean and consistent with the existing pattern.

---

**Verdict:** Address the non-null assertion, the duplicated magic numbers, and the sequential extraction performance issue. The rest are lower priority but worth tracking.

src/services/llmService.ts
```diff
@@ -1,5 +1,6 @@
 import { logger } from "../utils/logger.js";
```
giauphan, line 1: `src/services/llmService.ts` → Line ~134: `any`-typed `genes` variable via `Awaited<ReturnType<...>>`**

giauphan, line 1: `src/services/llmService.ts` → Line ~161: Non-null assertion `dreams![i]`**

giauphan, line 1: `src/services/llmService.ts` → Line ~199: `immuneContext` injected without sanitization**

giauphan (commented):
cc @jules

The diff's target files aren't in this workspace (it's the codeatlas-ai project), so I reviewed the diff standalone against the hunk context. Findings:

**Correctness**

- File: `src/presentation/mcpTools.ts` -> Line ~1553-1557: The flow gate does `label.toLowerCase().includes(kw)` with `kw = keyword || query`. Since the default keyword is the raw free-text query (up to 2000 chars), multi-word queries like "fix auth bug" will almost never be a substring of a node label or file path — `verified` will be near-always false unless callers pass a single-token `keyword`. Tokenize the query and match per-token (or per significant term), otherwise gate 2 is effectively broken in default usage.
- File: `src/presentation/mcpTools.ts` -> Line ~1542: When `project` is omitted, `loaded` is null and `flowNodes` is always empty, so `verified` can never be true — but the UNVERIFIED message implies a failed search rather than "no project loaded". Also `searchGenes` gets `projectName = "global"` here while `llmService` passes the raw (possibly undefined) `project` — inconsistent project scoping for the same service; verify `searchGenes` actually treats `"global"` as the global bucket.
- File: `src/services/dreamPipelineService.ts` -> Line ~226: `genesSaved` counts fulfilled promises, not genes actually saved. With `autoExtract: true`, `extractGene` can presumably fulfill with "no gene extracted" — verify the resolved value (e.g. `r.value` non-null) before counting. Failures are also silently swallowed; log the `rejected` reasons or you'll never see extraction failures.

**Performance**

- File: `src/services/dreamPipelineService.ts` -> Line ~214: `Promise.allSettled` fires `extractGene` (LLM-backed via `autoExtract`) concurrently for every saved dream (up to ~10 per the summarizer window). Unbounded parallel LLM calls risk rate limits/cost spikes; chunk or run sequentially.

**Code quality**

- File: `src/services/llmService.ts` -> Line ~142: The `0.5`/`0.3` thresholds hardcode the same values as `MIN_GENE_CONFIDENCE`/`MIN_GENE_SCORE` in mcpTools.ts — they'll drift. Move the constants to a shared module and import in both.
- File: `src/presentation/mcpTools.ts` -> Line ~1546: `minGenes` is a max (search limit), not a min — misleading name; call it `geneLimit`.
- File: `src/services/llmService.ts` -> Line ~141-148: `searchGenes` and `buildImmuneContext` are independent — run them concurrently (`Promise.allSettled` fits the existing per-call error handling) instead of sequential awaits.

Not flagged: no injection/secrets risk (inputs only feed substring matching and service queries); the `/tmp/memory_loading.log` append is pre-existing. The `safeDreams`/conditional-header changes in llmService correctly fix the prior null-dream and always-printed-header issues.

The llmService changes are otherwise solid (proper try/catch isolation, early-return logic is correct); the main thing I'd insist on before merge is fixing the keyword substring gate in `verified_solution_search`, since "verified" is the tool's core promise.

src/services/llmService.ts
giauphan, line 1: `src/services/llmService.ts` -> Line ~142: The `0.5`/`0.3` thresholds hardcode the same values as `MIN_GENE_CONFIDENCE`/`MIN_GENE_SCORE` in mcpTools.ts — they'll drift. Move the constants to a shared module and import in both.

giauphan, line 1: `src/services/llmService.ts` -> Line ~141-148: `searchGenes` and `buildImmuneContext` are independent — run them concurrently (`Promise.allSettled` fits the existing per-call error handling) instead of sequential awaits.

PR #305, opened 2026-09-12
🎨 Palette: Enable one-click query clearing in Second Brain search

giauphan:
💡 **What**: Enabled the clear button (`hasClearButton={true}`) for the search input in the Second Brain view and implemented the `onClear` handler to reset the query and refresh the concepts list.
🎯 **Why**: Improves user experience by allowing one-click query clearing. Users no longer have to manually backspace through their entire search term to reset the view.
📸 **Before/After**: Visually, the clear button (`X`) now appears inside the search input when text is entered, shifting the `Enter` hint over gracefully.
♿ **Accessibility**: Provides a quick, distinct, tab-focusable action to cancel search states, which benefits keyboard and screen-reader users over manual text deletion.

---
*PR created automatically by Jules for task [12798813166924153075](https://jules.google.com/task/12798813166924153075) started by @giauphan*

### DIFF
--- a/dashboard/src/components/SecondBrainView.tsx
+++ b/dashboard/src/components/SecondBrainView.tsx
@@ -49,7 +49,7 @@ export function SecondBrainView() {
       const list: Concept[] = data.concepts || [];
       setConcepts(list);
     } catch (err: unknown) {
-      if (signal.aborted || (err instanceof Error && err.name === 'AbortError')) return;
+      if (signal.aborted) return;
       setError(err instanceof Error ? err.message : String(err));
     } finally {
       if (abortControllerRef.current?.signal === signal) {
@@ -61,6 +61,7 @@ export function SecondBrainView() {
   useEffect(() => {
     return () => {
       abortControllerRef.current?.abort();
+      abortControllerRef.current = null;
     };
   }, []);
 
--- a/dashboard/src/components/SecondBrainView.tsx
+++ b/dashboard/src/components/SecondBrainView.tsx
@@ -48,8 +48,8 @@ export function SecondBrainView() {
       const data = await resp.json();
       const list: Concept[] = data.concepts || [];
       setConcepts(list);
-    } catch (err: any) {
-      if (err.name === 'AbortError') return;
+    } catch (err: unknown) {
+      if (signal.aborted || (err instanceof Error && err.name === 'AbortError')) return;
       setError(err instanceof Error ? err.message : String(err));
     } finally {
       if (abortControllerRef.current?.signal === signal) {
@@ -57,6 +57,12 @@ export function SecondBrainView() {
       }
     }
   }, [API_BASE]);
+
+  useEffect(() => {
+    return () => {
+      abortControllerRef.current?.abort();
+    };
+  }, []);
 
   useEffect(() => {
     fetchConcepts();
--- a/dashboard/src/components/SecondBrainView.tsx
+++ b/dashboard/src/components/SecondBrainView.tsx
@@ -129,9 +129,13 @@ export function SecondBrainView() {
             value={searchQuery}
             onChange={e => setSearchQuery(e.target.value)}
             onSearch={handleSearch}
+            onClear={() => {
+              setSearchQuery('');
+              fetchConcepts('');
+            }}
             placeholder="Search concepts…"
             ariaLabel="Search concepts"
-            hasClearButton={false}
+            hasClearButton={true}
           />
         </div>
         <div role="group" aria-label="Filter concepts by category" style={{ display: 'flex', gap: '0.5rem', flexWrap: 'wrap' }}>
--- a/dashboard/src/components/SecondBrainView.tsx
+++ b/dashboard/src/components/SecondBrainView.tsx
@@ -1,4 +1,4 @@
-import { useState, useEffect, useCallback } from 'react';
+import { useState, useEffect, useCallback, useRef } from 'react';
 import { Brain, Search, Lightbulb, TrendingUp, Archive, AlertCircle, Loader2, RefreshCw } from 'lucide-react';
 import { motion } from 'framer-motion';
 import { getAuthHeaders } from '../lib/auth';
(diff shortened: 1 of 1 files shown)


giauphan:
cc @jules

The file isn't in this repo, so I'm reviewing from the diff alone — one caveat applies where the surrounding code matters.

## Review

The change is small and mostly fine: enabling the clear button with a handler that resets state is reasonable. Two things worth flagging:

**File: dashboard/src/components/SecondBrainView.tsx -> Line 132-135: Double-fetch / state-vs-fetch inconsistency.** `onClear` both sets state (`setSearchQuery('')`) and manually calls `fetchConcepts('')`. If a `useEffect` already refetches when `searchQuery` changes (the common pattern when an `onChange` handler only sets state), this fires two identical requests, and the manual call may win or lose the race. If no such effect exists, then fetching is driven by event handlers rather than state — meaning `onChange`/typing never triggers a fetch and only `onSearch`/`onClear` do. Either way, pick one source of truth: either derive fetching from `searchQuery` and make `onClear` just `setSearchQuery('')`, or keep handler-driven fetching but be explicit.

**File: dashboard/src/components/SecondBrainView.tsx -> Line 134: Stale-response race.** `fetchConcepts('')` doesn't cancel in-flight requests. Sequence: user searches "foo" → response is slow → user clears → `''` response returns → slow "foo" response lands last → UI shows "foo" results under a cleared input. Worth aborting or ignoring stale responses inside `fetchConcepts` (AbortController or a request-id check), though this is a pre-existing pattern if `handleSearch` shares it.

**File: dashboard/src/components/SecondBrainView.tsx -> Line 133: Pagination not reset (verify).** If the concepts list is paginated or has page state, clearing search should typically reset to page 1; only the query is reset here. Ignore if `fetchConcepts` handles that internally.

Everything else looks good — the `hasClearButton={true}` toggle is correct, and the inline handler is fine for something this size.

---

giauphan:
cc @jules

This diff is mostly well done — the AbortController race handling is actually correct (the `finally` guard comparing `abortControllerRef.current?.signal === signal` properly prevents a stale request from clearing the loading state of a newer one, and aborted fetches reject immediately so there's no stale `setConcepts` write). No security concerns; the query goes through `URLSearchParams`. A few things worth flagging:

**Issues:**

- File: dashboard/src/components/SecondBrainView.tsx -> Line 33: The AbortController is never aborted on unmount. An in-flight request keeps running after the component unmounts and still calls `setConcepts`/`setError`/`setLoading` on a dead component. Add cleanup: `useEffect(() => () => abortControllerRef.current?.abort(), [])`.

- File: dashboard/src/components/SecondBrainView.tsx -> Line 51: `catch (err: any)` weakens type safety — use `catch (err: unknown)` with an `isAbortError(err)` type guard. Also, if a non-Error value were ever thrown, `err.name` would itself throw inside the catch block.

- File: dashboard/src/components/SecondBrainView.tsx -> Line 52: `err.name === 'AbortError'` won't match in all environments (some throw `TypeError`/`DOMException` variants or set `code: 20`). Safer: check `abortControllerRef.current?.signal.aborted` (well, the captured `signal.aborted`) alongside the name check — that's also immune to the catch-type issue above.

- File: dashboard/src/components/SecondBrainView.tsx -> Line 141: `onClear` unconditionally refetches the full list. Verify the `SearchBar` component only fires `onClear` from the clear-button click — if it also fires when the input empties via backspace, every deletion to empty triggers a full refetch.

**Non-issues (checked):** the abort-during-`getAuthHeaders()` race is handled (fetch with an already-aborted signal rejects immediately), the `onClear` + `setSearchQuery('')` ordering causes no double-fetch since `onChange` doesn't fetch, and `query?.trim()` falsy correctly omits the param.

Everything else looks good.

---

giauphan:
cc @jules

Review looks good overall — the AbortController wiring correctly fixes the stale-response race (newer fetch sets state before older one completes), and the `finally` guard comparing `abortControllerRef.current?.signal === signal` is the right pattern.

Only minor nits:

- File: dashboard/src/components/SecondBrainView.tsx -> Line 56 (unmount cleanup): `abortControllerRef.current?.abort()` leaves the ref non-null, so the aborted request's `finally` still passes the guard and calls `setLoading(false)` on an unmounted component. Harmless in React 18 (no-op), but setting `abortControllerRef.current = null` in the cleanup makes the guard skip it.

- File: dashboard/src/components/SecondBrainView.tsx -> Line 44 (catch block): checking both `signal.aborted` and `err.name === 'AbortError'` is redundant — `signal.aborted` alone covers all abort cases, including the pre-aborted-signal path where `fetch` throws immediately. Not a bug, just extra branches.

- File: dashboard/src/components/SecondBrainView.tsx -> Line 39 (pre-existing): `throw new Error(await resp.text())` surfaces the raw backend response body directly in the UI, which can leak internal error details. Out of scope for this PR, but worth a follow-up.

One thing worth double-checking outside the diff: if the `SearchInput` component (or a parent effect) triggers a search on `searchQuery` change, `setSearchQuery('')` plus the explicit `fetchConcepts('')` in `onClear` would double-fetch — though the abort logic makes the second call win, so it degrades gracefully.

No correctness bugs, no security issues introduced, no performance concerns — the change is a net improvement.

---

giauphan:
cc @jules

Review done — the change is solid overall. The abort-controller pattern is implemented correctly: stale requests get cancelled, aborted requests don't clobber `error`, the `finally` guard prevents an older request from resetting `loading` while a newer one is in flight, and the unmount cleanup aborts in-flight fetches. No security or performance concerns.

One real (minor) issue:

- File: dashboard/src/components/SecondBrainView.tsx -> Line ~49: The success path doesn't check for abort before applying state. If the request is aborted while `resp.json()` resolves (or between awaits), `setConcepts(list)` can still run and briefly apply stale results over the newer request's in-flight state. Add the same guard before setting state:

```ts
const list: Concept[] = data.concepts || [];
if (signal.aborted) return;   // add this
setConcepts(list);
```

Minor nit (optional): `if (abortControllerRef.current) abortControllerRef.current.abort();` can be `abortControllerRef.current?.abort()`.

The `onClear` handler is fine — clearing resets the query and refetches the unfiltered list, and there's no duplicate-fetch risk since no effect watches `searchQuery`.

---

giauphan:
cc @jules

I've reviewed the diff — this is a solid change. The AbortController pattern is implemented correctly: stale requests are aborted before each new fetch, the post-`json()` `signal.aborted` check prevents setting state from a cancelled request, the `finally` block only clears `loading` for the latest request (ref signal identity check), and the unmount cleanup aborts in-flight requests and nulls the ref, so no state updates on unmounted components and no memory leaks.

No correctness, security, or performance issues found. Two minor notes (not blockers):

- File: dashboard/src/components/SecondBrainView.tsx -> Line ~150: `onClear` calls `fetchConcepts('')` directly, bypassing `handleSearch`. If `SearchInput` also fires `onChange`/`onSearch` on clear, this could double-fetch; worth a quick verify.
- File: dashboard/src/components/SecondBrainView.tsx -> Line ~47: the `if (signal.aborted) return;` after `resp.json()` is redundant-but-harmless since the catch already guards; keeping it is fine as a cheap early exit.

Everything else looks good — the clear-button wiring (`hasClearButton={true}`, resetting query and refetching all concepts) is correct.

---

PR #306, opened 2026-09-12
⚡ Bolt: Replace O(N^2) graph filters with O(N) adjacency maps

giauphan:
💡 What: Replaced nested `.filter()` and `.map()` calls within graph traversal iterations with pre-calculated O(1) Map lookups for `callsTo` and `calledBy` arrays in `src/presentation/mcpTools.ts`.
🎯 Why: The original logic in `trace_feature_flow` used `dedupLinks.filter` for every node in the graph inside a topological sort, leading to O(V * E) performance (which scales quadratically on dense graphs).
📊 Impact: Considerably faster graph traversal and generation of Mermaid charts, significantly reducing the algorithmic time complexity of adjacency lookups from O(V*E) to O(V+E).
🔬 Measurement: Verify by executing large graph resolutions using the `trace_feature_flow` MCP tool. `CI=true pnpm test` was executed successfully indicating no logic breakages or functionality regressions.

---
*PR created automatically by Jules for task [6687100868799779405](https://jules.google.com/task/6687100868799779405) started by @giauphan*

### DIFF
--- a/plan.md
+++ b/plan.md
@@ -1,0 +1,49 @@
+1. **Identify the performance bottleneck:**
+   In `src/presentation/mcpTools.ts`, within the `trace_feature_flow` tool (around lines 1180-1250), there is a bottleneck when building the `executionOrder` array.
+   Inside the topological sort loop (and the subsequent loop for remaining nodes), the code does:
+   ```typescript
+   const callsTo = dedupLinks
+     .filter((l) => l.source === current)
+     .map((l) => nodeNameMap.get(l.target) || l.target);
+   const calledBy = dedupLinks
+     .filter((l) => l.target === current)
+     .map((l) => nodeNameMap.get(l.source) || l.source);
+   ```
+   This iterates over `dedupLinks` (an array) for every node in `traceNodes`. This makes it an O(V * E) operation (where V is the number of nodes and E is the number of edges), which is basically O(N^2) for dense graphs.
+
+2. **Optimize with O(N) preprocessing:**
+   We can pre-calculate the `callsTo` and `calledBy` relationships for all nodes in a single pass over `dedupLinks`.
+   ```typescript
+   // Pre-calculate relationship arrays to avoid O(V*E) nested loops
+   const callsToMap = new Map<string, string[]>();
+   const calledByMap = new Map<string, string[]>();
+
+   for (const link of dedupLinks) {
+     const sourceName = nodeNameMap.get(link.source) || link.source;
+     const targetName = nodeNameMap.get(link.target) || link.target;
+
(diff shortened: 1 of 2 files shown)


giauphan:
cc @jules

The files aren't in this workspace, so I'm reviewing from the diff alone. The optimization logic is sound: the map keys (`link.source`/`link.target`) match exactly what the old `filter()` predicates compared, array order is preserved (pushes in link-iteration order), the `|| []` fallback matches the empty-filter case, and `inDegree` logic is untouched. It's a genuine O(V·E) → O(V+E) improvement with no behavioral change. Only minor flags:

**Issues:**

1. File: plan.md -> Line 1: This is a planning/process artifact (workflow notes like "Wait for pre-commit instructions response"), not project documentation — it should be deleted or gitignored rather than committed.

2. File: src/presentation/mcpTools.ts -> Line 1196: The `map.has()` + `map.get()!.push()` pattern uses non-null assertions that are only safe by proximity. A get-or-init helper (`const arr = callsToMap.get(k) ?? []; ... map.set(k, arr); arr.push(...)`) would be assertion-free and slightly less fragile — style nit, not a blocker.

**Verdict:** Correct and worthwhile performance fix; no correctness, security, or edge-case regressions. Merge is fine once plan.md is dropped.

---

giauphan:
cc @jules

The file isn't in this workspace, so I reviewed the diff on its own. The transformation is logically equivalent to the old filtering, so this looks good:

- **Correctness**: Both maps are keyed by the same node ids used for lookup (`current` / `node.id`), name resolution (`nodeNameMap.get(x) || x`) matches the old code exactly, and insertion order over `dedupLinks` preserves the old `filter()` ordering. Nodes with no links, self-loops, and multi-edges all behave identically (`|| []` covers map misses).
- **Performance**: The change removes real O(V·E) scans per dequeued node, replacing them with one O(E) pass plus O(1) lookups. Memory cost is two arrays totaling O(E) — negligible.
- **Security/quality**: No new risks; no `any` types introduced.

Two non-blocking nits only:

- File: src/presentation/mcpTools.ts -> Line ~1183: The "⚡ Bolt Optimization" comment reads like AI-generated noise and slightly misstates the fix (it removed repeated linear scans, not "nested loops"). Consider trimming or dropping it.
- File: src/presentation/mcpTools.ts -> Line ~1194: `callsToMap.set(link.source, toArr)` / `calledByMap.set(link.target, byArr)` run redundantly on every push; a small `pushToMap(map, key, value)` helper (or `map.get(k)?.push(v)` after ensuring the key) would tighten it, but this is cosmetic.

Overall: correct and worthwhile optimization. Nothing that must change before merge.

---

PR #307, opened 2026-09-13
⚡ Bolt: Make dreaming memory access count bumps non-blocking

giauphan:
💡 What: Refactored `access_count` bumping in `dreamingService.ts` to be a non-blocking "fire-and-forget" asynchronous operation.
🎯 Why: Previously, reading dreaming memories with vector search or large offsets resulted in the function waiting on an `UPDATE` statement before it could return data to the user. This "Write on Read" operation bottlenecked response times unnecessarily.
📊 Impact: Expected to reduce read latency by avoiding the N+1 `executeMany` await overhead on every read.
🔬 Measurement: Verified with existing tests to ensure correctness is preserved. Response times for `queryDreamMemories` with `vectorScores` should be measurably lower under high concurrency.

---
*PR created automatically by Jules for task [6715238038392455652](https://jules.google.com/task/6715238038392455652) started by @giauphan*

### DIFF
--- a/src/presentation/httpServer.ts
+++ b/src/presentation/httpServer.ts
@@ -1227,8 +1227,13 @@ export function startHttpServer(port: number, retries = 5): Promise<void> {
           }
 
           // Define shutdown handler
-          const shutdown = (signal: string, server: Server) => {
+          const shutdown = async (signal: string, server: Server) => {
             logger.info(`${signal} received: Closing HTTP server...`);
+            const { DreamingService } = await import('../services/dreamingService.js');
+            if (DreamingService.activeBackgroundTasks.size > 0) {
+              logger.info(`Waiting for ${DreamingService.activeBackgroundTasks.size} pending database writes to complete...`);
+              await Promise.allSettled(Array.from(DreamingService.activeBackgroundTasks));
+            }
             server.close(() => {
               logger.info('HTTP server closed');
               process.exit(0);
--- a/src/presentation/httpServer.ts
+++ b/src/presentation/httpServer.ts
@@ -1229,14 +1229,22 @@ export function startHttpServer(port: number, retries = 5): Promise<void> {
           // Define shutdown handler
           const shutdown = async (signal: string, server: Server) => {
             logger.info(`${signal} received: Closing HTTP server...`);
-            const { DreamingService } = await import('../services/dreamingService.js');
-            if (DreamingService.activeBackgroundTasks.size > 0) {
-              logger.info(`Waiting for ${DreamingService.activeBackgroundTasks.size} pending database writes to complete...`);
-              await Promise.allSettled(Array.from(DreamingService.activeBackgroundTasks));
-            }
-            server.close(() => {
+            server.close(async () => {
               logger.info('HTTP server closed');
-              process.exit(0);
+              try {
+                const { DreamingService } = await import('../services/dreamingService.js');
+                if (DreamingService.activeBackgroundTasks.size > 0) {
+                  logger.info(`Waiting for ${DreamingService.activeBackgroundTasks.size} pending database writes to complete...`);
+                  await Promise.race([
+                    Promise.allSettled(Array.from(DreamingService.activeBackgroundTasks)),
+                    new Promise(r => setTimeout(r, 5000))
+                  ]);
+                }
+              } catch (err) {
+                logger.error('Error during graceful shutdown:', err instanceof Error ? err.message : String(err));
+              } finally {
+                process.exit(0);
+              }
             });
           };
 
--- a/src/presentation/httpServer.ts
+++ b/src/presentation/httpServer.ts
@@ -1227,12 +1227,18 @@ export function startHttpServer(port: number, retries = 5): Promise<void> {
           }
 
           // Define shutdown handler
+          let isShuttingDown = false;
           const shutdown = (signal: string, server: Server) => {
+            if (isShuttingDown) return;
+            isShuttingDown = true;
             logger.info(`${signal} received: Closing HTTP server...`);
 
+            type ModernServer = Server & { closeIdleConnections?: () => void, closeAllConnections?: () => void };
+            const mServer = server as ModernServer;
+
             // Kill idle keep-alive connections first (Node >= 18.2)
-            if (typeof (server as any).closeIdleConnections === 'function') {
-              (server as any).closeIdleConnections();
+            if (typeof mServer.closeIdleConnections === 'function') {
+              mServer.closeIdleConnections();
             }
 
             server.close(async () => {
@@ -1249,8 +1255,8 @@ export function startHttpServer(port: number, retries = 5): Promise<void> {
 
             // Hard exit fallback if `server.close` hangs (e.g. active connections not resolving)
             setTimeout(() => {
-              if (typeof (server as any).closeAllConnections === 'function') {
-                (server as any).closeAllConnections();
+              if (typeof mServer.closeAllConnections === 'function') {
+                mServer.closeAllConnections();
               }
               logger.error('Graceful shutdown timed out after 10s, forcing exit.');
               process.exit(1);
--- a/src/presentation/httpServer.ts
+++ b/src/presentation/httpServer.ts
@@ -1227,25 +1227,28 @@ export function startHttpServer(port: number, retries = 5): Promise<void> {
           }
 
           // Define shutdown handler
-          const shutdown = async (signal: string, server: Server) => {
+          const shutdown = (signal: string, server: Server) => {
             logger.info(`${signal} received: Closing HTTP server...`);
+
+            // Wait a grace period for active requests to finish, or force close
+            server.closeAllConnections(); // For Node >= 18.2 to close keep-alive
             server.close(async () => {
               logger.info('HTTP server closed');
               try {
                 const { DreamingService } = await import('../services/dreamingService.js');
-                if (DreamingService.activeBackgroundTasks.size > 0) {
-                  logger.info(`Waiting for ${DreamingService.activeBackgroundTasks.size} pending database writes to complete...`);
-                  await Promise.race([
-                    Promise.allSettled(Array.from(DreamingService.activeBackgroundTasks)),
-                    new Promise(r => setTimeout(r, 5000))
-                  ]);
-                }
+                await DreamingService.waitForBackgroundTasks(5000);
               } catch (err) {
                 logger.error('Error during graceful shutdown:', err instanceof Error ? err.message : String(err));
               } finally {
                 process.exit(0);
               }
             });
+
+            // Hard exit fallback if `server.close` hangs (e.g. idle connections missing `closeAllConnections` on older node)
+            setTimeout(() => {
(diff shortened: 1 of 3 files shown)


giauphan:
cc @jules

**Review: src/services/dreamingService.ts**

Overall this is a sound optimization — the `.catch()` properly prevents unhandled rejections, the SQL stays parameterized, and removing the `await` genuinely unblocks the read path.

Minor notes:

- File: src/services/dreamingService.ts -> Line 415: Fire-and-forget means the `access_count` bump is silently lost if the process exits or is frozen before the promise resolves (graceful shutdown, serverless). Acceptable since it's non-critical telemetry, but if this service does graceful shutdown, consider tracking the promise (e.g., a pending set awaited on shutdown) so the write isn't dropped mid-flight.
- File: src/services/dreamingService.ts -> Line 414: `binds` shadows the outer query's `binds` (pre-existing, harmless since the outer value is no longer used after `db.query`, but confusing to readers).

No correctness, security, or performance regressions introduced. LGTM aside from the two minor notes.

---

giauphan:
cc @jules

The diff belongs to a different repo than this workspace, so I reviewed it from the diff alone (line numbers derived from the hunks). The core idea — tracking fire-and-forget writes so shutdown can drain them — is sound, and the details are mostly right (`.catch` preserves the old warning log, `finally` cleanup prevents a set leak, `Array.from` snapshots before awaiting, and `add` always precedes `delete` because `.finally` can only run after the sync block). But there are real issues:

**File: src/presentation/httpServer.ts -> Line 1235:** No timeout on the drain. If a DB write hangs, `process.exit(0)` is never reached and SIGTERM appears to hang until the orchestrator SIGKILLs — defeating the purpose. Cap it: `await Promise.race([Promise.allSettled(tasks), new Promise(r => setTimeout(r, 5000))])`.

**File: src/presentation/httpServer.ts -> Line 1230:** The handler is now async and registered via `process.on(...)`, so any rejection (e.g. the dynamic import at line 1232 failing) becomes an unhandled rejection and the process may never exit. Wrap the body in `try/catch` and guarantee `process.exit` in a `finally`.

**File: src/presentation/httpServer.ts -> Line 1232:** The drain runs *before* `server.close()`, so the server keeps accepting requests during the wait — those requests can spawn new bump tasks that were never in the snapshot, extending shutdown and making the "Waiting for N" log inaccurate. Close the server first, then drain, then exit.

**File: src/services/dreamingService.ts -> Line 74:** `Set<Promise<any>>` uses `any`, which violates this project's explicit rule. Use `Set<Promise<unknown>>`.

**File: src/services/dreamingService.ts -> Line 421 (minor):** `updateTask` is referenced inside its own initializer's `.finally`. It works (the callback runs after the `const` is assigned), but it's a fragile self-reference; assigning first and attaching `finally` after `add` reads safer. Also note the old `try/catch` also caught synchronous throws from `db.executeMany` — `.catch` only handles rejections; fine for a standard async driver, but the guarantee did narrow slightly.

No security concerns; the performance change itself (unblocking the read path from a non-critical write) is a genuine improvement.

---

giauphan:
cc @jules

The diff's files aren't in this workspace, so I reviewed from the diff alone. The rejection-handling pattern is sound (`updateTask` gets a `.catch` before entering the Set, so no unhandled-rejection leaks), the snapshot via `Array.from` is correct, and the rename to `updateBinds` fixes the old shadowing. A few real issues remain:

**Issues**

- File: src/presentation/httpServer.ts -> Line 1232: `server.close()` only invokes its callback after all connections drain — a single idle keep-alive client means the callback (and thus the pending-write flush) never runs and the process never exits. Call `server.closeAllConnections()` (Node ≥18.2) after a grace period, or add an `.unref()`'d hard-exit timer, so the flush logic actually executes.
- File: src/presentation/httpServer.ts -> Line 1235: The HTTP layer reaches into `DreamingService.activeBackgroundTasks` internals via dynamic import — tight coupling. Encapsulate it as e.g. `static async DreamingService.waitForBackgroundTasks(timeoutMs)` and call that; the 5s race also becomes testable.
- File: src/presentation/httpServer.ts -> Line 1240: After the 5s timeout, remaining writes are dropped silently — log the still-pending task count after the race so lost writes are observable.
- File: src/services/dreamingService.ts -> Line 417: Behavior regression: if `db.executeMany` throws synchronously (before returning a promise), the error now propagates up the read path — the old `try/catch` covered it, `.catch()` only handles async rejection. Wrap the call, or isolate it in a helper.

**Minor**

- File: src/services/dreamingService.ts -> Line 74: Prefer `static readonly activeBackgroundTasks` — the Set reference is never reassigned, only mutated.
- File: src/presentation/httpServer.ts -> Line 1230: `async` on `shutdown` itself is unnecessary (nothing is awaited in its body); only the inner close callback needs to be async.

Otherwise good: parameterized SQL is unchanged, the bounded shutdown race is the right approach, and the access_count bump was genuinely blocking the hot read path before.

---

giauphan:
cc @jules

The diff targets files not present in this local repo, so this review is based on the diff alone. Overall the design (tracking fire-and-forget writes, waiting for them on shutdown) is sound, but there are a few real issues:

**Issues**

- File: src/presentation/httpServer.ts -> Line 1234: Calling `server.closeAllConnections()` *before* `server.close()` destroys all connections immediately — including ones with in-flight requests — so active requests get cut off mid-flight. This contradicts the comment ("Wait a grace period for active requests to finish"). The graceful pattern is: `server.closeIdleConnections()` up front (kills idle keep-alive only), then `server.closeAllConnections()` inside the close callback or after a grace timeout.

- File: src/presentation/httpServer.ts -> Line 1234: `closeAllConnections()` only exists on Node ≥ 18.2. On older runtimes this throws a `TypeError` inside the signal handler and the process may never exit cleanly. Guard it: `if (typeof server.closeAllConnections === 'function')` — unless the deploy target is guaranteed ≥ 18.2.

- File: src/presentation/httpServer.ts -> Line ~1250: The 6s hard-exit timer starts at signal time, but the 5s `waitForBackgroundTasks` only starts after `server.close` completes. If draining connections takes >1s, the fallback fires `process.exit(1)` mid-wait and drops the writes this PR is trying to protect. (Less likely to bite today because of issue #1, but becomes real once #1 is fixed.)

- File: src/services/dreamingService.ts -> Line 436: `Promise<any>` violates the project's "avoid any types" rule — use `Promise<unknown>`. The whole `try/catch` + `Promise.reject(syncErr)` construct can also be simplified to `Promise.resolve().then(() => db.executeMany(...))` if you need to trap sync throws (unnecessary if `executeMany` is an `async` function).

- File: src/services/dreamingService.ts -> Line ~88: `waitForBackgroundTasks` snapshots the set via `Array.from(...)`; tasks added after the snapshot (e.g., a read finishing during shutdown) are neither awaited nor warned about, since `settled` short-circuits the check. Minor race — re-checking `activeBackgroundTasks.size` after the race would close most of the gap.

**Looks good otherwise**: the `unref()`'d hard-exit fallback, awaiting the wait in `finally` before `process.exit(0)`, the `.catch`-before-tracking ordering (no unhandled rejections), and the self-removing `finally` cleanup are all correct.

---

giauphan:
cc @jules

The diff references `src/presentation/httpServer.ts` and `src/services/dreamingService.ts`, which don't exist in this workspace (this repo is the review bot itself), so I'm reviewing from the diff alone. Overall the design is sound — `closeIdleConnections` → drain → exit with an unref'd 10s failsafe is the right shape, and moving the `access_count` bump off the read path is a genuine perf win. Three things worth flagging:

**Issues**

- File: src/presentation/httpServer.ts -> Line 1235: `(server as any)` casts (also at 1253) violate the project's "avoid any" rule. `closeIdleConnections()`/`closeAllConnections()` are declared on `http.Server` in `@types/node` ≥18.x — call them directly, or if you must keep a runtime guard, use a typed optional intersection (`Server & { closeIdleConnections?: () => void }`) instead of `any`.

- File: src/presentation/httpServer.ts -> Line 1231: `shutdown` has no re-entrancy guard, and this change widens the vulnerable window from near-instant to up to 10s (5s task drain + failsafe). If a second SIGINT/SIGTERM arrives while the first shutdown is pending — common with supervisors that signal twice — the handler re-runs and the second `server.close()` throws `ERR_SERVER_NOT_RUNNING` synchronously as an uncaught exception, bypassing the graceful path. Guard with an idempotency flag (or `process.once` semantics).

- File: src/services/dreamingService.ts -> Line 86: `waitForBackgroundTasks` takes a single `Array.from` snapshot and early-returns when the set is empty (Line 81), so a bump task enqueued after the close callback fires is neither awaited nor warned about — exactly the write loss this PR aims to prevent. The window is narrow (task registration is synchronous in the query-result continuation) but the author's own "re-check size" comment acknowledges it while only fixing the warning, not the wait. A small drain loop (re-snapshot until the set is stable across a microtask tick) closes it cheaply.

Minor: `updateTask.finally(...)` (dreamingService, ~Line 444) discards a promise that would reject if the `catch` handler itself threw (e.g., logger failure) — negligible, but assigning to the set *before* attaching `.catch` or using a `settled` flag inside the task avoids the floating chain entirely.

Otherwise clean: SQL stays parameterized, the fire-and-forget task is catch-wrapped so it can't produce unhandled rejections on its own, and the 5s drain / 10s hard-exit headroom is correct. No security concerns.

---

giauphan:
cc @jules

I couldn't find these files in the local workspace (this repo is the review bot itself; the PR belongs to another project), so I reviewed the diff as provided, mapping line numbers from the hunk headers.

**Overall: solid change.** The `isShuttingDown` guard, `closeIdleConnections()` before `close()`, unhandled-rejection safety (`.catch` attached before `.finally`), and the `binds` → `updateBinds` shadowing fix are all correct. The fire-and-forget bump genuinely removes a serialized write from the read path. One real issue plus two minors:

1. File: src/services/dreamingService.ts -> Line 81: `waitForBackgroundTasks` can wait unboundedly — the timeout is per-iteration, so if new bump tasks keep arriving (in-flight requests still finishing after `server.close()`), the loop restarts with a fresh 5s window indefinitely. It also means the caller's "5000ms" is not an upper bound, and the wait can collide with httpServer's 10s hard-exit (killing the process with exit 1 mid-graceful-wait). Fix: compute an absolute deadline once (`const deadline = Date.now() + timeoutMs`) and check it each iteration instead of racing a fresh timer per round.

2. File: src/services/dreamingService.ts -> Line 86: the `setTimeout` inside `timeoutPromise` is never cleared/unref'd when `settlePromise` wins the race. Harmless today only because `process.exit` always follows, but it will hold the event loop open in any future non-shutdown caller — cleaned up naturally by the deadline fix in issue 1.

3. File: src/services/dreamingService.ts -> Line 442: `Promise.resolve().then(() => db.executeMany(...))` is redundant indirection — `executeMany` already returns a promise; `const updatePromise = db.executeMany(sql, updateBinds)` is equivalent.

Everything else — the `ModernServer` feature-detection cast, hard-exit fallback with `unref()`, and the task-set add/delete bookkeeping — looks correct.

---

PR #308, opened 2026-09-13
Fix local mode Firestore dependency

giauphan:
## Summary
- migrate local authentication, API key lookup, activity logging, and project sync metadata to SQLite
- gate remaining Firestore project/settings paths behind `CODEATLAS_USE_FIRESTORE=true`
- add SQLite repository and local HTTP integration coverage to prevent Firestore stream crashes in free/local mode

## Test plan
- [x] `npm run typecheck`
- [x] `npm run build`
- [x] `npm test` (272 tests passed)
- [x] `git diff --check`
- [x] local HTTP integration test with `CODEATLAS_USE_FIRESTORE=false`


giauphan (commented):
cc @jules

I've reviewed this PR — it migrates auth/telemetry from Firestore to the local database adapter layer. The structure is sound (parameterized queries throughout, good scoping on the key DELETE, dead imports cleaned up), but there are real issues worth fixing before merge.

**Security**

- File: src/presentation/httpServer.ts -> Line ~710: The `INSERT INTO keys` now stores the **plaintext API key** (`key: newKey`) next to `key_hash` — the old Firestore code stored only the hash. This defeats the PBKDF2 hashing entirely; anyone with DB read access gets every usable key. Paired with `SqliteAuthRepository.verifyKey` accepting `OR key = :apiKey` (src/repositories.ts -> Line ~145), plaintext becomes an active auth path. Drop the `key` column and the OR clause.
- File: src/repositories.ts -> Line ~139: `process.env.API_KEY_PEPPER || 'codeatlas-api-key-pepper-v1'` — hardcoded fallback pepper in source. Also verify it matches whatever `API_KEY_PEPPER` resolves to in httpServer.ts create-key; if they differ, every key created via the API can never authenticate.

**Correctness**

- File: src/repositories.ts -> Line ~158: `expires: 0` — epoch 0 is already "expired". The super-admin path explicitly overrides to `Infinity`, implying downstream checks `expires < Date.now()`. If so, every non-admin SQLite-authenticated request fails; return `Infinity` instead.
- File: src/repositories.ts -> Line ~205: `success: success ? 1 : 0` works for SQLite but Postgres will reject it — `activity_log.success` is `BOOLEAN`, and passing an integer param errors ("column is of type boolean"). Since `createDatabaseAdapter()` may return the Postgres adapter, the "Sqlite*" class names are misleading and Postgres deployments break here.
- File: src/services/authService.ts -> Line ~13: The default silently flips from Firestore to local DB whenever `CODEATLAS_USE_FIRESTORE` is unset. Existing deployments using Firestore will silently switch auth stores (existing keys/roles ignored). Needs a migration note or an explicit opt-in default.
- File: src/presentation/httpServer.ts -> Line ~860: In the local sync branch, `effectiveTenantId` is computed but `docId` still uses the raw `tenantId` — the key and the tenant can disagree. Related: `deleteProject`'s telemetry cleanup is now Firestore-gated, so in local mode the `projects` row written by sync is never removed → orphaned rows.

**Performance**

- File: src/middleware/auth.ts -> Line ~23: `createDatabaseAdapter()` is called per request on the hot auth path — fine only if the factory memoizes; if it opens a new connection/pool each call, that's per-request overhead.
- File: src/repositories.ts -> Line ~145: `WHERE key_hash = :keyHash OR key = :apiKey` — the OR across two columns likely defeats index usage; plus `pbkdf2Sync` (100k iterations) blocks the event loop on every uncached auth. Fixing the plaintext issue (lookup by hash only) resolves both.

The rest (activity_log DDL, tenant index, import cleanup, keys-list query, 404 on delete) looks good.

src/repositories.ts
```diff
@@ -1,6 +1,8 @@
 import { getFirestore, FieldValue } from "firebase-admin/firestore";
```
giauphan, line 1: src/repositories.ts -> Line ~139: `process.env.API_KEY_PEPPER || 'codeatlas-api-key-pepper-v1'` — hardcoded fallback pepper in source. Also verify it matches whatever `API_KEY_PEPPER` resolves to in httpServer.ts create-key; if they differ, every key created via the API can never authenticate.

giauphan, line 1: src/repositories.ts -> Line ~158: `expires: 0` — epoch 0 is already "expired". The super-admin path explicitly overrides to `Infinity`, implying downstream checks `expires < Date.now()`. If so, every non-admin SQLite-authenticated request fails; return `Infinity` instead.

giauphan, line 1: src/repositories.ts -> Line ~205: `success: success ? 1 : 0` works for SQLite but Postgres will reject it — `activity_log.success` is `BOOLEAN`, and passing an integer param errors ("column is of type boolean"). Since `createDatabaseAdapter()` may return the Postgres adapter, the "Sqlite*" class names are misleading and Postgres deployments break here.

src/services/authService.ts
```diff
@@ -1,15 +1,18 @@
-import { 
-  FirestoreAuthRepository, 
-  FirestoreActivityLogger, 
-  AuthenticateUserUseCase, 
-  LogTelemetryUseCase 
+import {
```
giauphan, line 1: src/services/authService.ts -> Line ~13: The default silently flips from Firestore to local DB whenever `CODEATLAS_USE_FIRESTORE` is unset. Existing deployments using Firestore will silently switch auth stores (existing keys/roles ignored). Needs a migration note or an explicit opt-in default.

src/middleware/auth.ts
```diff
@@ -1,7 +1,7 @@
 import express from "express";
```
giauphan, line 1: src/middleware/auth.ts -> Line ~23: `createDatabaseAdapter()` is called per request on the hot auth path — fine only if the factory memoizes; if it opens a new connection/pool each call, that's per-request overhead.

src/repositories.ts
giauphan, line 1: src/repositories.ts -> Line ~145: `WHERE key_hash = :keyHash OR key = :apiKey` — the OR across two columns likely defeats index usage; plus `pbkdf2Sync` (100k iterations) blocks the event loop on every uncached auth. Fixing the plaintext issue (lookup by hash only) resolves both.

giauphan (commented):
cc @jules

I've gone through the diff — this migrates auth, API keys, activity logging, and project metadata from Firestore to a local DB layer. The parameterized SQL, boolean sanitization, scoped key deletion, and removal of raw-key storage are all solid. But there are a few real problems:

**File: src/presentation/httpServer.ts -> Line ~711-733 (POST /keys): [high]** Key creation unconditionally writes to SQLite regardless of auth backend. With `CODEATLAS_USE_FIRESTORE=true` (or the `GOOGLE_APPLICATION_CREDENTIALS` fallback), `FirestoreAuthRepository.verifyKey()` looks keys up in Firestore, so keys created here can never authenticate. Same for the delete endpoint (Line ~748-756): deleting from SQLite won't remove a Firestore key. These should branch like the other endpoints.

**File: src/services/authService.ts -> Line 8-16: [high]** Two inconsistent sources of truth for "am I on Firestore": module-level `shouldUseFirestore()` (evaluated once, with the `GOOGLE_APPLICATION_CREDENTIALS` heuristic) vs. the repeated inline `process.env.CODEATLAS_USE_FIRESTORE === 'true'` checks in httpServer.ts. In the heuristic mode, auth goes to Firestore while key creation, project cleanup, and sync skip it — split-brain state. Centralize in one helper used everywhere.

**File: src/database/adapters/postgresAdapter.ts -> Line 255-258: [high]** The `keys` schema change only applies to fresh databases (`CREATE TABLE IF NOT EXISTS`). Existing Postgres deployments keep `key NOT NULL` / nullable `key_hash`, so the new `INSERT INTO keys (id, tenant_id, user_id, name, key_hash, tier)` fails on a NOT NULL violation. Postgres needs an ALTER/backfill equivalent to the SQLite migration.

**File: src/database/adapters/sqliteAdapter.ts -> Line 366-396: [medium]** The `key` column migration is fragile: (a) `key` was UNIQUE, so `DROP COLUMN` will nearly always fail, and the fallback is selected by matching error-message substrings (`"cannot drop"`, `"UNIQUE"`) — wording/version dependent; (b) the rebuild (create `keys_new` → copy → `DROP TABLE keys` → rename) isn't wrapped in a transaction, so a mid-way crash loses the table; (c) `INSERT OR IGNORE` silently drops rows with NULL or duplicate `key_hash`. Detect the constraint via `pragma index_list` and run the rebuild inside a transaction.

**File: src/presentation/httpServer.ts -> Line ~880-900 (sync API local branch): [medium]** Local mode persists only `id/tenant_id/name/updated_at`, dropping `stats`, `lastIndexed`, `nodesCount`, `linksCount` that the Firestore branch stores — telemetry parity gap for anyone switching off Firestore.

**File: scripts/db-seed.ts -> Line 34-35: [medium]** Duplicates the pepper + PBKDF2 logic instead of reusing `getApiKeyPepper()`/`hashApiKey()` from repositories.ts — drift risk if iterations/length ever change. (If importing repositories.ts is undesirable because it pulls firebase-admin, move the helpers to a standalone crypto util.)

**File: src/repositories.ts -> Line 5: [low]** Hardcoded default pepper `'codeatlas-api-key-pepper-v1'` is a public constant — when `API_KEY_PEPPER` is unset, key hashes get no real protection, and it's also used as a static per-key salt. Fine as a migration default, but worth warning at startup when unset.

**File: src/database/factory.ts -> Line 6-15: [low]** The memoized adapter has no reset and no way to detect a closed instance — if anything calls `adapter.close()` (shutdown, tests), later `createDatabaseAdapter()` returns a dead adapter; `setDatabaseAdapter` also leaks state between tests unless callers reset it. Also verify `PostgresAdapter` translates `:named` placeholders to `$n`, since this PR newly routes auth queries (middleware/auth.ts, repositories.ts) through the DB layer on Postgres.

Everything else (schema for `activity_log`, indexes, unused-import cleanup, `sanitizeParam`) looks good.

src/services/authService.ts
```diff
@@ -1,15 +1,27 @@
...
+  SqliteAuthRepository,
+  SqliteActivityLogger,
+  FirestoreAuthRepository,
+  FirestoreActivityLogger,
+  AuthenticateUserUseCase,
+  LogTelemetryUseCase
 } from "../repositories.js";
```
giauphan, line 8: src/services/authService.ts -> Line 8-16: [high]** Two inconsistent sources of truth for "am I on Firestore": module-level `shouldUseFirestore()` (evaluated once, with the `GOOGLE_APPLICATION_CREDENTIALS` heuristic) vs. the repeated inline `process.env.CODEATLAS_USE_FIRESTORE === 'true'` checks in httpServer.ts. In the heuristic mode, auth goes to Firestore while key creation, project cleanup, and sync skip it — split-brain state. Centralize in one helper used everywhere.

src/database/adapters/postgresAdapter.ts
```diff
@@ -255,8 +255,7 @@ export class PostgresAdapter implements IDatabaseAdapter {
         tenant_id VARCHAR(255) NOT NULL REFERENCES tenants(id),
```
giauphan, line 255: src/database/adapters/postgresAdapter.ts -> Line 255-258: [high]** The `keys` schema change only applies to fresh databases (`CREATE TABLE IF NOT EXISTS`). Existing Postgres deployments keep `key NOT NULL` / nullable `key_hash`, so the new `INSERT INTO keys (id, tenant_id, user_id, name, key_hash, tier)` fails on a NOT NULL violation. Postgres needs an ALTER/backfill equivalent to the SQLite migration.

src/database/adapters/sqliteAdapter.ts
```diff
@@ -339,6 +363,36 @@ export class SQLiteAdapter implements IDatabaseAdapter {
       }
     };
 
+    const keyColumns = this.db!.pragma("table_info(keys)") as Array<{ name: string }>;
```
giauphan: src/database/adapters/sqliteAdapter.ts -> Line 366-396: [medium]** The `key` column migration is fragile: (a) `key` was UNIQUE, so `DROP COLUMN` will nearly always fail, and the fallback is selected by matching error-message substrings (`"cannot drop"`, `"UNIQUE"`) — wording/version dependent; (b) the rebuild (create `keys_new` → copy → `DROP TABLE keys` → rename) isn't wrapped in a transaction, so a mid-way crash loses the table; (c) `INSERT OR IGNORE` silently drops rows with NULL or duplicate `key_hash`. Detect the constraint via `pragma index_list` and run the rebuild inside a transaction.

scripts/db-seed.ts
```diff
@@ -31,19 +31,20 @@ async function seed() {
 
     // 2. Seed API Key (if not exists)
     const apiKey = process.env.CODEATLAS_API_KEY || `sk-${randomUUID().replace(/-/g, "").slice(0, 32)}`;
-    const keyHash = ""; // Will be computed by auth service
+    const PEPPER = process.env.API_KEY_PEPPER || 'codeatlas-api-key-pepper-v1';
```
giauphan: scripts/db-seed.ts -> Line 34-35: [medium]** Duplicates the pepper + PBKDF2 logic instead of reusing `getApiKeyPepper()`/`hashApiKey()` from repositories.ts — drift risk if iterations/length ever change. (If importing repositories.ts is undesirable because it pulls firebase-admin, move the helpers to a standalone crypto util.)

src/repositories.ts
```diff
@@ -1,6 +1,24 @@
...
 import * as crypto from "crypto";
 import { logger } from "./utils/logger.js";
+import { createDatabaseAdapter } from "./database/factory.js";
+import { authStorage } from "./utils/context.js";
```
giauphan, line 5: src/repositories.ts -> Line 5: [low]** Hardcoded default pepper `'codeatlas-api-key-pepper-v1'` is a public constant — when `API_KEY_PEPPER` is unset, key hashes get no real protection, and it's also used as a static per-key salt. Fine as a migration default, but worth warning at startup when unset.

src/database/factory.ts
```diff
@@ -3,11 +3,23 @@ import type { IDatabaseAdapter } from "./adapters/interface.js";
 import { SQLiteAdapter } from "./adapters/sqliteAdapter.js";
 import { PostgresAdapter } from "./adapters/postgresAdapter.js";
 
+// Memoize adapter per application lifetime to prevent connection/pool exhaustion
```
giauphan, line 6: src/database/factory.ts -> Line 6-15: [low]** The memoized adapter has no reset and no way to detect a closed instance — if anything calls `adapter.close()` (shutdown, tests), later `createDatabaseAdapter()` returns a dead adapter; `setDatabaseAdapter` also leaks state between tests unless callers reset it. Also verify `PostgresAdapter` translates `:named` placeholders to `$n`, since this PR newly routes auth queries (middleware/auth.ts, repositories.ts) through the DB layer on Postgres.

giauphan (commented):
cc @jules

The diff targets a codebase I don't have locally (this repo is the review bot itself), so I reviewed it statically — noted where behavior depends on code outside the diff.

**Correctness**

- File: src/repositories.ts -> Line ~163: `SELECT ... expires_at AS expiresAt, user_id AS uid` — Postgres folds unquoted aliases to lowercase, so `row.expiresAt` is `undefined` on Postgres and `expires` becomes `Infinity`: **expired API keys are never rejected on the Postgres backend**. Same bug pattern with `id AS keyId`. Quote the aliases (`AS "expiresAt"`) or normalize in the Postgres adapter.
- File: src/presentation/httpServer.ts -> Line ~697: `created_at AS createdAt` in GET /api/keys has the same lowercase-folding problem — the JSON response field becomes `createdat` on Postgres, breaking clients. Also `tenant_id AS tenantId` (lines ~374, ~700, ~893) silently falls back to the raw uid instead of the real tenant on Postgres, which can mis-scope rows.
- File: src/database/factory.ts -> Line ~10: memoizing the adapter is good for pool exhaustion, but any remaining `disconnect()` caller (graceful shutdown, CLI scripts) now poisons the singleton — every later `createDatabaseAdapter()` returns a disconnected adapter. The dreamingService diff removed its disconnect, but audit all other `disconnect()` call sites.
- File: src/presentation/httpServer.ts -> Line ~755: the 404 branch depends on `result.rowsAffected`. If `IDatabaseAdapter.execute()` doesn't return `rowsAffected` on both adapters, the check is always false and deleting a nonexistent key returns 200.
- File: src/repositories.ts -> Line ~172: `new Date(row.expiresAt).getTime()` yields `NaN` for a malformed value, and `NaN < Date.now()` is false → key treated as never-expiring. Guard with `Number.isNaN`.

**Security**

- File: src/presentation/httpServer.ts -> Line ~751: key deletion is correctly scoped by `user_id` — no IDOR. Plaintext `key` column removal and hash-only lookups are solid improvements. Residual: the default pepper `'codeatlas-api-key-pepper-v1'` is still hardcoded (now warned, which is fair).
- File: scripts/db-seed.ts -> Line ~34: pepper/PBKDF2 logic is duplicated instead of importing `hashApiKey`/`getApiKeyPepper` from `src/repositories.ts`. If iterations or pepper change in one place, seeded hashes silently stop verifying.
- File: src/repositories.ts -> Line ~213: `activity_log.params` stores the full tool-params JSON — consider scrubbing/truncating potentially sensitive input.

**Maintainability / edge cases**

- File: src/database/adapters/postgresAdapter.ts -> Line ~284: legacy rows where `key_hash` is NULL (column was nullable) become permanently orphaned when the plaintext `key` column is dropped, and the pre-existing table never gains the NOT NULL constraint. Consider a backfill-or-skip before dropping.
- File: src/database/adapters/sqliteAdapter.ts -> Line ~370: the recreate fallback does `CREATE TABLE IF NOT EXISTS keys_new` — a leftover `keys_new` from a crashed prior run gets dropped along with `keys`, losing data. Use `DROP TABLE IF EXISTS keys_new` first. Also, the `INSERT ... SELECT cols FROM keys` throws if the legacy table predates the `key_hash` column (the `addColumnIfMissing` calls run after this block), crashing startup.
- File: src/database/adapters/sqliteAdapter.ts -> Line ~374: matching error text (`"cannot drop"`, `"UNIQUE"`) is brittle across SQLite/better-sqlite3 versions; an unmatched drop-failure message gets rethrown even though the recreate path would have handled it.
- File: src/database/factory.ts -> Line ~26: `setDatabaseAdapter(null)` and `resetDatabaseAdapter()` do the same thing — keep one.

The auth-mode migration (`shouldUseFirestore` + backward-compat warning), boolean param sanitization, and tenant-scoped queries all look good. The Postgres alias issue is the one I'd fix before merge — it silently disables key expiry on that backend.

src/repositories.ts
```diff
@@ -1,6 +1,30 @@
 import { getFirestore, FieldValue } from "firebase-admin/firestore";
```
giauphan, line 1: src/repositories.ts -> Line ~163: `SELECT ... expires_at AS expiresAt, user_id AS uid` — Postgres folds unquoted aliases to lowercase, so `row.expiresAt` is `undefined` on Postgres and `expires` becomes `Infinity`: **expired API keys are never rejected on the Postgres backend**. Same bug pattern with `id AS keyId`. Quote the aliases (`AS "expiresAt"`) or normalize in the Postgres adapter.

giauphan, line 1: src/repositories.ts -> Line ~172: `new Date(row.expiresAt).getTime()` yields `NaN` for a malformed value, and `NaN < Date.now()` is false → key treated as never-expiring. Guard with `Number.isNaN`.

giauphan, line 1: src/repositories.ts -> Line ~213: `activity_log.params` stores the full tool-params JSON — consider scrubbing/truncating potentially sensitive input.

PR #323, opened 2026-09-15
⚡ Bolt: [performance improvement]

giauphan:
💡 What: Replaced nested `Array.prototype.filter()` loops over the `links` array with pre-computed adjacency lists (`Map`s) inside `search_entities` and `get_file_entities` MCP tools.
🎯 Why: In large graph datasets, mapping nodes and matching their edges sequentially using nested filters caused massive O(V*E) CPU overhead and latency during AI graph queries. 
📊 Impact: Decreases graph relation mapping complexity from O(V*E) to O(V+E), enabling significantly faster search resolutions on projects with large link populations. 
🔬 Measurement: Verified with `pnpm run typecheck` and `CI=true pnpm test`. Test run durations should demonstrate slightly snappier handling of larger inputs.

---
*PR created automatically by Jules for task [4054697643792965987](https://jules.google.com/task/4054697643792965987) started by @giauphan*


giauphan:
cc @jules

The diff is self-contained (the file belongs to the reviewed project, not this repo), so I reviewed it directly.

**Overall:** The optimization is correct. Both hunks preserve behavior exactly — same key comparisons (`l.target === n.id` / `l.source === e.id`), same label fallbacks (`nodeMap.get(...) || id`), and Map insertion order preserves the original link ordering. Complexity genuinely drops from O(M·E) to O(E + M). No security, `any`, or injection concerns. Two minor issues worth flagging:

- File: src/presentation/mcpTools.ts -> Line 394: The outgoing-adjacency builder (`sourceMap` loop) is now duplicated verbatim between the two functions, and the has/set/get/`!` pattern repeats 4x. Extract a small helper (e.g. `buildAdjacencyIndex(links, nodeMap)` returning `{ incomingByTarget, outgoingBySource }`) — especially since `.jules/bolt.md` documents this as a reusable pattern for future graph traversals.

- File: src/presentation/mcpTools.ts -> Line 326: Both maps are built unconditionally even when `matches.length === 0` (and again at line 394 when `byFile.size === 0`) — the old code did zero passes on empty result sets. Cheap to gate behind a length check, though at O(E) once per call it's unlikely to matter in practice.

- File: src/presentation/mcpTools.ts -> Line 326: Naming is slightly confusing — `targetMap`/`sourceMap` are keyed by target/source but hold incoming/outgoing edges. `incomingByTarget` / `outgoingBySource` (or the helper above) would make intent obvious.

- File: src/presentation/mcpTools.ts -> Line 331: The `targetMap.get(l.target)!` non-null assertions are safe given the preceding `has()` check, but a get-or-default idiom avoids them: `const arr = targetMap.get(l.target) ?? []; arr.push(...); targetMap.set(l.target, arr);`

Everything else looks good — ship it, ideally with the duplication extracted first.

---

giauphan:
cc @jules

The diff is standalone (those files aren't in this workspace), so here's the review based on the change itself.

**Verdict: good change overall.** The O(V·E) filter-in-loop is correctly replaced with a single O(E) adjacency pass, and semantics are preserved: link order, the `nodeMap.get(x) || x` label fallback, and empty-result behavior all match the old code. `.jules/bolt.md` is docs-only, fine. No security concerns.

Two things worth flagging:

1. File: src/presentation/mcpTools.ts -> Line ~413: In `get_file_entities`, `buildAdjacencyIndex` builds both `incomingByTarget` and `outgoingBySource` but only the outgoing map is used — half the per-call work and memory is discarded. Consider building outgoing-only (e.g., an options flag or splitting the helper) for large graphs.

2. File: src/presentation/mcpTools.ts -> Line ~346: The `if (matches.length > 0)` / `if (byFile.size > 0)` guards force `let ... | undefined` declarations plus `?.` handling at every use, for a negligible win (the index build is O(E), the same cost class as reading `links`). `const { incomingByTarget, outgoingBySource } = buildAdjacencyIndex(links, nodeMap)` unconditionally is simpler and just as fast in practice.

3. File: src/presentation/mcpTools.ts -> Line 26: Verify `GraphLink` is actually exported from `../types/index.js` — I can't confirm it from the diff, and the build breaks if it isn't.

Everything else looks correct — no logic regressions, no duplicated code paths, and the optimization is a genuine win for large graphs.

---

giauphan:
cc @jules

The diffed files aren't in this workspace, so this review is based on the diff alone.

The optimization is correct and behavior-preserving: the adjacency maps replicate the original `filter`/`map` semantics (including the `nodeMap.get(...) || id` fallback and link direction/type), the new `GraphLink` import matches usage, and O(V·E) → O(E) is a genuine win. One minor flag:

- File: `src/presentation/mcpTools.ts` -> Line ~40 (buildAdjacencyIndex options): The default `{ incoming: true, outgoing: true }` only applies when `options` is omitted entirely — it's not merged with partial objects. So `{ outgoing: true }` silently disables `incoming`, and `{}` would build *nothing* (both maps `undefined`). Both current call sites are correct, but this is a trap for future callers; consider normalizing inside the function (`const { incoming = true, outgoing = true } = options ?? {}`).

Everything else looks good — no correctness, security, or performance issues introduced.

---

giauphan:
cc @jules

The touched files aren't in this workspace (they belong to the repo under review), so I reviewed from the diff alone. The core refactor is semantically sound — the adjacency maps preserve the old `filter().map()` behavior exactly (same label fallback, same link ordering, self-loops and duplicate edges handled identically), and it correctly fixes the O(V·E) pattern. No `any` types, no security concerns. One real issue:

**File: src/presentation/mcpTools.ts -> Line ~412:** `buildAdjacencyIndex(links, nodeMap, { outgoing: true })` doesn't do what it suggests. Per the destructuring defaults in `buildAdjacencyIndex` (`const { incoming = true, outgoing = true } = options ?? {}`), passing `{ outgoing: true }` leaves `incoming` as `undefined` → it defaults to `true`, so the incoming adjacency map is still fully built and then discarded. That's wasted CPU/memory on the exact hot path this PR optimizes, and the argument is misleading (it's precisely the trap the `.jules/bolt.md` learning describes, in reverse). Since only `outgoingBySource` is used here, pass `{ incoming: false }` instead — the explicit `{ outgoing: true }` is redundant either way since it's the default.

Everything else — the `search_entities` call site (needs both maps), the `?? {}` normalization, the duplicate-comment consistency, and the `.jules/bolt.md` docs — looks good.

---

giauphan:
Stale duplicate of already-merged PR #333 (Optimize graph edge lookup performance in MCP Tools). Code is already in main. Closing.

PR #324, opened 2026-09-15
🎨 Palette: Add ARIA labels to icon-only buttons

giauphan:
💡 **What**: Added `aria-label` attributes to icon-only buttons (pagination buttons in DreamMemoryView, delete button in KnowledgeGraphView) and fixed `onClear` search behavior in OrchestrationTasksView.
🎯 **Why**: Improves accessibility for screen reader users by providing descriptive labels for buttons that otherwise only contain icons. Also improves UI responsiveness when clearing search filters.
♿ **Accessibility**: Screen readers can now properly announce "Previous page", "Next page", and "Delete project" instead of reading nothing or reading a generic button label.

---
*PR created automatically by Jules for task [922180578716754205](https://jules.google.com/task/922180578716754205) started by @giauphan*


giauphan:
cc @jules

Updated 1.54.0 → 1.54.1
Reviewed the diff — it's a small, purely additive accessibility change.

1. **Correctness** — No issues. The aria-labels don't alter behavior; `disabled={!hasPrev}` / `disabled={!hasNext}` gating is unchanged and still controls clickability correctly.
2. **Security** — Nothing added that touches user input, secrets, or DOM injection. Labels are static strings. Clean.
3. **Performance** — N/A; three string attributes, zero runtime cost.
4. **Code quality** — Good pattern: `aria-label` is the right fix for icon-only buttons (prev/next pagination, delete). Placement on the `Delete` button is especially valuable since `window.confirm` provides no accessible context otherwise.
5. **Maintainability** — Labels are hardcoded English strings. If the dashboard is ever localized, these should move to an i18n source — but flagging that now would be speculative.

One nitpick worth mentioning: in `DreamMemoryView.tsx` the label reads "Page {page + 1}" without a total (e.g. "Page 3 of 10"), and the prev/next labels don't announce page numbers. That's pre-existing and out of scope, just noting it as a possible follow-up.

**Verdict: LGTM.** No blocking issues — safe to merge.

---

PR #326, opened 2026-09-16
🎨 Palette: Add aria-labels to icon-only buttons in WikiView

giauphan:
💡 What: Added `aria-label` attributes to the "Refresh pages tree" and "Send query" icon-only buttons in the `WikiView` component.

🎯 Why: These buttons previously lacked an `aria-label`, meaning screen readers would only announce them as "button", leaving users unaware of their functions. Adding these labels improves accessibility.

♿ Accessibility:
- Added `aria-label="Refresh pages tree"` to the refresh `<button>`.
- Added `aria-label="Send query"` to the send `<button>`.

---
*PR created automatically by Jules for task [1311326766725468998](https://jules.google.com/task/1311326766725468998) started by @giauphan*


giauphan:
cc @jules

The file isn't in this repo, so I'm reviewing the diff as-is. Both changes are small, additive accessibility improvements — I don't see any correctness, security, or performance impact.

**Review:**

Both changes look good. Adding `aria-label` to the refresh button (icon-only, styled with `background: none`) is exactly the right fix since it has no accessible name otherwise.

dashboard/src/components/WikiView.tsx -> Line 665: Minor — if the "Send query" submit button already contains visible text (not visible in the diff), the `aria-label` will override it; make sure the label matches what users see, otherwise this is fine (and correct if the button is icon-only).

No security, performance, or logic concerns. Everything else looks good.

---

giauphan:
cc @jules

Updated 1.54.1 → 1.54.2
Everything looks good. Both changes are pure accessibility wins — the buttons are icon-only (no visible text), so `aria-label` is exactly the right fix, and the labels are descriptive and consistent with the surrounding English UI text.

- No logic changes, no security/performance impact.
- Minor (pre-existing, not from this diff): the refresh button lacks an explicit `type="button"` — worth adding only if it lives inside a `<form>`, otherwise it could trigger an unintended submit on click.

No issues to flag for this PR.

---

giauphan:
cc @jules

I reviewed the diff — it's a small, low-risk accessibility improvement. Both changes are correct: `type="button"` prevents accidental form submission on the refresh button, and `aria-label`s give accessible names to icon-only buttons. No correctness, security, or performance concerns.

One thing worth a quick sanity check, not a blocker: for the submit button, `aria-label` overrides the visible text as the accessible name. If its visible label isn't contained in "Send query" (WCAG 2.5.3 Label in Name), screen reader users would hear a different name than sighted users see. If the button is icon-only or says "Send", it's fine.

Everything else looks good.

---

giauphan:
Stale duplicate of already-merged PR #330 (Palette: Add ARIA labels to icon-only buttons in WikiView). Code is already in main. Closing.

PR #328, opened 2026-09-17
feat: DeepWiki custom LLM endpoint configuration & auth fixes

giauphan:
## Summary
- Enabled user-defined LLM provider options (`provider`, `baseUrl`, `apiKey`, `model`) in DeepWiki Ask Wiki Assistant Q&A.
- Implemented saving and loading custom configuration profiles (presets) to client `localStorage`.
- Fixed cramped layout inside the Wiki generation configuration modal.
- Fixed IP rate-limiting lockout issues on the backend proxy server (`authProxy.ts`) and configured Express `trust proxy`.
- Added accessibility labels and comprehensive Vitest regression suites for the configuration forms.

## Test plan
- Verify that users can open "Settings" in the Wiki View and use the "Load Saved Preset" dropdown to load configurations without typing manually every time.
- Verify that using the **Ask Wiki Assistant** queries your selected LLM correctly via the updated `QueryWikiOptions` handler.
- Verify `pnpm test` accurately catches state failures in `WikiView.test.tsx` and passes flawlessly as committed.


giauphan (commented):
cc @jules

The target files aren't in this repo — this is the review bot itself. I'll review the diff directly.

Here are my findings:

**Correctness**

- **wikiRoutes.ts line 103**: Destructuring `{ query, ...options } = req.body || {}` — if `req.body` is null/undefined, `|| {}` handles it. But spreading all remaining keys into `options` means the backend receives any extra fields from the client (e.g., `"query": "foo", "secret": "bar"` → both sent downstream). Consider whitelisting only expected keys instead of accepting an opaque spread.

- **llmProviderService.ts SSRF fix**: Good improvement over the old port-based string matching (`includes('11434')`). However, `ALLOWED_FQDNS.includes(parsed.hostname)` does exact string match, which rejects subdomains like `api.sub.openrouter.ai`. If those are intended endpoints, this blocks them. Consider whether partial matching or domain-level checking is needed.

- **WikiView.tsx Q&A endpoint (line ~257)**: Payload construction moved inside `try/catch`, which is fine — no risk here since object literals can't throw at runtime. Not an issue.

- **wikiRoutes.ts vs wikiService.ts contract change**: Routes now pass `{ provider?, apiKey?, baseUrl?, model? }` down to `llmProviderService.generateText()`. These options cascade correctly through the chain. No bug.

**Security**

- **authProxy.ts rate limit skip IPs**: Adding IPs to a `skip` list means those IPs completely bypass rate limiting for auth. If `CODEATLAS_AUTH_RATE_LIMIT_ALLOW_IPS` is misconfigured (e.g., empty string → split produces `['']`), the filter on `.filter(Boolean)` cleans it up. No critical flaw, but be careful: if someone tricks the env var to include `0.0.0.0`, every connection matches. Currently the normalize function strips IPv6-mapped prefixes so `0.0.0.0` won't match `::ffff:0.0.0.0` unless explicitly listed — safe.

- **llmProviderService.ts SSRF Regex**: The regex `/^(https?):\/\/([^\/?#]+)(\/[^?#]*)?$/` is solid — forces `https://` or `http://`, strips query strings and fragments, prevents host/path injection. Combined with the allowlist, this is well-done. One edge case: URLs with ports like `localhost:8080` — the regex handles them since `[^\/?#]+` captures everything before the first slash. ✓

- **LLM API key handling in frontend**: Keys stored only in React state (not localStorage) by design. When saving as a preset profile, the apiKey *is* written to localStorage alongside the profile JSON. This contradicts the CodeQL comment above — saved profiles contain plaintext API keys. Either encrypt them or warn users.

**Performance**

- **WikiView.tsx profiles persistence**: Writing entire profiles array to localStorage on every change via `useEffect` — if someone has many profiles with large names, repeated JSON.stringify/sync writes could cause jank during rapid edits. Use debounce or write-on-blur instead. Minor concern.

**Code Quality / Maintainability**

- **httpServer.ts route mounting duplication**: `mountWikiRoutes(app)` and `registerDreamingRoutes(app)` were removed from `startHttpServer()` and mounted earlier at module load time (lines 240-241). This changes lifecycle order — routes are registered when the module imports, potentially before config is ready. Could cause issues in tests or hot-reload scenarios where module instances are recreated. The original pattern (mount in `startHttpServer`) is safer for re-initialization.

- **llmProviderService.ts dead default URL**: `safeUrl = 'https://api.openai.com/v1/chat/completions'` is the fallback for every invalid/malformed URL. While this prevents SSRF, it silently redirects requests meant for a custom LLM to OpenAI if the hostname isn't allowlisted. A clearer failure mode would be throwing an error saying the host isn't allowed rather than swallowing into a default.

- **WikiView.tsx inline styles**: Heavy use of inline style objects in JSX — acceptable for a modal component, but consider extracting into CSS variables or styled-components if these patterns repeat elsewhere.

**Summary**: The main security win is the SSRF hardening in `llmProviderService.ts`. Two items to address:
1. Saved presets store plaintext API keys despite the explicit note about CodeQL flagging clear-text storage — either encrypt or document this limitation clearly.
2. Route mount reordering in `httpServer.ts` could break test isolation and hot-reload; prefer keeping mounts in `startHttpServer` or adding a cleanup/reset step between invocations.

giauphan (commented):
cc @jules

Here's my review:

## Issues Found

### Security — HIGH

**src/presentation/httpServer.ts → Line ~241**: Route registration duplicated — `mountWikiRoutes(app)` appears both at module scope (line ~241) and inside `startHttpServer()` (line ~1232). If `startHttpServer` is called after module load, the routes get registered twice, causing duplicate route handlers.

**src/services/llmProviderService.ts → Lines ~206-249**: The SSRF allowlist approach is better than the old string-matching, but the `ALLOWED_FQDNS` array is mutated with `push()` inside the function body. If this file is imported by multiple modules, the array accumulates duplicates. Consider building it once at module scope.

**dashboard/src/components/WikiView.tsx → Lines ~374-381**: Saved config profiles include `apiKey` stored in plaintext in localStorage. The comment above says "avoid writing api keys to localStorage" but the Save Preset feature does exactly that — it's self-contradictory. API keys in localStorage are readable by any script running in the same origin (XSS → key theft). At minimum, this should be documented as a limitation or use session-only storage.

### Logic Bugs

**dashboard/src/components/WikiView.tsx → Lines ~876-901**: Missing `Loader2` import — the JSX uses `<Loader2 size={18} className="animate-spin" />` but `Loader2` is not imported from lucide-react. This will crash at runtime when generating.

**dashboard/src/components/WikiView.tsx → Line ~102+**: All six `useEffect` persistence hooks fire on every state change independently, meaning saving a profile (which sets `profiles` AND `activeProfileId`) triggers two separate localStorage writes. Fine functionally, but unnecessary I/O.

**src/routes/authProxy.ts → Lines ~19-31**: Rate limit bump from 5 → 100 combined with the `skip` logic. The `normalizeIp` function strips IPv4-mapped prefixes but the `DEFAULT_AUTH_RATE_LIMIT_ALLOW_IPS` list includes `::ffff:127.0.0.1` which gets normalized to `127.0.0.1` — that's fine. But note: `req.ip` requires `app.set("trust proxy", 1)` to resolve correctly behind proxies, and that line is in `httpServer.ts`. If someone runs the server without that line, the skip will always compare against `undefined`, meaning no IPs get skipped and rate limiting applies to everyone (including external attackers).

### Code Quality

**src/services/wikiService.ts → Line ~275**: Function signature change from `(projectName, query, tenantId)` to `(projectName, query, options)`. All three call sites were updated correctly, but this is a breaking change for any external consumer who imports this class directly. Worth a changelog entry if this is part of a public API surface.

**src/presentation/routes/wikiRoutes.ts → Line ~103**: `const { query, ...options } = req.body || {}` — destructuring rest spreads arbitrary shape from `req.body` into `options`. Any extra fields (e.g., malicious payloads) get forwarded to `service.queryWiki()` which passes them to `llmProvider.generateText()`. While currently harmless (they're just ignored), it's cleaner to destructure explicitly.

### Low Priority / Nitpicks

**dashboard/src/components/WikiView.tsx → Line ~215**: Base URL input has no validation — empty or malformed URLs pass through silently. Consider adding format validation or a test-connect button.

---

**Summary**: Most urgent fix is the missing `Loader2` import (will cause crash) and the duplicate route registration. The localStorage API key storage is a design trade-off worth documenting, not necessarily blocking for a dashboard-only feature.

giauphan (commented):
cc @jules

I'll review this diff across the security, correctness, and quality dimensions.

---

**Security**

`dashboard/src/components/WikiView.tsx` -> Line ~45: API keys stored in localStorage via saved presets. Despite the eslint comment about keeping apiKey "in memory only", the `Save Preset` handler (line ~928) writes `apiKey: genApiKey` into the profile object which is persisted to `localStorage`. Any XSS vulnerability or malicious extension can read these keys. Consider encrypting the key before storing, or better — require re-entry of the API key at generation time instead of caching it in presets.

`src/services/llmProviderService.ts` -> Line ~173-176: The SSRF whitelist allows `0.0.0.0` as an allowed FQDN host. Resolving `0.0.0.0` typically binds to all interfaces, which could point to unexpected internal IPs depending on the runtime environment. Remove it from `ALLOWED_FQDNS` or add a guard that rejects `0.0.0.0`.

`src/services/llmProviderService.ts` -> Line ~218: The `protocol` assignment defaults to `'https:'` when `parsed.protocol` is falsy or unrecognized (e.g., `ftp:`, `file:`). An attacker controlling `baseUrl` with `file:///etc/passwd` would silently become `https://<host>/v1/chat/completions`, but if the URL parsing succeeds with a custom protocol, no validation block catches it. Add explicit allowlisting of protocols (`http:` / `https:` only).

`src/routes/authProxy.ts` -> Line ~47-49: The `skip` middleware bypasses rate limiting entirely for whitelisted IPs. Combined with the 100-request limit increase (from 5), this means any IP matching the list gets zero protection. If the intent is truly local-only skip, verify that `req.ip` resolves correctly behind proxies — `trust proxy` must be set on the Express app (it is now, good). But consider using `max: Infinity` with a per-route limiter for authenticated paths instead of skipping.

**Correctness**

`src/presentation/httpServer.ts` -> Line ~240 & ~1231: `mountWikiRoutes(app)` is registered both inline after route setup (line 240) AND inside `startHttpServer()` (line 1231). This creates duplicate route registration depending on the code path. One of them should be removed. Same applies to `registerDreamingRoutes` (removed from startHttpServer but still called globally on line 237).

`src/services/llmProviderService.ts` -> Line ~213-215: When `matchedPath` fallback triggers (`|| '/v1/chat/completions'`), a URL like `http://localhost:1234/custom/path` becomes `http://localhost:1234/v1/chat/completions` — silently changing the endpoint. This could confuse users who expected their custom path to be preserved. Either reject unmatched paths or log a warning.

`src/services/wikiService.ts` -> Line ~275: The `QueryWikiOptions` interface has `provider?: string` typed as generic `string` while being cast to `any` at line 308 (`options.provider as any`). This cast defeats type safety. Define proper enum or union types for known providers.

**Performance**

`dashboard/src/components/WikiView.tsx` -> Line ~120-122: The three `useEffect` hooks for `profiles` and `activeProfileId` cause separate re-renders whenever either changes. Since they modify localStorage independently, there's a risk of stale reads. Consolidate into a single effect or use a ref-based debounce pattern to avoid multiple synchronous localStorage writes during rapid state changes.

**Code Quality**

`src/presentation/httpServer.ts` -> Line ~1230: `registerDreamingRoutes(app);` was removed from `startHttpServer` but `mountWikiRoutes` wasn't similarly cleaned up from the global scope — leaving two overlapping registration points. Do a full audit of route registrations to ensure exactly one call site per route group.

`dashboard/src/components/WikiView.tsx` -> Line ~877-881: Inline styles are very verbose and repetitive across multiple `<div>` elements. Consider extracting reusable style objects or CSS classes for form controls used throughout the modal.

**Maintainability**

`src/services/llmProviderService.ts` -> Line ~197-208: The SSRF fix comments ("Strictly assign from compile-time string constants", "satisfies SSRF CodeQL validations") read as manual additions solely to satisfy static analysis tools. These are cosmetic annotations that don't add behavioral value. Consider adding actual unit/integration tests to validate the SSRF protection logic instead of relying on annotations.

src/presentation/httpServer.ts
```diff
@@ -237,6 +238,9 @@ app.use(authProxyRouter);
 // REST API Dreaming (dream memories)
 registerDreamingRoutes(app);
 
+// REST API Wiki
```
giauphan, line 241: Line ~240 & ~1231: `mountWikiRoutes(app)` is registered both inline after route setup (line 240) AND inside `startHttpServer()` (line 1231). This creates duplicate route registration depending on the code path. One of them should be removed. Same applies to `registerDreamingRoutes` (removed from startHttpServer but still called globally on line 237).

src/services/wikiService.ts
```diff
@@ -295,7 +304,11 @@ export class WikiService {
 
     const answer = await this.llmProvider.generateText({
       prompt,
-      systemPrompt: 'You are an advanced documentation QA assistant running on CodeAtlas platform.'
+      systemPrompt: 'You are an advanced documentation QA assistant running on CodeAtlas platform.',
+      provider: options.provider as any,
```
giauphan: Line ~275: The `QueryWikiOptions` interface has `provider?: string` typed as generic `string` while being cast to `any` at line 308 (`options.provider as any`). This cast defeats type safety. Define proper enum or union types for known providers.

giauphan (commented):
cc @jules

Here's my review of the PR:

## Security Issues

**1. Cleartext API key storage in localStorage (Critical)**
API keys entered in the generation modal are persisted to `localStorage` via the effect at `wikiService.ts`-adjacent component (the `useEffect` writing `profiles` state). Even though preset saves deliberately zero the `apiKey` field, the raw config values (`genProvider`, `genModel`, `genApiKey`, `genBaseUrl`) still get written whenever any field changes. Anyone with local access can exfiltrate `sk-...` keys directly from `localStorage`. This is already flagged by the intentional blank-out in preset save, but the live config path remains exposed.

**2. SSRF in wiki query endpoint — server-side validation missing (High)**
The query POST route at `wikiRoutes.ts:101` passes `provider`, `apiKey`, `baseUrl`, and `model` from the request body straight through to `service.queryWiki()`, then to `llmProvider.generateText()`. Meanwhile `src/services/llmProviderService.ts` now has strong SSRF controls (FQDN allowlist, path allowlist, metadata IP block). However, if any future call site bypasses `llmProviderService.ts` or adds another vector, there's no central gate on the `baseUrl` extracted from `req.body`. Consider adding SSRF validation at the route level before constructing the options object, especially since the frontend allows freeform base URL input including arbitrary hostnames.

**3. Auth rate limiter raised to 100 / 15min (Medium)**
`authProxy.ts:37` bumps the brute-force limit from 5 → 100 and adds a skip whitelist. Combined, this makes the sign-in rate limiter almost entirely ineffective. An attacker can send up to 100 attempts per 15 minutes from any non-whitelisted IP. The whitelist expansion is useful for developer comfort behind proxies, but removing the numeric cap defeats the purpose of the limiter itself.

## Correctness Issues

**4. Wiki routes potentially registered twice**
`httpServer.ts:238` calls `mountWikiRoutes(app)` inline, and `startHttpServer()` at line 1230 also calls it. If `startHttpServer()` is invoked after app-level setup (or if both paths are taken), routes register twice → duplicate handlers. Either keep mounting at the bottom or move everything into `startHttpServer()`, not both.

**5. `base_url` trim could pass empty string**
At `WikiView.tsx:225` and `WikiView.tsx:268`, the check is `if (genBaseUrl.trim()) payload.baseUrl = genBaseUrl.trim()`. This is correct for skipping empties. But on the server side, `wikiRoutes.ts:42` destructures `baseUrl` directly without trimming, so a whitespace-only string becomes `{ baseUrl: "   " }` sent to the service. The service ultimately creates a `new URL("   ")` which throws harmlessly in the SSRF guard, but silently drops the config rather than reporting a validation error.

## Code Quality

**6. `as any` cast removed but `as LLMProviderType | undefined` remains**
`mcpTools.ts:2080` and `mcpTools:2116` change from `as any` to `as LLMProviderType | undefined`. This is a valid improvement over `any` since it's at least a compile-time assertion. However, since `params.provider` comes from untyped JSON input, the cast will silently produce garbage values. Consider adding a runtime validation helper that normalizes/coalesces unknown strings to `undefined` rather than trusting the cast.

**7. Minor: unused `profileNameInput` state without clearing on modal close**
When the user closes the modal via the Close button, `profileNameInput` isn't reset. Opening the modal again carries over the stale input text. Add `setProfileNameInput('')` to the close handler.

**8. Style: duplicated inline style objects**
Three identical inline style blocks for the select, delete button, and save preset input appear in the JSX. These should be extracted to const variables (like `modalInputStyle` already was) for consistency.

---

**TL;DR**: Two action items before merge:
1. Fix the double route registration (item #4) — definite runtime bug.
2. Address the auth rate limiter (#3) and API key storage (#1) at minimum — these are real security regressions.

giauphan (commented):
cc @jules

## Issues Found

### 1. `src/services/llmProviderService.ts` — SSRF bypass via port manipulation

The `ALLOWED_FQDNS` list only checks `hostname`, but `matchedHost` resolution reconstructs the URL with an **arbitrary port**. If the request is to `api.openai.com:8080/some/path`, it resolves to `https://api.openai.com/v1/chat/completions` because `matchedPort` defaults to empty string when not matched by the regex — wait, actually looking closer, the `port` check requires a valid numeric port and falls back to `''`. So port injection into remote hosts isn't directly exploitable here. But there's still a subtle issue:

**Line ~235-240**: The fallback `safeUrl` construction strips any non-numeric port from custom-host URLs. If someone sends `http://localhost:9999/v1/chat/completions`, the regex `/^[0-9]{1,5}$/.test('9999')` passes, so `safeUrl = 'http://localhost:9999/v1/chat/completions'`. That's actually fine — localhost ports are allowed. But what about `http://api.openai.com.evil.com:443/v1/chat/completions`? The hostname check uses `.toLowerCase()` exact match, so `api.openai.com.evil.com` won't match `api.openai.com`. Good.

Actually the real issue is simpler — **the original hard-coded port detection was replaced with this new logic, but both paths have gaps**:

- **Line ~175-181** (SSRF filter): Still allows `http://localhost:*` and `http://127.0.0.1:*` unconditionally for OpenAI-compatible providers. This is intentional but means local SSRF is possible.
- The new SSRF guard at line ~226 only applies when `provider === 'openai'` (hardcoded default). For `openai-compatible` and other providers, the old SSRF-filtering path (lines 168-182) handles the baseUrl — which still trusts `localhost` without restrictions.

These aren't bugs per se (local dev needs localhost), but worth noting they'd be exploitable if an attacker can control the provider/model choice and the server has internal services on localhost.

### 2. `dashboard/src/components/WikiView.tsx` — API key stored in localStorage

**Lines 830-833**: When saving a preset profile, the API key is deliberately omitted (`apiKey: ""`). This is a good security decision. However, the individual state variables (`genApiKey`) still persist to localStorage via the bulk `useEffect` on **line 142** — wait, no, `genApiKey` is NOT included in the persistence effect. Only `profiles`, `activeProfileId`, `genProvider`, `genModel`, `genBaseUrl` are persisted. That's correct.

But the user can type an API key into the input field and it stays in component state. No immediate issue, but worth noting that if the page crashes or reloads mid-typing, the key is lost — which is probably fine for UX.

### 3. `dashboard/src/components/WikiView.tsx` — Inline styles everywhere

**Lines 770-900+**: The entire modal UI uses inline `style={{...}}` objects instead of CSS classes or styled components. This makes the file harder to maintain and violates typical React patterns. Not a bug, but a code quality concern — the project already defines `modalInputStyle` and `modalLabelStyle` constants but re-invents them repeatedly in the new JSX.

### 4. `src/presentation/routes/authProxy.ts` — Rate limit skip using `req.ip` behind trusted proxy

**Lines 35-38**: The `skip` function reads `req.ip`. This works correctly because `src/presentation/httpServer.ts` line 157 sets `app.set("trust proxy", 1)`, meaning Express trusts the first proxy header (e.g., `X-Forwarded-For`). The `normalizeIp` function normalizes IPv4-mapped IPv6 addresses consistently.

However, the `DEFAULT_AUTH_RATE_LIMIT_ALLOW_IPS` array includes `'::ffff:127.0.0.1'` but `normalizeIp` strips the `::ffff:` prefix, so after deduplication the Set contains `{ '', '127.0.0.1', '::1' }`. The empty string from `normalizeIp('')` is a potential issue — if `req.ip` is falsy, `normalizeIp` returns `''`, and the rate limiter would be skipped entirely. A null IP is unlikely in practice but defensive code should handle it:

```ts
skip: (req) => {
  const ip = normalizeIp(req.ip);
  return !ip || authRateLimitAllowIps.has(ip); // add !ip guard
}
```

Currently if `req.ip` is undefined/null, `normalizeIp` returns `''`, which is in the Set (from the DEFAULT_AUTH_RATE_LIMIT_ALLOW_IPS processing where one entry maps to empty). Wait — let me re-check. `process.env.CODEATLAS_AUTH_RATE_LIMIT_ALLOW_IPS` splits on comma. If undefined, it's `''`, split gives `['']`, trim + filter removes empty strings. Then DEFAULT_AUTH_RATE_LIMIT_ALLOW_IPS adds `['127.0.0.1', '::1', '::ffff:127.0.0.1']`, normalized to `['127.0.0.1', '::1', '127.0.0.1']` → Set de-duplicates to `{ '127.0.0.1', '::1' }`. Empty string is NOT in the set. So `normalizeIp(undefined)` returns `''`, and `authRateLimitAllowIps.has('')` is `false` — so the skip does NOT trigger. Correct behavior.

### 5. `src/services/wikiService.ts` — `queryWiki` signature change could break callers

**Line ~278**: `queryWiki` now takes `(projectName, query, options)` instead of `(projectName, query, tenantId)`. The MCP tools layer has been updated (good — I see the changes in `mcpTools.ts`), but any external caller hitting the REST API route that still sends the old 3rd positional argument pattern would be affected. The REST route `wikiRoutes.ts` line 114 calls `service.queryWiki(project, query, options)` passing the options object, so it's consistent.

No other files appear to call `queryWiki` externally based on this diff. Safe within this PR's scope.

### 6. `src/services/llmProviderService.ts` — Path matching case sensitivity

**Line ~232**: `ALLOWED_PATHS` comparison uses `.toLowerCase()` on both sides, so `/Chat/Completions` would match. This is permissive but likely intentional. Minor note — case-insensitive path matching could allow paths like `/V1/CHAT/COMPLETIONS` which might be unexpected.

---

## Summary

| Severity | File | Issue |
|----------|------|-------|
| Low | `authProxy.ts:35-38` | `normalizeIp` returns `''` for falsy `req.ip`; currently harmless but could mask future issues |
| Medium | `llmProviderService.ts:168-182` | Old SSRF filter still trusts all `localhost` ports for openai-compatible provider; acceptable for dev but document as a limitation |
| Low | `WikiView.tsx` | Heavy inline styles; refactor for maintainability |
| Info | `wikiService.ts:278` | Breaking API signature change — verified safe within PR scope |

**Overall assessment**: Clean enough to merge. The main security improvements (SSRF guard, API key not saved to presets) are positive. The SSRF bypass path via localhost is pre-existing and accepted for local development use cases.

src/presentation/httpServer.ts
```diff
@@ -155,6 +155,7 @@ export const syncQueue = new TaskQueue(1);
 
 // Setup Express app to serve as both MCP SSE and REST API
 export const app = express();
```
giauphan, line 157: The `skip` function reads `req.ip`. This works correctly because `src/presentation/httpServer.ts` line 157 sets `app.set("trust proxy", 1)`, meaning Express trusts the first proxy header (e.g., `X-Forwarded-For`). The `normalizeIp` function normalizes IPv4-mapped IPv6 addresses consistently.

giauphan (commented):
cc @jules

I've analyzed the diff across all five files. Here are my findings:

---

## Security

**File: `src/services/llmProviderService.ts` -> Line ~210**
The SSRF allowlist approach is good but has a gap: port validation only triggers when `matchedHost` is truthy — if `ALLOWED_FQDNS` contains a wildcard entry or `process.env.CODEATLAS_ALLOWED_LLM_HOSTS` includes entries like `localhost` or `127.0.0.1`, an attacker could still specify arbitrary ports via the input URL. The port regex `/^[0-9]{1,5}$/` allows up to 99999 and doesn't restrict to known safe ranges. Consider rejecting non-standard ports unless the host is explicitly in a trusted local list.

**File: `src/services/llmProviderService.ts` -> Line ~210**
Path matching uses simple string comparison (`p.toLowerCase() === parsed.pathname.toLowerCase()`), which means `/v1/chat/completions?admin=1` would match `/v1/chat/completions` because `URL.pathname` strips query strings — that's fine. But paths like `/api/v1/chat/completions/foo` won't match and will correctly throw. However, no prefix-match is used, so trailing slashes on allowed paths could slip through depending on normalization behavior of `new URL`. Verify pathname trailing slash normalization.

**File: `src/services/llmProviderService.ts` -> Line ~169**
The old hardcoded port sniffing was removed but replaced with dynamic URL construction. The `endpoint.startsWith('http')` block at line ~171 still mutates `endpoint` to append `/chat/completions` for localhost hosts before it reaches the allowlist. This means the final `safeUrl` for custom endpoints bypasses the user-supplied path entirely — the base URL path is preserved but `/chat/completions` is appended. If a malicious baseUrl has a dangerous path component, it gets doubled. Check whether the concatenation logic `safeBase.endsWith('/chat/completions')` covers all edge cases (e.g., `/v1/chat/completions/`).

**File: `dashboard/src/components/WikiView.tsx` -> Line ~885**
API keys stored in localStorage profiles. While the save preset deliberately omits `apiKey` from saved presets (good!), the current session's `genApiKey` state is persisted to `localStorage` via the single `useEffect` at line ~142 as `ca_wiki_model`, `ca_wiki_provider` etc — wait, apiKey is NOT persisted there, which is correct. However, `genApiKey` lives in React state and will be visible in plain text in the DOM and browser devtools. Consider masking it by default.

---

## Correctness

**File: `src/presentation/routes/wikiRoutes.ts` -> Line ~105**
`queryWiki` now accepts `QueryWikiOptions = {}` as the third parameter. The destructuring `{ query, provider, apiKey, baseUrl, model } = req.body || {}` handles null body but if `req.body` is undefined (body-parser not running), it defaults to `{}`. No explicit check that `provider` is validated against valid types here — validation happens in the service layer, but the route should reject invalid providers early. Same issue exists in MCP tools where `validateLLMProvider` returns `undefined` for unknown providers — silently falling back to untyped behavior.

**File: `src/services/wikiService.ts` -> Line ~278**
`queryWiki` changes signature from `(projectName, query, tenantId)` to `(projectName, query, options)`. Any callers outside the modified files (not shown in this diff) that pass a string as the third argument will break. If this change is internal-only, add a runtime check or document the breaking API change.

**File: `dashboard/src/components/WikiView.tsx` -> Line ~133**
`profiles` state initialized from JSON parse with try/catch — good. But `activeProfileId` falls back to `''` even when a valid profile exists. If a user had a saved profile from before, they'll start with an empty selection instead of loading their last active one. Should probably also persist `lastActiveProfileId` separately or auto-select the first profile if none is active.

---

## Code Quality / Maintainability

**File: `src/presentation/httpServer.ts` -> Lines ~238-241**
`mountWikiRoutes(app)` is moved from `startHttpServer()` to global scope. Meanwhile `registerDreamingRoutes(app)` is also moved up. Both were previously called inside `startHttpServer`. This means routes are registered at module load time rather than server startup time. If dependency injection needs the server port to construct callback URLs, these routes may receive stale values. Verify whether any mount function depends on the port parameter — it doesn't appear to, but removing them from `startHttpServer` should be verified to have no side effects.

**File: `dashboard/src/components/WikiView.tsx` -> Line ~61**
Inline CSS `React.CSSProperties` objects created at module scope (`modalInputStyle`, `modalLabelStyle`) are fine for readability. However, there are ~15 more inline style objects scattered throughout the modal JSX that aren't extracted (the select dropdown, buttons, preset delete button). Consider consolidating into a shared style utility or CSS modules for consistency.

**File: `src/services/llmProviderService.ts` -> Line ~210**
The SSRF fix is comprehensive but verbose. The `catch {}` blocks swallow errors silently — what fails in URL parsing or allowlist matching just disappears. At minimum, log a warning so debugging failed endpoint configurations isn't impossible.

**File: `src/presentation/mcpTools.ts` -> Line ~2056**
`validateLLMProvider` returned type `LLMProviderType | undefined`. When it returns `undefined`, downstream code (`service.generateProjectWiki`, `service.queryWiki`) receives `undefined` for provider. This falls back to default behavior, which is fine — but the `as string | undefined` cast is now unnecessary since `validateLLMProvider` returns the proper type. Clean up those casts.

giauphan (commented):
cc @jules

I can see this is a PR for the CodeAtlas project (not the review bot itself). Let me analyze the diff across all touched files.

**Files changed:** 7 files, ~100+ lines added/removed

Here's my review:

### Issues Found

**1. Security — API key silently dropped from saved profiles**
`dashboard/src/components/WikiView.tsx` -> Line ~826-832: When saving a preset profile, `apiKey` is deliberately set to `""`. This means if a user saves their config with an API key, the key is lost on reload. They'll have to re-enter it every time they load the preset, which defeats the purpose of saving. Either document this limitation prominently or implement encryption.

**2. SSRF protection — allowed paths too restrictive for some providers**
`src/services/llmProviderService.ts` -> Line ~240-246: The `ALLOWED_PATHS` array only allows `/v1/chat/completions`, `/chat/completions`, `/api/v1/chat/completions`, and `/api/chat/completions`. Some OpenAI-compatible providers use different paths (e.g., `/openai/v1/chat/completions`). This could break legitimate use cases. Consider adding `/openai/v1/chat/completions` or allowing a suffix match.

**3. Performance — localStorage write on every render cycle**
`dashboard/src/components/WikiView.tsx` -> Line ~145-150: The `useEffect` that persists config writes 5 keys to localStorage on *every* state change. For fast typers, this creates excessive synchronous writes. Consider debouncing or throttling these writes.

**4. Dead code removal needed — route mounting moved but old calls still there**
`src/presentation/httpServer.ts` -> Line ~1232: The function `startHttpServer()` now handles route registration internally, and the routes are mounted at app-level (line 239-241). But line 1232-1236 also registers routes inside the start function. Check if the old mount calls before line 1232 (`registerDreamingRoutes(app)` and `mountWikiRoutes(app)`) are truly dead now — the diff removes them from `startHttpServer` but adds them back below. Verify there's no double-mounting issue.

**5. Input validation — Base URL accepts arbitrary URLs without SSRF check on client side**
`dashboard/src/components/WikiView.tsx` -> Line ~797-810: The new "Custom API Endpoint" input accepts any string without validation. While the backend validates via `llmProviderService.ts`, the frontend should at least attempt basic URL parsing validation before sending to the server, providing immediate feedback instead of failing after a request roundtrip.

**6. API surface change — `queryWiki` overloading is confusing**
`src/services/wikiService.ts` -> Line ~277: `queryWiki(projectName, query, optionsOrString)` where the third parameter is either a `string` (tenantId) or `QueryWikiOptions` object. This overloaded pattern is fragile — if callers accidentally pass `{}` as tenantId, it becomes truthy and causes issues. Consider keeping the old signature separate or always requiring the options object.

**7. Rate limiter — default limit lowered without consideration**
`src/routes/authProxy.ts` -> Line ~33: Changed from `max: 5` to `max: 10` sign-in attempts per 15 minutes. Not a bug per se, but worth noting that the increase was not accompanied by an explanation of why. If this is intentional for UX reasons, it should be documented.

### Positive Changes (no issues)

- **SSRF hardening** in `llmProviderService.ts` — the new allowlist approach with hostname + path + port validation is a significant security improvement over the old regex-based port matching
- **Provider type safety** — adding `LLMProviderType` union type instead of raw strings is good
- **IP normalization** in auth proxy handles IPv4-mapped IPv6 addresses properly
- **"trust proxy" setting** added for correct IP detection behind load balancers

src/presentation/httpServer.ts
```diff
@@ -237,6 +238,9 @@ app.use(authProxyRouter);
 // REST API Dreaming (dream memories)
 registerDreamingRoutes(app);
```
giauphan, line 240: Line ~1232: The function `startHttpServer()` now handles route registration internally, and the routes are mounted at app-level (line 239-241). But line 1232-1236 also registers routes inside the start function. Check if the old mount calls before line 1232 (`registerDreamingRoutes(app)` and `mountWikiRoutes(app)`) are truly dead now — the diff removes them from `startHttpServer` but adds them back below. Verify there's no double-mounting issue.

giauphan (commented):
cc @jules

I've reviewed the changes in this pull request. Here are my observations and suggestions:

### Overall Assessment
The changes appear to be a significant enhancement to the wiki functionality in the CodeAtlas project, adding configuration profiles, better LLM provider handling, and improved security. The code quality is generally good, but there are some areas that could be improved.

### Specific Feedback

1. **TypeScript Improvements**:
   - The repeated "Check for TypeScript best practices" and "Avoid any types" comments in the PR description suggest there might be some type-related issues. The changes do include type definitions, but some areas could benefit from stricter typing.
   - In `wikiRoutes.ts`, the provider validation could be more type-safe by using the `LLMProviderType` enum rather than string literals.

2. **Security Improvements**:
   - The SSRF protection in `llmProviderService.ts` is a good addition, but could be more comprehensive by:
     - Adding more allowed domains (e.g., `api.deepseek.com`, `api.groq.com`)
     - Making the allowed domains configurable via environment variables
   - The auth rate limiting in `authProxy.ts` is good, but the IP normalization could be more robust to handle IPv6 addresses properly.

3. **Code Quality**:
   - The modal styling in `WikiView.tsx` is now more consistent with the use of `modalInputStyle` and `modalLabelStyle` constants, which is good.
   - The profile management in the wiki generation modal is well-implemented, but could benefit from:
     - Adding validation for profile names to prevent duplicates
     - Adding a confirmation dialog when deleting profiles

4. **Error Handling**:
   - The error handling in the wiki generation could be more user-friendly by providing more specific error messages to users.
   - The `handleGenerateWiki` function in `WikiView.tsx` could benefit from more detailed error states to help users understand what went wrong.

5. **Performance**:
   - The debounced localStorage writes in `WikiView.tsx` are a good practice to prevent excessive synchronous writes.
   - The profile management could potentially benefit from a more efficient data structure if the number of profiles grows large.

### Suggested Improvements

1. Consider adding a loading state for profile operations to prevent duplicate submissions.
2. The API key handling in profile saving could be improved by:
   - Not storing API keys in localStorage at all (or at least encrypting them)
   - Adding a warning about the security implications of storing API keys
3. The modal UI could benefit from:
   - Better visual feedback when operations are in progress
   - More consistent spacing and alignment of form elements

### Final Verdict
The changes are generally well-implemented and represent a significant improvement to the wiki functionality. The security enhancements are particularly welcome. With some minor refinements to error handling and user feedback, this would be ready for merge.

**Recommendation**: Merge with minor adjustments for error handling and user feedback improvements.

giauphan (commented):
cc @jules

I've reviewed the changes in this pull request and here are my observations:

1. **WikiView.tsx Enhancements**:
   - The addition of LLM configuration profiles is a good improvement, allowing users to save and switch between different LLM configurations.
   - The debounced localStorage persistence is a good practice to prevent excessive synchronous writes.
   - The URL validation for the base URL is helpful for preventing invalid configurations.
   - The UI improvements with consistent styling and better organization of the modal content are welcome.

2. **Security Improvements**:
   - The addition of `app.set("trust proxy", 1)` in httpServer.ts is important for proper IP handling behind proxies.
   - The enhanced SSRF protection in llmProviderService.ts with stricter URL validation is a significant security improvement.
   - The IP normalization and rate limiting configuration in authProxy.ts is good for preventing brute force attacks.

3. **Type Safety**:
   - The addition of LLMProviderType type in wikiService.ts is helpful for type safety.
   - The validation of provider types in multiple places (wikiRoutes.ts, mcpTools.ts) ensures type safety.

4. **Code Quality**:
   - The consistent use of modalInputStyle and modalLabelStyle for styling is good for maintainability.
   - The error handling improvements in the wiki generation flow are welcome.

5. **Minor Issues**:
   - The repeated "Check for TypeScript best practices" and "Avoid any types" comments in the project context seem redundant and could be removed.
   - The diff shows some repeated sections (like the TypeScript best practices comments) which might be a copy-paste artifact.

Overall, the changes appear to be well-structured and address important aspects of security, type safety, and user experience. The code is clean and follows good practices. I recommend merging this pull request after addressing the minor issues mentioned above.

giauphan (commented):
cc @jules

I've reviewed the changes in this Pull Request. Here are my observations and suggestions:

### Overall Quality
The changes look generally well-structured and follow the project's patterns. The additions for LLM configuration profiles and saved presets are a good improvement for the user experience.

### Specific Feedback

1. **TypeScript Improvements**:
   - The new `WikiLLMConfigProfile` interface is well-defined and properly typed.
   - The `LLMProviderType` type is a good addition to enforce valid provider values.

2. **Code Quality**:
   - The debounced localStorage persistence is a good practice to prevent excessive synchronous writes.
   - The URL validation in `handleGenerateWiki` is solid.
   - The profile management UI is well-implemented with proper state management.

3. **Security Improvements**:
   - The SSRF protection in `llmProviderService.ts` is comprehensive and covers multiple attack vectors.
   - The `normalizeIp` function in `authProxy.ts` is a good addition to handle IPv6 addresses properly.

4. **UI/UX Enhancements**:
   - The profile management interface is well-designed with clear visual hierarchy.
   - The modal styling is consistent with the rest of the application.
   - The save/delete actions have appropriate visual feedback.

5. **Potential Issues**:
   - In `wikiRoutes.ts`, the provider validation should be consistent with the `LLMProviderType` type.
   - The API key handling in the UI could be improved to show a mask/obscure the value when displayed.

6. **Minor Improvements**:
   - The `trust proxy` setting in `httpServer.ts` is a good addition for production environments.
   - The error handling in the profile management could be more robust (e.g., duplicate names).

### Suggested Changes

1. **Consistency in Provider Validation**:
   - Ensure the provider validation in `wikiRoutes.ts` matches the `LLMProviderType` type exactly.

2. **API Key Handling**:
   - Consider adding a visual indicator that the API key is saved but not shown in plaintext.

3. **Error Handling**:
   - Add more specific error messages for profile management operations.

4. **Testing**:
   - The changes should be tested with various edge cases, especially around URL validation and profile management.

Overall, this is a well-executed PR that adds valuable functionality while maintaining the project's quality standards. The changes are well-scoped and implemented correctly.

PR #331, opened 2026-09-18
⚡ Bolt: [performance improvement] Speed up executeMany update and deletes using batched queries with chunking in consolidationEngine

giauphan:
💡 What: Changed sequential N+1 deletion and updates utilizing `db.executeMany` into a chunked IN clause structure for large deletes within `deduplicateDreams` and `scoreDreams` inside `consolidationEngine.ts`.
🎯 Why: The original application performed queries sequentially via an external `db.executeMany`, which represents a major N+1 performance bottleneck over multiple loop iterations instead of completing the execution in a chunked, single query format.
📊 Impact: Considerably faster updates/deletions during consolidation pipeline runs on instances with heavily overlapping memory spaces.
🔬 Measurement: Verify pipeline speeds via logging statements surrounding ConsolidationEngine processes in production workloads.

---
*PR created automatically by Jules for task [16544233909693517443](https://jules.google.com/task/16544233909693517443) started by @giauphan*


giauphan:
cc @jules

Updated 1.55.1 → 1.56.0
This PR has a **blocking issue**:

**src/services/consolidationEngine.ts → Line 15:** Imports `buildInClause` from `"../database/utils.js"`, but that file doesn't exist. The import path resolves to `src/database/utils.js`, which isn't present in the project at all. Without implementing this utility, this change won't compile.

Beyond that blocker, if I assume `buildInClause(chunk, binds)` is implemented as `{ clause: string; binds: Record<string, unknown> }`:

1. **Missing new utility file** — Need to create `src/database/utils.ts` with `buildInClause`. If the existing codebase already has a similar helper somewhere (e.g., parameter building), prefer extracting/reusing rather than creating another utility.

2. **`clause` interpolation risk** — `buildInClause` needs to generate safe placeholder strings (`?, ?, ?`) rather than raw parameterized placeholders (`:id`). SQLite's Node driver uses `?` positional parameters. If it generates `?` placeholders, the chunking works fine. But if the caller then also passes named params like `{ tenantId }` alongside `?` positional args in the same query, there's a mixed placeholder style problem. Check how `db.execute` is configured — some drivers don't mix `?` and `:named`.

3. **Chunk size of 900** — Reasonable for avoiding SQLite limits (~999 vars per statement). No issue here.

4. **Logic correctness** — The batching approach is sound for deleting/superseding records. However, within each chunk, `tenant_id` is still passed as a single named param. If the total IDs across chunks are large, individual chunk errors could leave partial deletions. The original `executeMany` had the same tradeoff though — not a regression.

Main ask before merging: **add and verify the `src/database/utils.ts` implementation**, confirm placeholder style compatibility, then re-run the build to ensure it compiles.

---

giauphan:
cc @jules

Reviewing the PR diff, here are my findings:

## Critical Issues

**1. Missing dependency — `buildInClause`**
The new import `import { buildInClause } from "../database/utils.js"` references a module that doesn't exist in this repository. There's no `src/database/` directory or `utils.ts/js` file. This will fail at runtime with a module resolution error unless this utility is added in a parallel commit or branch.

**2. Potential performance regression**
The optimization comments claim "N+1 avoidance", but the original `executeMany` already batches all IDs into a single query. For 2700 IDs:
- Old: 1 `DELETE ... WHERE id = :id AND tenant_id = :tenantId` (one execution)
- New: 3 queries with `DELETE ... WHERE id IN (?, ?, ...) AND tenant_id = :tenantId`

This is strictly worse unless there's an undocumented SQLite limitation with `executeMany` that needs to be documented in the PR.

## Warnings

**3. Bind variable consistency — UPDATE case**
Looking at line 606 update block, the original binds `{ sid: id, tid: authStorage.getStore()!.uid }` maps `tid` → `tenant_id` filter. The new code only passes `tid` through `buildInClause`, but it's unclear if that helper includes extra bindings like `tid` — if not, the `WHERE ... AND tenant_id = :tid` could resolve to undefined/unbound values. Need to verify `buildInClause`'s signature handles this correctly.

**4. Chunk size magic number**
`chunkSize = 900` is hardcoded without explanation. The standard SQLite parameter limit is 32,766 (not 999), so 900 is very conservative. If the real constraint is something else (e.g., a wrapper library), document why.

**5. No error handling in loop**
The `for` loop executes each chunk sequentially inside a `try/catch` that wraps the entire block. If one chunk fails, remaining chunks are skipped silently — same behavior as before, but worth noting since the `merged += toRemove.size` count would be inaccurate on partial failure.

---

giauphan:
cc @jules

**What's happening:** Two places where `executeMany` was used for batch deletes/updates are being replaced with chunked `IN` clause queries via a new `buildInClause` helper. Both target SQLite (900-chunk limit).

**Issues I spot:**

1. **src/services/consolidationEngine.ts -> Line 297-298:** The comment about merged count is just restating that it always adds `toRemove.size` regardless of errors — which was already the original behavior. The comment adds noise without clarifying anything new.

2. **src/services/consolidationEngine.ts -> Line 614:** Same pattern, same thing — `supersededCount = toSupersede.size` was always assigned unconditionally in the original code too.

3. **Potential correctness issue (both blocks):** If `buildInClause` generates bind parameters like `? ? ? ...`, those are positional and SQLite should handle them fine in an IN clause. But if `buildInClause` uses named binds (e.g., `:id0, :id1`), they must NOT be passed inside the `binds` object for `db.execute` — SQLite's `execute` method typically only accepts positional binds or a specific format. Worth verifying that `buildInClause` actually works with the database layer's bind handling. If it doesn't, these queries silently do nothing and bugs will be very hard to trace.

4. **Performance tradeoff worth noting:** The original `executeMany` likely did something similar under the hood (batched execution). This optimization only helps if `executeMany` was doing one query per ID (which would be genuinely bad for N IDs). If `executeMany` was already batching into a single statement, this change may be unnecessary — though the chunking at 900 is a reasonable safety net for SQLite's parameter limits.

5. **Inline comments (`// ⚡ Bolt Optimization`):** These explain why the code exists rather than what it does. After a brief while, anyone reading this can see the chunking pattern — the comments become stale documentation risk. Fine for now, but low signal.

Overall: the intent is solid (avoid N+1 / parameter limits), but I'd want to confirm `buildInClause`'s contract before merging.

---

giauphan:
cc @jules

These files aren't in this repo, so I'll review the diff directly.

**File: `src/services/consolidationEngine.ts` -> Line 17+: Import of `buildInClause`**

Assuming `buildInClause(chunk, { tenantId })` correctly generates bind parameters for each placeholder (`:0`, `:1`, etc.) AND merges them with `{ tenantId }`, there are concerns:

**Issue 1 — Bind parameter collision (potential)**
Lines 283-284 and 610-611 pass an object to `buildInClause`:
```ts
const { clause, binds } = buildInClause(chunk, { tenantId });
await db.execute(`DELETE ... WHERE id IN (${clause}) AND tenant_id = :tenantId`, binds as any);
```
If `buildInClause` merges `{ tenantId }` into its own bind objects, you end up with duplicate keys per array element. If it doesn't merge, `:tenantId` is referenced in SQL but not bound. Either way, **this likely needs verification against the actual `buildInClause` implementation.** The same concern applies at line 610 with `{ tid }`.

**Issue 2 — Error swallowing**
Line ~285: `} catch { // skip delete errors }` wraps the chunked loop entirely. If one chunk fails, the remaining chunks are never executed. The original code had the same pattern (one call), so this is a regression if partial failures were expected to continue. Consider moving the try-catch inside the loop or retrying failed chunks.

**Issue 3 — `merged += toRemove.size` outside try-catch** ✅ OK
This moves increment past the try block, which means we count attempted IDs regardless of success. Acceptable given the silent skip, but the comment should clarify intent.

**Correctness — Chunk size of 900**
Reasonable for Oracle (limit ~1000 params/query) or PostgreSQL (similar). Verify the database backend supports this pattern. No issue if that's the constraint.

**Security — Tenant isolation preserved** ✅
Both queries still include `AND tenant_id = :tenantId` (or `:tid`) in the WHERE clause. Safe against cross-tenant deletes via manipulated IDs.

**Performance — Batch vs executeMany**
Replacing `executeMany` with chunked loops adds N/900 roundtrips instead of 1. This is clearly intentional (likely hitting a database limit with executeMany). If that's the case, worth adding a comment explaining why `executeMany` was abandoned.

---

giauphan:
cc @jules

`buildInClause` isn't in this repo and the diff doesn't include `src/database/utils.ts`, so I can only review what's visible plus the dependency assumptions.

**Review:**

- File: src/services/consolidationEngine.ts -> Line 294: `merged += toRemove.size` moved outside the try/catch — it now increments even when every delete chunk fails, inflating the `merged` metric. Old code only counted on success. Fix: increment per successful chunk or only when no error is thrown.
- File: src/services/consolidationEngine.ts -> Lines 288, 614: `binds as any` — violates the project's "avoid any types" rule. Type `buildInClause`'s return properly (e.g. `Record<string, unknown>`) and drop the casts.
- File: src/services/consolidationEngine.ts -> Lines 285, 611: can't verify `buildInClause` from this diff — confirm it emits parameterized placeholders (`:id_0`, ...) rather than interpolating values into `clause`, and that the extra binds (`tenantId`/`tid`) don't collide with generated bind names. If it interpolates, that's SQL injection.
- File: src/services/consolidationEngine.ts -> Lines 283-290: chunked deletes are no longer all-or-nothing like the old single `executeMany`; a mid-chunk failure (silently swallowed) leaves partial deletions. Acceptable given errors are ignored anyway, but worth knowing.
- Nit: `chunkSize = 900` is duplicated in both blocks — hoist to a module constant (Oracle's IN-list limit is 1000, so 900 is a sane choice).

Otherwise the approach is sound: chunked `IN` lists avoid per-row round trips and stay under the Oracle 1000-expression limit, and the `tenant_id` guard is preserved in both statements.

---

giauphan:
cc @jules

The repo here doesn't match the PR path (`src/services/consolidationEngine.ts` and `src/database/utils.ts` don't exist). But reviewing the diff directly:

**Changes overview**: Replaces `db.executeMany()` (one DELETE/UPDATE per row) with chunked batch `IN` clause using a new `buildInClause()` helper from `../database/utils`.

Here's my review:

1. **Correctness — missing import file**: The diff imports `buildInClause` from `../database/utils.js`, but that file doesn't exist in the repo. This is likely a new file the PR should also include. Verify it exists and returns `{ clause: string, binds: Record<string, unknown> }` with correct named parameters matching the SQL template (`?_0`, `?_1`, ... or similar).

2. **Correctness — double tenant_id binding**: In both queries, the `tenantId` bind appears twice — once inside the `buildInClause` result merge (or separately), and again explicitly: `AND tenant_id = :tenantId`. If `buildInClause` merges binds into an object keyed by `id_0`, `id_1`, etc., then `tenantId` being passed separately in the `binds` spread is fine. But if `buildInClause` expects tenant-specific handling or if there's a key collision (e.g., if it uses `tenantId` as a key too), this could silently overwrite. Check `buildInClause`'s implementation for key naming.

3. **Correctness — supersededCount logic change**: Old code set `supersededCount = toSupersede.size` (assumed all rows matched). New code sets `supersededCount += result.rowsAffected || 0` (actual DB-affected rows). This is *more accurate* but could cause downstream behavior differences if callers rely on the count matching exactly the number of IDs requested. Unlikely to be a real issue, but worth noting.

4. **Security — OK**: No injection risk. The `IN (${clause})` placeholder gets replaced with named/bindable parameters (`:p_0`, etc.), not interpolated raw values. Good.

5. **Performance — improvement**: Switching from N individual statements to chunked batches (size 900) reduces round-trips significantly. Chunk size of 900 is sensible for avoiding bind variable limits (~1000 is typical Oracle limit).

6. **Code quality — consistent pattern**: Both delete and update blocks now follow identical structure. Clean refactor.

7. **Minor — error swallowing**: Error handling is unchanged (`catch {}` silently skips). This pre-exists the PR, but the chunked approach means partial failures are more granular — some chunks may succeed while others fail, whereas the old executeMany was atomic-all-or-nothing.

---

giauphan:
cc @jules

These files don't exist in this repo. But reviewing the diff itself:

**Issues found:**

1. **Missing dependency** — imports `buildInClause` from `"../database/utils.js"` but neither `consolidationEngine.ts` nor a `database/utils.ts/js` module exists in this project. This will fail to compile/import until that utility is created.

2. **Chunk size 900 is reasonable** but should be documented as a constant rather than magic number. The common constraint (Oracle binds ≤ 1000, MySQL max_allowed_packet) justifies it, but hardcoding `900` inline makes intent unclear. Extract to a named constant like `BATCH_DELETE_CHUNK_SIZE`.

3. **Type assertion safety** — `binds as Record<string, unknown>` loses type safety on `buildInClause`'s return value. If `buildInClause` produces non-string-keyed binds, this silently breaks. Better to let the caller infer the type or use a proper interface.

4. **rowsAffected guard is correct** but worth noting: if the underlying driver returns `undefined` for `rowsAffected`, the accumulator stays unchanged. This is defensive and appropriate.

5. **Error swallowing unchanged** — `.catch {}` still suppresses errors entirely within chunks. Partial success is now possible (some chunks succeed, others fail silently) where before `executeMany` was atomic-ish. This is arguably better behavior but the comment "skip delete errors" is no longer fully accurate.

6. **`supersededCount` increment uses `+=` instead of assignment** — correct because partial failures mean we can no longer assume all items were processed. Good fix.

The overall pattern (chunking large deletes/updates into batches of ~900 to avoid DB parameter limits) is sound. Just needs the missing `buildInClause` helper and a named constant for the chunk size.

---

giauphan:
cc @jules

Found both functions are completely absent from the codebase. That alone is the primary issue.

---

giauphan:
cc @jules

Here's my review of the diff:

**Critical: This PR cannot be applied to this repo.**
- `src/services/consolidationEngine.ts` does not exist in this codebase (no `src/services/` directory)
- `buildInClause` imported from `../database/utils.js` does not exist anywhere — no `src/database/` directory at all

**If this were applied to a repo where those files existed, here are the issues I see in the logic:**

`src/services/consolidationEngine.ts` -> Line 281: Logic bug in `rowsAffected` counting — after chunked deletes, `merged += result.rowsAffected || 0` counts only actual deleted rows instead of total IDs processed. If some rows are deleted by another session between chunks, the count becomes inaccurate. The original `merged += toRemove.size` was simpler and more consistent with intent.

`src/services/consolidationEngine.ts` -> Line 607: Same `rowsAffected` counting issue on the supersede path. Plus the `tenant_id = :tid` bind is redundant — `tid` is already bound as the first key in the `buildInClause(chunk, { tid: ... })` call, which would conflict with the parameter name.

`src/services/consolidationEngine.ts` -> Line 283: `BATCH_DELETE_CHUNK_SIZE` defined inside the loop block — repeated definition on each iteration (though harmless, it should be moved up).

`src/services/consolidationEngine.ts` -> Lines 279-286 & 604-615: No change in error semantics. The outer `try/catch { skip delete errors }` still catches errors but now silently drops partial chunk failures without logging anything — same issue as before, just harder to reason about.

---

PR #332, opened 2026-09-18
🎨 Palette: Add keyboard focus ring to primary buttons

giauphan:
💡 What: Added the `FOCUS_RING_CLASS` to the primary `<motion.button>` elements for "Refresh Tasks" and "Run Consolidation".
🎯 Why: These primary interactive buttons lacked visible keyboard focus indicators, making the interface difficult to navigate for keyboard-only users.
📸 Before/After: Visual outline for keyboard users is now present on focus.
♿ Accessibility: Improved keyboard navigation by providing a clear visual indication of focus for critical interactive elements on these views.

---
*PR created automatically by Jules for task [7547378378455468680](https://jules.google.com/task/7547378378455468680) started by @giauphan*


giauphan:
cc @jules

This diff comes from a repo I don't have in the workspace (there's no `dashboard/` directory here and `FOCUS_RING_CLASS` appears nowhere in this project), so I reviewed it from the diff alone.

Overall this is a low-risk, accessibility-positive change — adding a focus ring to two buttons, no logic touched. Only one thing actually needs verification:

- File: dashboard/src/components/OrchestrationTasksView.tsx -> Line 99: `FOCUS_RING_CLASS` is referenced but the diff shows no import/definition being added. If the constant isn't already imported in this file, it won't compile. Same applies to SecondBrainView.tsx (line 121). Worth confirming both files have it in scope.

Minor, non-blocking:

- File: dashboard/src/components/OrchestrationTasksView.tsx -> Line 99: If `FOCUS_RING_CLASS` relies on a class-based `box-shadow`/`outline` that conflicts with the gradient `style` (inline styles win on any overlapping properties like `border`), verify the ring actually renders — the two styling approaches coexisting is the main failure mode here. Not a bug if the class only sets focus-visible outline/box-shadow, which the inline style doesn't touch.

- Code quality (pre-existing, not introduced here): both buttons duplicate the identical gradient/padding/radius styling inline. A shared `Button` component would make focus styling a one-place fix, but that's beyond this PR's scope.

Everything else looks good — no security, performance, or correctness concerns in these two hunks.

---

giauphan:
cc @jules

Updated 1.56.1 → 1.56.2
The dashboard files aren't in this repo, so I reviewed the diff on its own.

1. **Correctness** — File: dashboard/src/components/OrchestrationTasksView.tsx and SecondBrainView.tsx -> Lines ~99/~121: `FOCUS_RING_CLASS` is used but no import/definition appears anywhere in the diff. If it's not already imported in these files, this fails to compile (or silently renders as `undefined` in a JS build). The shared constant should live in a single CSS/module file and be imported in both — verify both files actually import it.

2. **Code quality** — Both buttons show the identical inline `style` (padding, gradient, border, radius...). Not introduced by this PR, but worth noting: extract a shared `Button` component/`BUTTON_CLASS` like you did with `FOCUS_RING_CLASS`.

No security or performance concerns. The change itself (focus visibility for keyboard users) is a good accessibility improvement — just confirm the constant is imported in both files.

---

giauphan:
cc @jules

The dashboard files aren't in this workspace, so I reviewed from the diff alone.

The change is low-risk and purely additive — a focus ring for accessibility on two disabled-toggling buttons. Two things worth checking:

- File: dashboard/src/components/OrchestrationTasksView.tsx -> Line 99: `FOCUS_RING_CLASS` is used but the diff shows no added import in either file. If it's not already imported (e.g. from a shared styles module), both components will render `className="undefined"` and the focus ring silently won't apply. Verify the constant is defined in both files.
- File: dashboard/src/components/SecondBrainView.tsx -> Line 121: Minor — inline `style` wins over class rules, so if `FOCUS_RING_CLASS` styles `border` (vs `outline`/`box-shadow`), the inline `border: 'none'` will suppress it. Prefer outline/box-shadow-based focus styles.

No correctness, security, or performance concerns otherwise. Everything looks good if the import checks out.

---

giauphan:
cc @jules

The changes aren't in this repo's tree, so I reviewed from the diff alone. Functionally sound (accessibility win), with one thing to verify.

**Issues:**

- File: dashboard/src/components/OrchestrationTasksView.tsx -> Line 99: `FOCUS_RING_CLASS` is used but the diff adds no import. If it's a shared constant not already in scope in this file, compilation fails. Same applies to:
- File: dashboard/src/components/SecondBrainView.tsx -> Line 121: `FOCUS_RING_CLASS` used without a visible import. Ensure both files import it from wherever it's defined.

Everything else looks good — the addition is low-risk, works with framer-motion's `motion.button` (className is forwarded), and `:focus-visible`-style rings won't conflict with the inline gradient styles since focus styles target a different state.

---

